Issue #14, opened 2015-07-20
As an administrator, I want to be able to manage a campaign for organizations, so potential volunteers can see it


tonysurma:
Will be fleshed out more but looking for full campaign management (CRUD)

This user story covers the following:
•Creating the campaign
•Modifying the campaign
•Deleting the campaign

mheggeseth:
Working on this

PR #65, opened 2015-08-09
Remove uneeded Admin/Campaign/Create view

mheggeseth:
Refactor #14. Remove Admin/Campaign/Create (it was virtually identical to the Edit view).


AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignAdminController.cs
```diff
@@ -39,7 +39,7 @@ private ViewResult WithTenants(ViewResult view)
         // GET: Campaign/Create
         public async Task<IActionResult> Create()
         {
-            return await Task.Run(() => WithTenants(View()));
+            return await Task.Run(() => WithTenants(View("Edit")));
```
BillWagner: A better way to do this is to use Task.FromResult:

Task.FromResult(WithTenants(View("Edit"));

That way, the compiler doesn't generate all the async state machine, and this simply always returns a completed task.

mheggeseth:
Works for me. C# async stuff is new to me so I was following the existing pattern. I've changed the methods in the controller to `await Task.FromResult()` whenever they don't return a trivial result.

Issue #33, opened 2015-08-08
5.1.1.3.2	Allow organization ‘admins’ to register with username and password on AllReady site as well


tonysurma:

gftrader:
I'll give this one a shot.

PR #66, opened 2015-08-09
Updated Registration to allow organization admins to register with em…

gftrader:
…ail and password on the site. SendGridDevelopmentWeb.cs prevents the need for an actual SendGrid account when performing registration (the new class saves the email as a .eml file on the local machine).

--

5.1.1.3.2 Allow organization ‘admins’ to register with username and password on AllReady site as well #33

@BillWagner @tonysurma Not sure if the way I did this is the best architecturally. I thought it might be possible to use IoC to configure whether or not the application should use SendGrid or the local file solution I created (SendGridDevelopmentWeb.cs), but I am not familiar with AspNet 5's IoC configuration. Also, I used a user secret for the "DevelopmentEmailFolder" key, which is where the application will write email files out when the application is run in debug mode.

Finally, the config.json file appears to be saying a lot changed, but I only added 2 keys ("DefaultFromEmailAddress", "DefaultFromDisplayName") - the rest should be the same.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SiteAdminController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SiteAdminController.cs
@@ -7,6 +7,7 @@ using AllReady.Models;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
+using AllReady.Services;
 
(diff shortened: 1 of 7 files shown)


BillWagner:
I looked at the code, and it's all good to me. @tonysurma if you can check with the vNext team and see if this is a reasonable way to handle offline testing for email in vNext, that would be great.

PR #68, opened 2015-08-09
initial work to move to Auth0

TimHess:
@BillWagner 
This is NOT finished, _userManager isn't pulling users back yet, but now you can see what I've got going so far


MisterJames:
@TimHess super awesome to have you helping out on this. Sorry we left this sitting so long; it looks like we've mostly paved through the code needed to get these objectives covered off (we're at a code-a-thon this weekend trying to wrap up as much as possible). There are a lot of moving parts!

We'll be wrapped up on Monday and a lot of the velocity will subside; please check back in there for outstanding issues. We're really looking forward to your future contributions.

cc/ @BillWagner

PR #70, opened 2015-08-10
Partial fix for making web app API url configurable in Mobile App

jonatwabash:
I've added a domain url line in the config.xml file, but I haven't managed to get the app to read that url when services.js first runs. If anyone has a great idea how to do so with Cordova, the help would be appreciated.


MisterJames:
@jonatwabash Thank you for your efforts here in helping out on this. Sorry we left this sitting so long; it looks like this is no longer an issue.

We're doing a code-a-thon this weekend with a group of 20+ of us all committing and trying to make things better. We'll be wrapped up on Monday and a lot of the velocity will subside; please check back in there for outstanding issues. 

We're really looking forward to your future contributions!

cc/ @BillWagner

PR #71, opened 2015-08-11
changing test.com to example.com

robfoulk:
test.com is a legitimate domain. Sample data should probably reference the
recommended example.com domain.  Changed the config file containing the
sample data and the readme to match.


mheggeseth:
@BillWagner Any reason this shouldn't get merged in? The only file not updated was CONTRIBUTING.md and that file is really just a copy of README.md anyway.

BillWagner:
It slipped through my list of open PRs. Thanks for bringing it back up for me to see.

Issue #45, opened 2015-08-08
5.1.2.2.3	Communicate to volunteers via email and txt when assignment requests are made with link to website (which ideally provides link to launch/install app) to interact with request


tonysurma:

MisterJames:
@tonysurma I can take this one. Here's what I'd like to suggest for path:

1. Command executes which iterates over the list
1. Queue is populated (Azure Service Bus?) with contact tasks
1. Service/webjob/something pulls from queue and processes tasks

I'll get the plumbing in place and won't make assumptions about who we'd use for providers. If we want to see it in action, I've worked with Twilio and Sendgrid, both are fairly easy to write. I'd wire them up with DI in the queue processor so we can easily swap out if needed down the road.

Curious...you have "mobile app" on here as a label; is that just to get the app installed per the issue title?

tonysurma:
@MisterJames It is safe to assume we will use twilio and sendgrid as those are the two we have and will plan on using (good guesses!)

and yes mobile app is just the get app installed potential link etc.

Thanks!

joelhulen:
@MisterJames As for watching the Azure queue, while I was thinking about taking this one on earlier (you beat me to the punch :), I thought that a webjob would make most sense for processing the contact tasks within the queue. The primary reason I think this is better than, say, a worker role, is because the webjob rides on top of the existing website, thus not requiring an additional service to be provisioned on the account. Plus, it's easy to scale either alongside the website, or separate from it.

Just my $0.02 :)

MisterJames:
@joelhulen Agreed, and I'll go the webjob route on that side of things. Thanks!

MisterJames:
Awesome, thanks @BillWagner. I'll get on the queue processing this evening.

BillWagner:
The two commits referenced above finish much of this. We still need the code that implements the queue processing as referenced above. That's in #154.

To simplify the list and work, I've closed this issue. Please reference #154 in the PR to finish this implementation.

PR #85, opened 2015-10-17
Adding sample data to support development/testing of the notification bits

MisterJames:
First commit is "wall of red", no changes to sample data, just reformatted.

Second commit is the new bits, adding new data for volunteers to be part of one of the activities to support notification development/testing


MisterJames:
In situations like this:

image

I am seeing frequent errors that I believe are related to the async bits, they are duplicate keys and similar errors. I have fixed it in my other branch (for #45) by appending `.Wait()` to the async calls and will check those changes in there.

Issue #59, opened 2015-08-09
Unit Tests: Improve Unit Test Coverage


mheggeseth:
The unit test project AllReady.UnitTest only has a 3 tests and only tests the ActivityApiController. Write more tests and improve the robustness of existing tests.

peteraritchie:
I'm doing this

rockfordlhotka:
I'm also working on this

MKSessions:
We know we don't have the test coverage that we want.  We'd welcome anyone adding more unit test coverage.

ericbrumfield:
Looking to pitch in, I'll start with ResourceAPIController to get my feet wet.

Kritner:
This is probably silly as far as missing a setting or something, but when I run "Analyze Code Coverage -> All tests" in the web only solution (I'm trying to easily identity areas that still need test coverage)  it only shows that 33.33% of code is not covered.  The strange this is, that 33% is made up of only a total of 12 blocks (4 missing).  The only assembly showing up in the first tier of the drop down is AllReady.Models.dll.  Any ideas why my code coverage is clearly not checking out the code coverage of the entire solution?

jonhilt:
Hello. I've been discussing with @BillWagner and @RichardCampbell  a proposal to to host a "hackathon" day at my company (codeweavers.net) to work on AllReady.

We're still in the planning stages but looking at the kind of areas we could work on for the day this seems like a gentle introduction for our developers (contributing to open source is new to most of them) and something which would add value.

Once we settle on a date, we'd be happy to take any gaps in the testing and look to plug those. Failing that, if there are any obvious areas/features for us to to look at (which would be of most benefit to the project) then please do let me know.

Thank you

BillWagner:
@jonhilt I think that is an awesome idea. Let us know when you have the date finalized.

DocHammoc:
I have a passion for unit tests ... so this may be a good place for me to jump in.

BillWagner:
@DocHammoc That would be perfect. Note that some improvement has happened here since this issue has been created, but we're still not where we want to be.

eisenpony:
Some people are having trouble running code coverage, so here is a spreadsheet with current numbers.
[Coverage.xlsx](https://github.com/HTBox/allReady/files/139273/Coverage.xlsx)

mgmccarthy:
[...]

PR #95, opened 2015-10-21
Align unit test fix with beta8 updrade

dpaquette:
Due to order of PRs, unit test project was not compiling after beta8 upgrade was merged


dpaquette:
Related to #93 and #59

PR #96, opened 2015-10-23
Added Scope for email to Facebook Authentication. 

bcbeatty:
Although it's not pulling the email from facebook as it should. 
Not sure if its the beta software or my code.


dpaquette:
I don't think we will want your publish profile checked in to master

bcbeatty:
It keeps adding it, even though I thought Properties* is set is .getignore

dpaquette:
Can you push a new commit deleting that file?

PR #97, opened 2015-10-23
Database Creation and Sample Data Cleanup

shawnwildermuth:
I've changed the way we're specifying the schema, cleaned up initialization of hte sample data, and forced migrations to happen on real databases (e.g. azure).


dpaquette:
Looks good to me 
:shipit:

Issue #36, opened 2015-08-08
5.1.1.3.5	Organization admins can list multiple volunteer skills/resources needed


tonysurma:
including optional date ranges and locations {‘skill/resource A’ needed [from ‘x’ to ‘y’ date range] [in ‘z’ location scope]}.  This listing will allow volunteers to later connect to the organization and offer to volunteer in delivery of the campaign as well as potentially be a data feed consumed by other third party apps that display volunteer needs.

DevUnit1:
I'll try this one

DevUnit1:
The skills/resources being listed - are they associated with a campaign, activity, or task?

tonysurma:
Thanks, at the activity level
________________________________

DevUnit1:
I discussed this feature with Jim and Richard.  For now, we're going to defer the resource part of this feature.  Below is my understanding of the skill system we'll be implementing - please feel free to comment on, correct, question, or add to any part of it!

* Skills will be hierarchical (i.e. Medical -> MD -> surgeon, Medical -> CPR Certified)
* Skills will exist in pre-defined templates
 * The administrator will be able to import the template into a campaign
 * The administrator will be able to search for templates
 * The administrator will be able to customize the templates
 * The administrator will be able to select the lexicon for the templates (such as using the UN names or not)
* Skills will be attributed to users
 * The administrator can add skill to user
 * The user can add the skill to themselves
   * Possible verification/override function?  TBD
* Skills will be assignable to Activities
* Skills will be assignable to Tasks

mheggeseth:
I'll take over ownership of this issue.

mheggeseth:
Can somebody clarify the difference (if any) in this context between a skill and a resource? In @DevUnit1's comment, "resources" are not mentioned. I want to make sure I'm not missing something.

tonysurma:
@mheggeseth - per discussion we are focusing on skills only and delaying any work on resources to a later date.  Therefore you are not missing anything and the clarification from @DevUnit1 is complete (until we bring resources back into it at a later date).  Thanks.

mheggeseth:
@tonysurma can you clarify what "Skills will exist in pre-defined templates" means? Is this like having "skill presets"? E.g. My organization runs a campaign each semester during finals to deliver pizza and wings to hungry college students and each time we need the same skill sets to populate that from a template.

Or does "template" mean something else in this case? Can you give an example of how a "template" would be used?

mheggeseth:
[...]

PR #101, opened 2015-10-24
WIP #36 skill associations

mheggeseth:
Could use a sanity check before proceeding with current plan for #36. See my most recent comment on the issue for a brief summary. I'll need to re-base pending the merge of Shawn's PR #97 refactoring the EF models, but I didn't really know what I was doing in EF anyway so that will be helpful.


dpaquette:
Maybe @tonysurma can chime in on the requirements perspective.

From a code perspective this looks ok to me. Don't forget to add a database migration for the changes to the model.

tonysurma:
@dpaquette did a quick requirements glance at what @mheggeseth wrote on the other issue and looks good to me

mheggeseth:
@dpaquette Thanks. How do I add a migration?

dpaquette:
It is command line based. Type the following from the AllReadyApp/Web-App/AllReady folder. It should give you an example.

    dnx ef migrations add --help

Note that you will need to change your config.json to use a SQL database instead of InMemory.

Issue #104, opened 2015-10-24
Simplify Permissions Checks


dpaquette:
Permissions checks can be simplified by checking against the ClaimsUser instead of requiring multiple calls to the UserManager. The ClaimsUser is always available from every controller and should have the information we need in order to check permissions. In general, very few controllers should have a dependency on the UserManager<ApplicationUser>.

dpaquette:
Going to tackle this one now

PR #105, opened 2015-10-25
Simplify authorization model

dpaquette:
Would like some feedback on this approach so far. This is just a first pass and I plan to do some more work on this.

Goal was to simplify the way we check authorization in terms of tenant admin / site admin users. Previously we were making a lot of unnecessary calls to the user manager. New approach is claims based and should result in much fewer calls to the database.

1) Removed the ApplicationUser -> Tenant relationship in the model. Instead of this, use the the .GetTenantId() extension method to get the current users tenant id.
2) Removed a lot of unnecessary checks to get the current user from the UserManager. Removed dependency on UserManager from a couple of controllers.

Related to #104


dpaquette:
Ping @MisterJames for feedback ;-)

Will plan to add some form of unit tests around this soon. Looking for a sanity check before proceeding.

MisterJames:
@dpaquette This is looking really good. :+1: Love the simplification in the controller. The sample data is WoR, but I think I see what you're doing there (looks like the only changes there were related to claims in creating users?). Build's failing, but otherwise certainly the right direction.

dpaquette:
Build failure will be resolved once PR #103 is merged.
Not sure what happened to SampleDataGenerator file there but yeah, the only change is the way claims are now used to setup the tenanted for the user.

dpaquette:
Ready for review here. 
-Removed UserManager from any controllers that didn't actually need it (this will make unit testing easier)
-Changed some data access methods to use UserId instead of requiring a full ApplicationUser instance.
-Added unit tests for the claims extensions used to check user types and get the tenant id

MisterJames:
All-in-all this looks really good. :+1: 

I would like to know more about whether or not "TenantAdmin" can be thought of as a role in the application, because then we could use the role claims/attributes to simplify this even a little more. But this is a great PR.

@tonysurma or @BillWagner would you have some input here? You can apply multiply role claims to a user, so even if users have to be multi-hat, I could see this working, again...without the background on the full intentions of what a TenantAdmin should be.

dpaquette:
From going through the code, yeah I think UserType (SiteAdmin and TenantAdmin) might be better suited as a Role claim instead of a custom UserType claim.

MisterJames:
@dpaquette Okay sounds good. Opened as a separate issue #107.

MisterJames:
@BillWagner IMO this one is ready to go.

:shipit:

tonysurma:
@misterjames moved discussion on input to new issue

mheggeseth:
@dpaquette Little problem with this one. The logic in `ActivityAdminController.UserIsTenantAdminOfActivity` was changed to prevent SiteAdmins from being able to edit activities if the SiteAdmin doesn't have a TenantId. I think the else block should probably return `User.IsUserType(UsetType.SiteAdmin)` instead of false.

    private bool UserIsTenantAdminOfActivity(Activity activity)
    {
        int? tenantId = User.GetTenantId();
        if (tenantId.HasValue)
        {
            Tenant tenant = _dataAccess.GetTenant(tenantId.Value);
            return User.IsUserType(UserType.SiteAdmin) || tenant.Campaigns.Any(c => c.Id == activity.CampaignId);
        }
        else
        {
            return false;
        }
    }

dpaquette:
Thanks for noticing that @mheggeseth. I'm working on this controller now so I will fix this as part of my next PR. I'm hoping to find a better way to write these types of authorization so the intention is clear and so we can unit test them properly.

dpaquette:
Actually I see this can be simplified even further as the Activity itself has a tenant ID.  I am guessing we should be checking against that instead.

mheggeseth:
Looks good. More deletions the better. :smile:

dpaquette:
Okay, this will be included in my next PR. If it is blocking you now you can grab the change from this branch: https://github.com/dpaquette/allReady/tree/feat_admin_activities

mheggeseth:
It's fine. I can work around it by logging in as the tenant admin.

PR #108, opened 2015-10-26
Removed MobileViewModels project from AllReadyApp Solution

bcbeatty:
Removed MobileViewModels project from AllReadyApp Solution


dpaquette:
I missed this solution file when I removed the project. Thanks for catching that @bcbeatty 
:shipit:

PR #110, opened 2015-10-27
Add a prerequisite for ASP.NET and Web Tools 2015

garrmark:
Starting with beta7 of ASP.NET one must download and install ASP.NET and Web Tools 2015 to be able to build the solution in VS 2015.  This took me a few hours to find, I'm hoping to save others some time.


dpaquette:
This is definitely good to have on the readme. We are actually on Beta8 now. Can you update the PR to reference beta8? Download is here http://www.microsoft.com/en-us/download/details.aspx?id=49442

garrmark:
Thanks, I updated it to reference the beta8 install.

Issue #16, opened 2015-07-20
As an administrator, I want to be able to define actionable items for a given campaign, so people can volunteer for those items


tonysurma:
This needs to be fully documented out and split into multiple enhancement issues for what is not yet implemented fully.

This user story covers both the definition of activities and tasks. 

Each campaign is split into a set of activities that are tied to a certain geographical location. Each activity, in turn, is split into a set of tasks that the administrator can assign to volunteer that have applied for the activity.

dpaquette:
I am going to start looking in to the 'Activities' portion of this. Any additional information you can provide @tonysurma

dpaquette:
I'm working on the Create / Edit Activities views right now. Currently, we have a single page that lists out all Activities for all Campaigns and all Organizations. From a UX perspective, this seems a little clumsy. I would expect that a user would start by selecting a particular Campaign before trying to Create or Edit an activity. 

I propose removing the page that lists *all* activities and move the Create / Edit actions to an Admin Campaign Details page.

Would like some feedback from @tonysurma and/or @BillWagner before proceeding.

tonysurma:
@dpaquette absolutely right.  CRUD of activities is scoped within an organization and tasks within activities for that matter.

A logical example that is made up is like this:
Organization A has a campaign to lessen the impact of fire related disasters in individual homes.  They run 3 activities: an awareness fair/booth/whatever on a given set of days at a given spot, a canvassing activity to get to all the homes in neighborhood N and see who needs/wants smoke dectors, and an activity to install smoke dectectors in that home.  In the last example there will be tasks for sets of days and locations for installs.  Volunteers can volunteer for activities and then admins can assign them to individual tasks or admins can directly assign (which is really a request for assignment as a volunteer can of course say no) volunteers to tasks as they may already have awareness of different volunteers and which tasks are most likely good for them.

dpaquette:
Okay great. I will start by scoping activities to a particular campaign. Once that is done we can tackle scoping tasks to a particular activity instead of a single list of all tasks.

dpaquette:
Taking a look at Task Admin now

dpaquette:
I have some ideas for restructuring the admin pages a little for better flow. Rather than having a grid of Campaigns with all the actions at the end, I am envisioning a grid of campaigns where you click on the campaign to see the details. From that details page you would see all the details and a list of Activities. You would then be presented with options to Edit the details, Add Activities, etc.  The same approach would be used for Activities and Tasks.  

[...]

PR #114, opened 2015-10-29
Improvements to Activity CRUD

dpaquette:
Part of implementation for #16.

Scoped Activity List / Create / Edit actions to a particular Campaign. View activities for a campaign by selecting "Activities" from the Admin -> Campaigns list. Since the activity is scoped to a particular campaign, the user no longer needs to select a Campaign / Tenant when creating an Activity.

Moved to Tag Helpers for the input controls were possible.

Lots of work to do on the UX for these views still but this should give us a better starting point.


dpaquette:
Would be great if someone could review this before I get started on the same thing for Tasks

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ActivityAdminController.cs
```diff
@@ -57,17 +57,31 @@ public override ViewResult View(string viewName, object model)
         }
 
         // GET: Activity
-        public async Task<IActionResult> Index()
+        [Route("Admin/Activity/{campaignId}")]
+        public IActionResult Index(int campaignId)
         {
-            return await Task.Run(() => View(_dataAccess.Activities));
+            Campaign campaign = _dataAccess.GetCampaign(campaignId);
```
mheggeseth: Want to check for null before proceeding?

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ActivityAdminController.cs
```diff
@@ -97,25 +111,46 @@ public override ViewResult View(string viewName, object model)
         }
 
         // GET: Activity/Create
-        public IActionResult Create()
+        [Route("Admin/Activity/Create/{campaignId}")]
+        public IActionResult Create(int campaignId)
         {
-            return View();
+            Campaign campaign = _dataAccess.GetCampaign(campaignId);
```
mheggeseth: Null check?

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ActivityAdminController.cs
```diff
@@ -57,17 +57,31 @@ public override ViewResult View(string viewName, object model)
...
+        public IActionResult Index(int campaignId)
         {
-            return await Task.Run(() => View(_dataAccess.Activities));
+            Campaign campaign = _dataAccess.GetCampaign(campaignId);
+            if (!UserIsTenantAdmin(campaign.ManagingTenantId))
+            {
+                return HttpUnauthorized();
+            }
+
+            ViewBag.Title = "Activities for " + campaign.Name;
```
mheggeseth: Do you need to set the Title here AND in the view?

joelhulen: Since it's a ViewBag, it's essentially setting it on the view. I think this is a good opportunity to propose setting up a base ViewModel that has standard properties for things like Title, Subtitle, etc. It may just be a matter of personal preference, but I've never been a big fan of the ViewBag, due to its lack of type safety and lack of discoverability.

mheggeseth:
I have no idea about the Tag Helpers, but everything seems to work. I left a few pedantic comments you could address, but otherwise I found nothing obvious. Good move on the global bootstrap datepicker.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ActivityAdminController.cs
dpaquette: I had intended to remove this from the controller. Will fix

dpaquette:
Addressed comments from review.

PR #118, opened 2015-10-30
change tenant to organization in html tags

SteveStrong:
changes to html only,  mostly verifying that I can contribute and understand the workflow


dpaquette:
Looks good :shipit:

PR #124, opened 2015-10-31
Skills CRUD Views/Controller

### LINKED ISSUE #111: Skills CRUD views
Skills (see #36) need CRUD views
* [x] Skill CRUDs should only be visible/editable by SiteAdmins
* [x] Create/Edit views should prevent circular references (child cannot be an ancestor of its parent skill)
* [x] Parent skill dropdown should have an "empty" option to denote that the skill has no parent
* [x] Warn when deleting a skill with children

mheggeseth:
Fixes #111. See issue for features implemented.


AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SkillAdminController.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public IActionResult Create(Skill skill)
+        {
+            if (!User.IsUserType(UserType.SiteAdmin))
+            {
+                return new HttpUnauthorizedResult();
+            }
```
dpaquette: This is not necessary because you have the Authorize("SiteAdmin") attribute on the controller.  You can remove this form each action method

mheggeseth: Cool. Update on the way.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SkillAdminController.cs
```diff
@@ -0,0 +1,101 @@
...
+                return RedirectToAction("Index", new { area = "Admin" });
+            }
+            return WithSkills(View(skill));
+        }
+
+        public IActionResult Delete(int? id)
+        {
+            if (id == null)
+            {
+                return new HttpStatusCodeResult(404);
```
dpaquette: I generally prefer `return HttpNotFound();` for 404s as it is a little more clear

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SkillAdminController.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            if (ModelState.IsValid)
+            {
+                _dataAccess.UpdateSkill(skill);
+                return RedirectToAction("Index", new { area = "Admin" });
+            }
+            return WithSkills(View(skill));
+        }
+
+        public IActionResult Delete(int? id)
```
dpaquette: Curious why nullable? Seems like `int id` would be better then the model binder is responsible for checking null and we don't have to worry about that inside the action method.

bcbeatty: It was preventing the 1st record from being inserted, since there isn't a FK back to an existing record.

dpaquette: @bcbeatty ??? Not sure that applies to the id passed in to the delete method.

bcbeatty: Sorry wrong thread

BillWagner:
Is this ready? I'm not sure from the comment threads.

dpaquette:
There are a few things I think could be cleaned up but they are nit-picky. I think we can check this in and address those comments separately if it keeps things moving along.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SkillAdminController.cs
mheggeseth: I can dig it.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SkillAdminController.cs
mheggeseth: No reason for nullable. Just the dregs of copy/paste. Makes sense to change it.

Issue #126, opened 2015-10-31
Unhandled InvalidOperationException when bad data submitted to Create Activity form


ksk100:
When creating an activity for some campaign (at `/Admin/Activity/Create/[someCampaignIdHere]`), if the validation checks fail, then an unhandled `InvalidOperationException` is produced.

This is because it can't find the appropriate view, because the create page is now using the edit view, but the action which handles the post doesn't specify this.  Just needs `"Edit"` adding as first parameter when returning the view.

dpaquette:
I can fix this today

ksk100:
Cool.  I noticed another couple of things in that action method too, one of which becomes visible after the "Edit" fix is made: if you submit bad data (e.g. a bad date which fails validation) and also don't create any required skills, then when the create page is reloaded an `ArgumentNullException` is thrown by line 138:

`@Json.Serialize(Model.RequiredSkills.Select(rs => new { Id = rs.SkillId }))`

in `Edit.cshtml`.  That's the line which produces the JSON for KO to populate the skills list.  On first load of the Create page the relevant action method creates an empty list, but when reloading after a failed post, `Model.RequiredSkills` comes from the post and ends up `null` if the user hasn't yet added any skills.

The second thing was that this action method isn't checking either that the campaign actually exists nor that the current user is an authorised manager of the tenant.

dpaquette:
Yeah a lot of these problems are because we are binding directly to the model class. I'm going to clean that up to avoid these types of issues

BillWagner:
Fixed in PR #128

PR #128, opened 2015-10-31
Fixes crashes after validation fails for Create Activity

dpaquette:
Resolves issue #126

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Models/Activity.cs
+++ b/AllReadyApp/Web-App/AllReady/Models/Activity.cs
@@ -7,6 +7,13 @@ namespace AllReady.Models
 {
     public class Activity
     {
+        public Activity()
+        {
+            Tasks = new List<AllReadyTask>();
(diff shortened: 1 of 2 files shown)


AllReadyApp/Web-App/AllReady/Models/Activity.cs
```diff
@@ -7,6 +7,13 @@ namespace AllReady.Models
 {
     public class Activity
     {
+        public Activity()
+        {
+            Tasks = new List<AllReadyTask>();
```
BillWagner: Since we're using all the latest, I'd like to see us use the C# 6 initialization syntax:
e.g.
Tasks { get; set; } = new List<AllReadyTask>();

dpaquette: Hmm. That is one C#6 feature that I'm not totally sold on yet.  I will change it though

PR #130, opened 2015-10-31
Change SampleData Generator to save in a persisted local database

bcbeatty:
Change SampleData Generator to save in a persisted local database
Change Skill.ParentSkillId to allow nulls, this was preventing adding a skill

### DIFF
--- a/AllReadyApp/Web-App/AllReady/DataAccess/SampleDataGenerator.cs
+++ b/AllReadyApp/Web-App/AllReady/DataAccess/SampleDataGenerator.cs
@@ -1,464 +1,498 @@
 ﻿using Microsoft.AspNet.Identity;
 using Microsoft.Framework.Configuration;
-using Microsoft.Framework.DependencyInjection;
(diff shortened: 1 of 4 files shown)


mheggeseth:
@bcbeatty Thanks for doing this! Just an FYI: to avoid the dreaded "wall-of-red" we're normalizing on Tab==4 spaces line indentation (the VS default).

bcbeatty:
That's my settings Tabs==4, keep tabs

mheggeseth:
We are using "Insert spaces" instead of "Keep tabs". :smiley:

mheggeseth:
I noticed a couple Skill CRUD fixes that'll need to be made with this. This fixes creating/editing a skill with no parent as well as properly orphaning child skills when their parent is deleted.  https://github.com/mheggeseth/allReady/commit/942d45a5edfc543dc2adfc7979b52591fbf3ec0f Feel free to pull them into your branch.

PR #150, opened 2015-11-04
Activity Create/Edit/File Upload page: converted to view models, both forms now validate separately

### LINKED ISSUE #48: Bombs on activity image upload if no file selected
On the Admin Edit Activity screen, if I load an activity with no image file and click the Upload button, the application throws a null reference exception.  It should instead ignore or not allow an image upload with no file selected.

ksk100:
For review:

This converts the Activity Create/Edit/File Upload page to use ViewModels: one parent, and one child each for Create/Edit form on one hand and File Upload form on the other.

This allows for separate validation of the two forms, which differ in their `Required` fields as noted in issue #48.

It fixes #48.

Image upload to Azure works if you alter `AzureStorage` in `config.json` appropriately.


dpaquette:
Thanks for the submission. 

Unfortunately I am in the middle of a big refactoring of this whole area of the code. I don't think that the changes you have made here will fit in with the direction we are planning to take this.

ksk100:
Okay, no problem.  I did ask about the approach on issue #48, but never mind.

Which areas is your refactoring encompassing (just so I can try to avoid this happening again)?

PR #157, opened 2015-11-06
Forces culture to en-US to give consistent date handling behaviour

### LINKED ISSUE #119: Date validation in non-US cultures
Hi everyone,

I'm new to the project; I've enjoyed investigating the source so far, and I hope I'll be able to contribute something useful.

I recently came across an issue with the date pickers on the website: the server side validation occurs according to the culture that the web server runs in.

However, the jQuery UI date picker used across the site is currently hard coded to use US date format (in `Views/Shared/_Layout`).

This means that when server is run in cultures with different date format, like the UK (where I am), the server side validation fails (the default date format here being `DD/MM/YYYY`).

Perhaps the most straightforward way to address this, at least for the time being, is to force the server to run with culture set to `en-US`.  That would make the site operational for those of us not in the US with just a few lines of code in `Startup.cs`.

Does that sound like a sensible way forward?

Thanks

Simon

ksk100:
For review.

Fixes #119.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Startup.cs
+++ b/AllReadyApp/Web-App/AllReady/Startup.cs
@@ -21,6 +21,8 @@ using Autofac.Framework.DependencyInjection;
 using MediatR;
 using Microsoft.AspNet.Authorization;
 using Microsoft.Dnx.Runtime;
+using System.Threading;
+using System.Globalization;
 
 namespace AllReady
 {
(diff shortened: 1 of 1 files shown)


dpaquette:
I haven't tested this but I think the aspnet5 way of limiting to a particular culture would be by adding the following to the Startup.Configure method:

```
var usCultureInfo = new CultureInfo("en-US");
var requestCulture = new RequestCulture(usCultureInfo);
app.UseRequestLocalization(new RequestLocalizationOptions
{
    DefaultRequestCulture = requestCulture,
    SupportedCultures = new List<CultureInfo>(new []{ usCultureInfo }),
    SupportedUICultures = new List<CultureInfo>(new [] { usCultureInfo})
}); 
```

@ksk100 Would you like to test that out and see if it also solves the problem you were having? If that works it would give us a path to adding more cultures when we are ready.

ksk100:
Of course, I'll have a look.

PR #158, opened 2015-11-06
Improved solution to setting US culture

ksk100:
David, yes that worked, so here is a PR for that improved solution.

For review; improved fix for #119 as suggested by @dpaquette.

- Simon


dpaquette:
Looks good!  :shipit:

PR #191, opened 2015-11-07
Fix for last PR where I screwed up the config file

shawnwildermuth:
Lost a change in the config file.


dpaquette:
Looks good :shipit:

PR #211, opened 2015-11-07
Tasks allow volunteers to see assigned tasks

### LINKED ISSUE #40: 5.1.2.1.1	Allow volunteer to see tasks assigned to them
As an authenticated volunteer, I can:
 - [ ] See the list of tasks assigned to me
 - [ ] I can navigate to or make use of the task accept/reject feature #41

bcbeatty:
A Volunteer will only see the tasks are assigned to them from an Administrator. Anonymous users will not see tasks. 
If an admin views is on the volunteer view(non admin area) in P1 they will be treated like a normal user.

Closes #40


dpaquette:
Sorry @bcbeatty .  You're going to need to do the merge dance :dancers: again since another PR was merged in

PR #227, opened 2015-11-07
Logging instead of writing to a queue

MisterJames:


AllReadyApp/Web-App/AllReady/Services/FakeQueueWriterService.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace AllReady.Services
+{
+    public class FakeQueueWriterService : IQueueStorageService
+    {
+        private readonly ILogger _logger;
+
+        public FakeQueueWriterService(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger(typeof(FakeQueueWriterService).FullName);
+        }
```
rynowak: DI'ing an `ILogger<T>` is a shorthand for this code

AllReadyApp/Web-App/AllReady/Services/FakeQueueWriterService.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public void SendMessage(string queueName, string message)
+        {
+            _logger.LogInformation($"Message to {queueName} queue: {message}");
```
rynowak: `LogInformation("...", queueName, messageName)`

AllReadyApp/Web-App/AllReady/Startup.cs
```diff
@@ -104,20 +104,34 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
 
         private IContainer CreateIoCContainer(IServiceCollection services)
         {
+            var loggerFactory = new LoggerFactory().AddConsole(LogLevel.Debug);
+            services.AddSingleton(x => loggerFactory);
+
```
rynowak: Don't need this. Do this instead:
```
            loggerFactory.MinimumLevel = LogLevel.Verbose;
            loggerFactory.AddConsole((category, level) =>
            {
                if (category.StartsWith("Microsoft."))
                {
                    return level >= LogLevel.Verbose;
                }

                return true;
            });
```

PR #301, opened 2015-11-09
Prompt for uploading image on Activity/Campaign pages set to 'Browse for image'

### LINKED ISSUE #278: Admin Edit Activity - Image Upload vs URL
For the admin edit activity form, the Image field is labeled as "ImageURL", but the field is actually a Browse to upload file. This should be browse and upload across all image fields.

image

ksk100:
For review.

This fixes #278.

It also does the same for the Campaign page.

And it adjusts the layout a bit so that the 'Browse' prompt is always inline with the file upload control, and not the image itself.


dpaquette:
Looks good.

PR #305, opened 2015-11-10
introduction of systems.js module loader to project and typescript to modularize mapping js

SteveStrong:
I focused on the campaigns/map view to verify the systems.js and typescript would could be integrates to help us manage JavaScript global namespace,   I would like to remove more script tags from the cshtml pages, but I think this is a good start.    

I also think that overtime we should probably more to leaflet for mapping and rendering geojson (api are better) and having all mapping concerns in mapping module should help with this.

I would like comments from @shawnwildermuth  and @dpaquette  if possible since this is the first time I have implemented a typescript solution.


dpaquette:
Also pinging @stimms

stimms:
The code looks reasonable enough. My comments are 

1. Do we need to check in the .js files? Once we're using typescript they are build artifacts
2. Do we really need a module loader? If we were building a SPA I would see the reason but with the model that allready uses we could just as easily concatenate all the js into a single file which would be cached and not reloaded for each page. People would pay the js cost up front and not as they moved through the app. Initial page load would be worse but then subsequent pages would be better. It seems to add a bunch of complexity that isn't needed.

dpaquette:
1. I think initially we were planning to check in the .js files with the intention of NOT checking them in soon (once we get our gulp file ironed out). We may wait until after the November MVP milestone for this.
2. I was also thinking that a module loader might be more than we need for the size and type of application. Concatenating all our JS files to a single file should be sufficient and will likely provide the best performance overall.

SteveStrong:
I am all for doing the gulp transforms to simplify loading modules, just remember one of your goals we to make sure out JavaScript was namespace, and that name spacing was easy.  It is also likely modules will be composed of standard functionality in the future.  Lets integrate this PR, and then you can take a crack of some  gulp magic to create a 'debuggable' single source file.   Also move the TS files outside of wwwroot (where that belong)

mheggeseth:
Don't commit `system.js` and `system-polyfills.js` as source. Pull the package in using the bower.json file and reference the files using the `wwwroot/lib/system.js/dist` directory.
`"system.js": "0.19.5"`

SteveStrong:
I added changes as suggested by the comments above,  bower and system.js  etc.   I also refactored all JS involved in mapping into a single  mappingtools.ts file   and adjusted all the .cshtml files to use this new module.   with the integration of this pull request we can hopefully 
1) simplify the integration of maps going forward.
2) lay the foundation for move js moving to typescript
3) restructure project directories to pull TS from wwwroot and have the generation of ts->js part of the gulp build process

dpaquette:
Is test.html intended to be checked in?

SteveStrong:
yes for now test.html is part of the checkin until I/we resolve how modules should work.  I will delete it in the future when I extablish a pattern for migrating other JavaScript files

stimms:
It seems like this migration to ts is going to be somewhat involved. Could we merge this PR to a branch and then @SteveStrong and I could pair on sorting the whole ts/js story out? I don't like the idea of having something like test.html in the master branch.

SteveStrong:
the core of this PR is to sync move mapping into it's own module.  and then have every view  that uses a map, use the same module.  things list test,html are to have a discussion on how to use typescript to migrate all the other js  and <script></script> tag inline js.  If merging the PR into a branch help us get rolling, I am all for it.  looking forward to pair with @stimms

BillWagner:
I created a TypeScriptSpike branch tonight. @SteveStrong Can you retarget your PR to that branch? I'll merge it it once you do that.

BillWagner:
@SteveStrong should this be closed, now that the other PR is in its branch?

stimms:
We can close it but if we did I'd open another one to track our js -> ts work. Up to you.

MisterJames:
Deleting this PR (long tail on merge/resolution) but we can tackle it again...this scope is the size of a code-a-thon.

PR #318, opened 2015-11-11
Remove font size from active. Fixes #315.

### LINKED ISSUE #315: Font size of login-status links should not change on click
The links in the "login status" section of the header (upper right corner) change font size when you click on them, and should not.

Hover:
image

Click:
image

This is because we are not being very careful about our CSS selectors. The following selector in `site.css` applies the smaller font-size to this section: `.navbar .login-status > li > a`. This has the exact same specificity (https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) as the selector that sets the larger font size `.navbar-default .navbar-nav > li > a`, but because the `login-status` rule appears after the other rule in `site.css`, it wins. *Until* you click on it. This matches the `.navbar-default .navbar-nav > li > a:active` selector whose specificity is exactly 1 greater, so the larger font size now wins.

**Fix:** up the specificity of the `login-status` selector and add a `:active` rule to it as well just for good measure.
```CSS
.navbar .navbar-nav.login-status > li > a,
.navbar .navbar-nav.login-status > li > a:active {
    font-size: 1.1em;
}
```

stimms:
There should be no reason to override the font size in the active state. We should never be changing the fontsize in active or it look jaggidy, just allow the existing font size to cascade.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/wwwroot/css/site.css
+++ b/AllReadyApp/Web-App/AllReady/wwwroot/css/site.css
@@ -136,7 +136,6 @@ Navigation
 
 .navbar-default .navbar-nav > li > a, .navbar-default .navbar-nav > li > a:active, .navbar-default .navbar-nav > li > a:visited, .navbar-default .navbar-nav > li > a:focus {
     color: #fff;
-    font-size: 1.4em;
 }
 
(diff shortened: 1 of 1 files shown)


dpaquette:
:shipit:

PR #307, opened 2015-11-10
Tenant contact info copy to campaign 

bcbeatty:
* Added tables Contact, TenantContact, and FK from Tenant to Location
* Modified Tenant Controler  use Query/Command
* Added new Location Edit and Display Templates 

WIP for #249 

To do: 
* Campaign FK to CampaignContact and Location
* Modify Campaign EditCampaignCommandHandler & CampaignDetailQueryHandler
* Modify UI to copy Tenant Contact and Location

PR #322, opened 2015-11-12
Restore incremental migrations after #307

mheggeseth:
#307 accidentally deleted the incremental migration `ActivitySignupDetails`. This broke http://allready-d.azurewebsites.net (and my local db, too). This PR restores `ActivitySignupDetails` and appropriately updates the `AddTenantContactCampaignContactContact` migration.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Migrations/20151108180202_ActivitySignupDetails.Designer.cs
+++ b/AllReadyApp/Web-App/AllReady/Migrations/20151108180202_ActivitySignupDetails.Designer.cs
@@ -1,0 +1,590 @@
+using System;
+using Microsoft.Data.Entity;
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Migrations;
+using AllReady.Models;
+
+namespace AllReady.Migrations
+{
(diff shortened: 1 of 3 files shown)


BillWagner:
@mheggeseth Can you think of a good way to catch this in our automated builds? It's happened a couple times, and I've missed it in reviewing the PRs as well.

mheggeseth:
If we had access to a SQL instance from the build server, we could write a unit test that runs the migrations (like `Startup.cs` does) against a dynamically-created empty database. If the test runs without exception, then it passes and the DB can be dropped. That would at least detect the case where the migrations themselves are wrong or conflict with each other.

However, that test would not have detected this mistake. The problem here was that the migrations would have worked when creating the database from scratch, but not when incrementally updating from the last set of migrations. This is simply because one of the migration scripts was deleted and the new one absorbed its changes meaning that when you tried to apply the new one, it was trying to add some columns that already existed in the database. You probably just have to watch out for this manually. Deleted migrations should be a red flag that something bad might happen.

This suggested test approach also wouldn't detect things like forgetting to do migrations entirely or having migrations that are not compatible with the code model (i.e. they are out of date). For that, we would need some really sophisticated integration tests that do things like inspect the DbContext and validate that you can insert sample data that complies with the model.

PR #323, opened 2015-11-12
this should move my typescript changes into the typescxriptSpike branch

SteveStrong:


SteveStrong:
so did I do this right?

PR #324, opened 2015-11-12
Adding version from git. Fixes #314.

### LINKED ISSUE #314: some indication of which commit current build is from?
I have done this on other projects where part of the build script is generating a file with the commit hash in it that is shown in the user interface of the app so that you know which build you are working on when you have an issue.  Anyone have a quick idea how to do that in our Kudu/asp.net5 beta environment?

Labelled it lower priority as we only have two builds (-d and -mvp urls for now) but would be helpful when you are researching bugs to know what set of commits the issue happened IMHO

stimms:
Pulling in the revision number from git and chucking it at the bottom of every page via the version.json file.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Startup.cs
+++ b/AllReadyApp/Web-App/AllReady/Startup.cs
@@ -27,6 +27,7 @@ namespace AllReady
             // Setup configuration sources.
             var builder = new ConfigurationBuilder()
                 .SetBasePath(appEnv.ApplicationBasePath)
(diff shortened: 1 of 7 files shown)


mheggeseth:
How does this work? When I run it locally, it just shows the string literals that you put in `version.json`.
image

stimms:
I missed the update to the appveyor file. Locally you're just going to get the GITVERSION string or rather whatever is in version.json.

Issue #329, opened 2015-11-12
Logging in using Administrator login form, but with standard user credentials, shows error but still logs user in


ksk100:
To reproduce, just use one of the sample standard users on the admin login form.

You'll see "Invalid login attempt" but you will still be logged in and if you refresh you'll see standard menus.

dpaquette:
I will fix this

PR #332, opened 2015-11-13
Avoid confusion when regular users attempts to login as admin

dpaquette:
Instead of showing error when regular users attempt to log in from the Admin Login page, just redirect them to the main page. They won't have access to admin pages anyway. This will avoid confusing any potential volunteers if they accidentally pick the wrong long in page.

Resolves issue #329 

In the future we may want to merge the 2 login pages. Not sure we need separate logins.


stimms:
Quite a few magic strings in there.

BillWagner:
I agree with @stimms that we are using quite a few strings in lots of places.
I'm merging this, and opening a new issue on the strings in the code. That can be a good 'jump in' issue.

dpaquette:
Aren't those just controller and action names?

PR #346, opened 2015-11-14
Removal of extra class 'IRequestHandler<T>'

MisterJames:
- not in use/wrong namespace/no matching signature implementation
- looks like it was added accidentally as part of an earlier commit


dpaquette:
That might have been me smashing the keyboard/ReSharper shortcuts. :shipit:

PR #355, opened 2015-11-15
Fix CampaignImpact to get activities back on the campaign detail page (yes, really)

### LINKED ISSUE #351: No campaign activities are showing on the site
This campaign should show up 'publicly' and it only shows on admin:

Is in the admin list:
https://allready-mvp.azurewebsites.net/Admin/Campaign/Details/8

Should be in this list:
https://allready-mvp.azurewebsites.net/Campaign/8

That is just one example but no activities are showing up at all

mheggeseth:
~~This is a **TEMPORARY** fix to #351. This will get activities to show up again on the campaign detail page but the main problem is that there are some problems with how the CampaignImpact table is defined in the database, but I haven'y fully figured that out yet. This can be merged, but there will be more to come.~~

Closes #351. The relationship between `CampaignImpact` and `Campaign` was messed up causing EF to behave in undesired, weird ways that I don't understand. The SQL schema generated for `CampaignImpact` made `Id` not a private key identity, but a foreign key link to `Campaign.Id`. This meant that the only possible valid `CampaignImpact` Ids are Campaign IDs and EF was having none of it. **Now**, I've added `Campaign.CampaignImpactId` as a foreign key to `CampaignImpact.Id` and `CampaignImpact.Id` is now a private key identity. 

**WARNING** This fix is destructive because it has to burn down and re-create the CampaignImpact table to ensure its `Id` column properly set up as a PKI. I thought I could get this done with incremental migrations, and even though there was a command in the migration that said to set the Id column to be an identity, SQL didn't make it happen. Bankruptcy was the only other way through.

Also fixed a namespace mixup between the `CampaignController` classes.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
@@ -10,7 +10,7 @@ using System.Threading.Tasks;
 using AllReady.Services;
 using AllReady.Extensions;
 
-namespace AllReady.Controllers
+namespace AllReady.Areas.Admin.Controllers
 {
     [Area("Admin")]
     [Authorize("TenantAdmin")]
--- a/AllReadyApp/Web-App/AllReady/Controllers/CampaignController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/CampaignController.cs
@@ -5,7 +5,7 @@ using AllReady.Models;
 using AllReady.ViewModels;
 using Microsoft.Data.Entity;
 
-namespace AllReady.Areas.Admin.Controllers
+namespace AllReady.Controllers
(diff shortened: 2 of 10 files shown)


mheggeseth:
Pinging @dpaquette or @MisterJames for review on my use of the nuclear option here.

BillWagner:
also pinging @tonysurma to determine impact on current private beta.

tonysurma:
Ok with losing campaign impact table contents, not ok with losing campaign table (which isn't what your saying but just stating out loud)

dpaquette:
Not sure why the incremental migration didn't work for you @mheggeseth but I guess we are using beta bits.

This should be fine as long as @tonysurma is okay with losing the campaign impact table. If the deployment goes sideways and we lose more than that, we should be able to rollback with a point in time restore in SQL Azure.

PR #363, opened 2015-11-17
did typescript semi conversion,  I tried to make simple screen casts about the experience

SteveStrong:
how I started

http://screencast.com/t/EPbHbZ8f

getting things to compile

http://screencast.com/t/Vm8X2KPj8

tricks on using external libs.

http://screencast.com/t/DlSlyZzYATdG


AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Activity/Edit.cshtml
```diff
@@ -154,7 +154,7 @@
     <script type="text/javascript">
 
             System.config({
-                baseURL: location.protocol + "//" + location.host + '/ts/'
```
stimms: I would tend to push this stuff down a level. So for instance on my projects I tend to have either a folder per controller with the associated files for views off that controller
image

or I'll inset another level so that I can have things like Controls, Pages, Infrastructure in the scripts directory

image

Obviously each file is namespaced in accordance with the hierarchy. Arranging javascript should be just like arranging "real" c# code.

AllReadyApp/Web-App/AllReady/Views/Campaign/Map.cshtml
```diff
@@ -22,7 +22,7 @@
     var modelCampaign = @Json.Serialize(Model);
 
     System.config({
```
stimms: We seem to be duplicating this system configuration in a lot of files. Perhaps we could centralize it into one file. We could load it in the _Layout

AllReadyApp/Web-App/AllReady/Views/Shared/_LayoutMainPage.cshtml
```diff
@@ -2,6 +2,7 @@
 <html>
 <head>
     <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
```
stimms: This is a good call but maybe should be part of a different PR.

AllReadyApp/Web-App/AllReady/gulpfile.js
```diff
@@ -49,13 +58,81 @@ gulp.task("watch", function () {
...
+
+    var allTypeScript = 'ts/**/*.ts';
+    var libraryTypeScriptDefinitions = './tools/typings/**/*.ts';
+    var sourceTsFiles = [allTypeScript,                //path to typescript files
+                         libraryTypeScriptDefinitions]; //reference to library .d.ts files
+
+
+    return gulp.src(sourceTsFiles)
+        .pipe(tsc({
+            noImplicitAny: false,
```
stimms: We should work towards turning this to true.

AllReadyApp/Web-App/AllReady/gulpfile.js
```diff
@@ -49,13 +58,81 @@ gulp.task("watch", function () {
...
-//http://weblogs.asp.net/dwahlin/creating-a-typescript-workflow-with-gulp
-var tsProject = ts.createProject('tsconfig.json');
 
+//https://www.npmjs.com/package/gulp-typescript
 gulp.task('typescript:Build', function () {
-    var tsResult = tsProject.src() // instead of gulp.src(...) 
-        .pipe(ts(tsProject));
-    return tsResult.js.pipe(gulp.dest('wwwroot/release'));
+
+    var allTypeScript = 'ts/**/*.ts';
```
stimms: To me all these definitions should be sucked up and put into a config object. 

`
paths.typescript = {
    source: 'ts/**/*.ts',
    definitions: 'tools/typings/**/*.ts'
}
`

AllReadyApp/Web-App/AllReady/gulpfile.js
```diff
@@ -49,13 +58,81 @@ gulp.task("watch", function () {
...
+
 });
 
+
+
+//https://www.npmjs.com/package/gulp-typescript
+//http://weblogs.asp.net/dwahlin/creating-a-typescript-workflow-with-gulp
+var tsProject = tsc.createProject('./tsconfig.json');
+
+//gulp.task('typescript:Build', function () {
```
stimms: These all look like good targets, I'm not sure why they're commented out

AllReadyApp/Web-App/AllReady/package.json
```diff
@@ -9,7 +9,9 @@
     "gulp-uglify": "1.2.0",
     "bower": "1.6.5",
     "rimraf": "2.2.8",
+    "gulp-tslint": "3.6.0",
     "gulp-typescript": "2.9.2",
```
stimms: We should take a dependency on typescript and pass it into gulp-typescript so that we don't rely on people having typescript 1.6.X installed on their machines.

AllReadyApp/Web-App/AllReady/wwwroot/js/activityDetails.js
```diff
@@ -1,5 +1,5 @@
-﻿var HTBox;
-(function (HTBox) {
+var HTBox;
+(function (HTBox, $) {
```
stimms: I don't like passing in a variable called `$` on jQuery projects that has a pretty specific meaning. Perhaps `container` instead.

stimms:
.js files and .map.js files should not be checked in. They are build artifacts and should be checked in no more than .dll and .pdb files.

mheggeseth:
There should probably be a `ko.filterableList.ts` file but I don't see it in the commit list.

dpaquette:
One thing to do before this can be merged to ensure the typescript files are compiled as part of the prepublish command in project.json.

```
"scripts": {
    "prepublish": [ "npm install", "bower install", "gulp clean", "gulp typescript:Build", "gulp min" ]
  }
```

dpaquette:
We should also have the gulp watch task setup to run the typescript:Build task anytime a typescript file changes.This would ensure that the wwwroot/js file is always up to date for developers.

Since the files in wwwroot/js are always auto generated we should eventually be removing those from source control. We don't have to do that for this PR but something to keep in mind.

stimms:
Okay can we merge this and then I'll submit follow up PRs? Other that @dpaquette's concerns I think everything else can be deferred.

BillWagner:
Will do!

PR #364, opened 2015-11-17
Simple HTML-enabling fix

### LINKED ISSUE #359: Link not clickable in confirm email message sent
email blech

Also guessed @MisterJames is likely one to fix given he worked on emails

MisterJames:
This is not the be-all-end-all solution. We need to progress (longer term) the templates idea from SendGrid.

- extended models to add HTML field
- updated view models affected by HTML
- updated email confirmation to set HTML
- updated service to push HTML out with email

Fixes #359

### DIFF
--- a/AllReadyApp/NotificationsProcessor/Functions.cs
+++ b/AllReadyApp/NotificationsProcessor/Functions.cs
@@ -47,6 +47,7 @@ namespace NotificationsProcessor
             myMessage.AddTo(emailMessage.Recipient);
             myMessage.From = new MailAddress(from, "AllReady");
             myMessage.Subject = emailMessage.Subject;
+            myMessage.Html = emailMessage.HtmlMessage;
             myMessage.Text = emailMessage.Message;
 
             // Create credentials, specifying your user name and password.
--- a/AllReadyApp/Web-App/AllReady/Services/MessageServices.cs
+++ b/AllReadyApp/Web-App/AllReady/Services/MessageServices.cs
@@ -22,6 +22,7 @@ namespace AllReady.Services
                 {
                     EmailMessage = message,
                     EmailRecipients = new List<string> {email},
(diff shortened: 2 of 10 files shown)


AllReadyApp/Web-App/AllReady/Services/MessageServices.cs
```diff
@@ -22,6 +22,7 @@ public Task SendEmailAsync(string email, string subject, string message)
                 {
                     EmailMessage = message,
                     EmailRecipients = new List<string> {email},
+                    HtmlMessage = message,
```
MisterJames: @tonysurma This is the one that addresses the HTML aspect, the link not rendering in the email.

AllReadyApp/NotificationsProcessor/Functions.cs
```diff
@@ -47,6 +47,7 @@ public static void ProcessEmailQueueMessage([QueueTrigger("email-pending-deliver
             myMessage.AddTo(emailMessage.Recipient);
             myMessage.From = new MailAddress(from, "AllReady");
             myMessage.Subject = emailMessage.Subject;
+            myMessage.Html = emailMessage.HtmlMessage;
```
MisterJames: @tonysurma This is what is passed to SendGrid. We **really** want to think about refining these messages.

MisterJames:
@dpaquette and @tonysurma: End-to-end confirmed:

image

 - Message is queued
 - HTML is set correctly (although the message should be prettied up)
 - Message arrives as HTML and is clickable
 - Controller picks up and processes confirmation as sent to user

image

dpaquette:
This looks good. As @MisterJames mentioned, we do need to refine the actual messages we are sending but at least this gives us the structure we need to send both Html and Plain Text emails.

:shipit:

tonysurma:
Excellent, thanks

PR #380, opened 2015-11-18
Correcting reset password from admin page

### LINKED ISSUE #366: Resetting Password From Admin Does Not Send Email
Before I tag this to a milestone can you peek at it James?

I deployed changes and couldn't login as a user to verify an email so I tried to hit reset password for one of my users.  I got no error, I see it in the web job log as processing the message but nothing shows up in my inbox or in the sendgrid activity page?

UPDATE: I verified that verify email messages get out but not reset password.  Maybe something silently errors there?

UPDATED UPDATE: To be clear I don't mean resetting as a user, I mean login as site admin, click reset password next to a user.  It says it sends email and I see it in the webjob processor log just nothing in sendgrid

MisterJames:
- code was commented out (not sure why)
- enabled the reset/send workflow for admins from the site page

Fixes #366


dpaquette:
looks good :shipit:

PR #384, opened 2015-11-18
Allow admins to send password resets to unconfirmed email addresses

### LINKED ISSUE #365: Enable resetting of passwords for not verified users
So if you create an account and never click the link (or for some reason lose that email) and also forget your password the admin cannot reset your password

nopwdreset

Can we unbreak this chain either by allowing for deleting of users (nuclear option) #353 or by allowing (maybe with a second click to confirm?) to send the reset to an unverified email?

dpaquette:
Also displaying a warning icon when user's email address is not verified so the admins know which email addresses are not verified.

Closes #365


stimms:
:shipit:

PR #391, opened 2015-11-20
WIP: prep for final step of typescript conversion

SteveStrong:
gulpify the concat of external bower css and js files into a single directory inside wwwroot.

http://screencast.com/t/jsMt4Rn4B


MisterJames:
Deleting this PR (long tail on merge/resolution) but we can tackle it again...this scope is the size of a code-a-thon.

PR #398, opened 2015-11-25
Fixes #370 - Password requirements

### LINKED ISSUE #370: Register: list password requirements
On the "Register" page, list requirements for the password (e.g. must include a digit), so that users don't have to guess.

I would label this as an "enhancement" and maybe "jump in" -- it's not strictly necessary, but it would avoid some minimal frustration and should be easy to add.

See also #209.

stevejgordon:
fixes #370 

I've added a base PasswordRequirements class to contain the password options as constants which can then be referenced where needed through the application.

I've linked the RegisterViewModel in the Register action as this was missing previously and added a string to the model which uses the PasswordRequirements generated user friendly string to give password guidance. Hopefully this then allow easier changes to these settings in the future without having to remember to make multiple changes to annotations etc if this is used.

I've also applied this on the reset password model and view as well.

It's my first go at contributing a pull request and hoping this is correct + the code makes sense!

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
@@ -72,7 +72,8 @@ namespace AllReady.Controllers
         [AllowAnonymous]
         public IActionResult Register()
         {
-            return View();
+            var model = new RegisterViewModel();
+            return View(model);
         }
 
         //
--- a/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
@@ -177,7 +177,8 @@ namespace AllReady.Controllers
         [AllowAnonymous]
         public IActionResult ResetPassword(string code = null)
         {
-            return code == null ? View("Error") : View();
+            var model = new ResetPasswordViewModel();
+            return code == null ? View("Error") : View(model);
         }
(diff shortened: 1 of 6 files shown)


MisterJames:
Hi Steve,

Thanks and congrats on your first PR! 

I don't think this aligns exactly with where we want to go on this (I think we were looking for text on the page, and if moving towards supporting changes more easily, we would have preferred a config approach). 

That said, your contributions are valued, and we need more like them for this project to exist. I have a couple of suggestions on how to move this branch forward and would like to offer some time to work with you to make it happen. Would you be interested in doing some pair-programming on this?

MisterJames:
Also...we won't close this PR, we can keep adding commits to it to get it where we like, if you'd like to continue (not sure if you knew we could do that or not). Hope to chat soon!

stevejgordon:
Hi James. No problem. Understand I went off at a bit of a tangent here. I'd like to continue and your offer of some time to work through it is much appreciated. I don't want to hold things up though if this is needed to move things forward. If we can find time to get this done right I'm certainly up for that.

MisterJames:
@stevejgordon Let's set up for next Saturday...what timezone are you in? Can you reach out to me via twitter or email?

stevejgordon:
@MisterJames I've just sent an updated email about this Saturday.

stevejgordon:
I'm going to close this and submit a fresh PR with a clean branch.

PR #408, opened 2015-12-02
Adds a task datetime validation warning

### LINKED ISSUE #276: Tasks should inherit date range boundries of the Activity
The start and end date for the tasks should be within range of the activity it is associated with.  The date and time of the activity can be edited by the admin within the range.

 - [ ] Default the dates on tasks to the dates of the activity
 - [ ] Display a warning to the user (an exclamation indicator, for instance) if the dates are out-of-range of the activity, but allow saving

bluechrometouch:
Fixes #276


BillWagner:
@bluechrometouch Is there a reason you closed this before it was reviewed? If you're working on a different fix, that's fine, but please let me know if you want to reopen it, and have us review and merge.

PR #409, opened 2015-12-02
Adds a task datetime validation warning

### LINKED ISSUE #276: Tasks should inherit date range boundries of the Activity

bluechrometouch:
Fixes #276


AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/TaskController.cs
```diff
@@ -40,7 +40,10 @@ public IActionResult Create(int activityId)
                 ActivityName = activity.Name,
                 CampaignId = activity.CampaignId,
                 CampaignName = activity.Campaign.Name,
-                TenantId = activity.Campaign.ManagingTenantId
+                TenantId = activity.Campaign.ManagingTenantId,
+                TimeZoneId = activity.Campaign.TimeZoneId,
+                StartDateTime = activity.StartDateTime.LocalDateTime,
```
dpaquette: Don't convert to LocalDateTime here. You should set the StartDateTime to activity.StartDateTime only. The time is stored and displayed in the timezone corresponding to that Campaign.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/TaskController.cs
```diff
@@ -55,6 +58,15 @@ public IActionResult Create(int activityId, TaskEditModel model)
                 ModelState.AddModelError("EndDateTime", "Ending time cannot be earlier than the starting time");
             }
 
+            var activity = _dataAccess.GetActivity(activityId);
+            if ((model.StartDateTime < activity.StartDateTime || model.EndDateTime > activity.EndDateTime) && model.IgnoreWarning == false)
+            {
+                ModelState.AddModelError("", "Although valid, task date and time is out the range of activity date and time of " + 
+                    activity.StartDateTime.LocalDateTime.ToString() + " to " + activity.EndDateTime.LocalDateTime.ToString());
```
dpaquette: When hosting in Azure, the server time zone is UTC so this will actually display in UTC time instead of the user's local time. What we actually want is to display the time in the timezone of the campaign. In this case activity.StartDateTime.ToString() would be sufficient.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/TaskController.cs
```diff
@@ -55,6 +58,15 @@ public IActionResult Create(int activityId, TaskEditModel model)
                 ModelState.AddModelError("EndDateTime", "Ending time cannot be earlier than the starting time");
             }
 
+            var activity = _dataAccess.GetActivity(activityId);
+            if ((model.StartDateTime < activity.StartDateTime || model.EndDateTime > activity.EndDateTime) && model.IgnoreWarning == false)
```
dpaquette: The model.StartDateTime will come in as the servers timezone. When comparing dates you will need to create a new DateTimeOffset in the Campaign's timezone.  See [EditTaskCommandHandler.cs](https://github.com/HTBox/allReady/blob/master/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Tasks/EditTaskCommandHandler.cs) for an example of how this is done.

dpaquette: For tasks, model.StartDateTime and model.EndDateTime can null. In that case I don't think we want this validation warning to show.

bluechrometouch:
This needs to be fixed. I will close if there are no more comment or when I have a fix.

dpaquette:
You can leave the PR open. When you push new commits to your branch they will show up here.

BillWagner:
@bluechrometouch See @dpaquette 's comment above. Just fix it in the same branch, and push to update the PR. Then, just add a comment here when it's ready to review again.

Thanks again for all the contributions.

bluechrometouch:
Ok, this has been updated.

dpaquette:
DateTimeOffset stuff looks good now.

BillWagner:
@bluechrometouch The code looks great. Can you fix the merge conflicts, and then I'll merge it into master.

Thanks!

bluechrometouch:
Done

Issue #406, opened 2015-12-02
EF Practices - ToArray Use Prior to Where Clause


stevejgordon:
On the linq EF queries I see ToArray being called before the where clauses,,,

Example from **AllReadyDataAccessEF7.Activity.cs**
```
        IEnumerable<ActivitySignup> IAllReadyDataAccess.GetActivitySignups(string userId)
        {
            return _dbContext.ActivitySignup
                        .Include(x => x.User)
                        .Include(x => x.Activity)
                        .Include(x => x.Activity.Campaign)
                        .ToArray()
                        .Where(x => x.User.Id == userId)
                        .OrderBy(x => x.Activity.StartDateTime);
        }
```
My understanding was that one should call ToList / ToArray after the where and orderby clauses to reduce the data being returned from SQL by the query. The above seems to suggest to me we only need the data for a single user but the above would request all of the users with their activity data from the database (as the linq is executed when .ToArray is called) and only then does the data get filtered to return results for a single userId. My suggestion would be to re-order the linq to try and avoid large datasets being returned.

Unless I'm wrong or there are design decisions behind the above, should this be something that is addressed? If so I'd be happy to contribute on that.

dpaquette:
The ToArray here definitely looks like it is in the wrong place. 

On a related note, I think we are trying to replace queries like this in the AllReadyDataAccessEF7 class with Query objects like we did in the Admin area.

stevejgordon:
I'm happy to go through and make some basic fixes to the linq queries as a stage one if that's the right approach here? I'd also be interested in giving time to working on the any changes to the structure of the data access as per the admin area if that's deemed a change the project is looking to make? I'm new to contributing to GitHub projects so will take guidance on what I should do and whether further issues should be added that describe the stories to apply.

BillWagner:
@stevejgordon Let's start with looking at the LINQ queries. I do believe we likely have other similar issues here

I think it would be advisable to submit a PR for each fix, so we don't get too far out of sync. That way, as we start moving in the direction @dpaquette points out, we can have you involved in that work, and not doing parallel changes that won't necessarily get picked up. 

We recently published this http://htbox.github.io/ which should help get you started with the code. If there is anything there that is unclear, please let us know.

stevejgordon:
Thanks @BillWagner. I'll try that and hopefully do things correctly. As a starter for 10 I've put in a small PR #411 for the two queries I originally noticed.

[...]

PR #411, opened 2015-12-04
Changes to query structure to call the ToArray at end

stevejgordon:
Per issue #406 small change on the Activity data access to prevent executing the Get Activities queries before supplying the Where and OrderBy clauses. This makes the SQL query more efficient and should return only the dataset required.


dpaquette:
Looks good to me. :shipit:

Issue #422, opened 2015-12-05
Remove zip code search on campaigns page


tonysurma:
until we get it working via #193 it should be removed as search campaigns is already there

kylebcooke:
On this.

MisterJames:
Fixed via #431

PR #431, opened 2015-12-05
Removing search control from campaign page

kylebcooke:
Marked div tag to remove, ok to remove this or do we just want to hide for later use? Issue #422


AllReadyApp/Web-App/AllReady/Views/Campaign/Index.cshtml
```diff
@@ -10,7 +10,7 @@
     <link rel="stylesheet" href="~/css/campaignStyle.css" asp-file-version="true" />
 </environment>
 
-<div class="row search-banner">
+<div class="row search-banner"> <!-- About to remove this div tag for issue #422 -->
```
MisterJames: Looks like you're on the right path. If you can double check the div class for use (perhaps a JS file is bound to this class?) then this should be okay to remove per the request.

MisterJames:
:shipit:

PR #442, opened 2015-12-08
Renaming entity classes and properties from Tenant to Organization

### LINKED ISSUE #426: Death to Tenants! Long live Organizations! (Database Edition)
After the simpler renaming changes listed in #423, completing the renaming of "Tenant" to "Organization" includes renaming database entities as well. Though this will be an involved change from a testing coordination perspective, the pain of the change increases significantly the longer we put it off so let's rip this band-aid off as soon as we can.

This issue is intended to start the conversation about when and how to accomplish this change from a testing/deployment perspective so this is where folks like @tonysurma, @OhMcGoo, and @BillWagner come in.

stevejgordon:
Fixes #426 

Changed all DB related entities. Class names and properties modified. DbContext updated. Cascading changes to any uses of these classes or properties across the code base. All tests passing and was able to load the application, create, edit organizations and campaigns for those Organizations.

There is a migration in place which works if the whole DB is dropped first. Note ideal. The issue is a conflicting FK constraint. If dropping the Db and reloading is possible at this time that might be okay. Otherwise I can see if the migration can be manually tweaked so that possibly this can work on an existing Db.

I thought I would put this up and see what others think / recommend.


BillWagner:
@stevejgordon Can you update this PR to fix the merge conflicts?  Thank you.

mheggeseth:
Can you also change the "TenantAdmin" security claim to be "OrgAdmin"? This is largely a database change.

stevejgordon:
@mheggeseth Note sure I've done things correctly in the merging sense since there are now a load of prior commits showing above. I rebased my local branch on the htbox master, fixed up some of the conflicts and made the claim change. Then pushed those changes up.

Please let me know if I need to do something different or if this looks okay now.

mheggeseth:
I *think* what you have here will work (at least GH says it will merge without issue), but I don't really understand why it shows a bunch of commits that are already on HTBox/master. Also, the diff includes those changes, so it's a little hard to see only the changes you made. Maybe try just making a new branch off HTBox/master and cherry picking just your changes on to that. If that works, you can just close this PR and open a new one for the new branch.

stevejgordon:
@mheggeseth No problem. Probably something I did! I've created a new branch which I hope should work and will submit a new PR shortly.

PR #450, opened 2015-12-12
Remove blog from main nav

### LINKED ISSUE #420: Remove blog from main nav and place in footer
Perhaps in the footer with #419 we can move the blog as it doesn't deserve top billing

ksk100:
For review.

Replaces 'Blog' link in main nav with small icon next to the social share buttons.

Went ahead and used the universal blog icon, as didn't hear any objections.

Added large and small versions of icon to keep inline with other social share buttons.

Closes #420.


ksk100:
Here's a preview:

image

PR #451, opened 2015-12-13
Updated concrete reference to SqlConnection to use an interface

Kritner:
to promote unit testing.  Added new BaseDatabaseConnection to startup for DI.  Should allow for easier addition of unit testing for ClosestLocationService in #59

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ActivityAdminController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ActivityAdminController.cs
@@ -88,7 +88,7 @@ namespace AllReady.Areas.Admin.Controllers
         {
             if (activity.EndDateTime < activity.StartDateTime)
             {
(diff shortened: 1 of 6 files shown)


MisterJames:
@Kritner Thanks for the PR!

I think this work is okay if you're going to be working on testing of ClosestLocationService, but I don't think we should be spreading this approach too far into the code base. Given that we're moving to the command/query approach (it's a refactoring that is happening organically), it would likely be best to spend time moving the ClosestLocationService in that direction, instead, which would give you the same freedom for testing purposes. 

Likewise, with a shift to cmd/qry, not only do we get the easy mock of the query object in your scenario, but we also make it easy to write an integration test against the query proper without spinning up other components.

Is there something I can do to help move this forward? Would you like to give a query a shot here?

Kritner:
Sure I'll give it a shot but if someone else does it that's fine too... Don't know how much time I'll actually have to play around, but I'll give it a go I'd I have the chance

-----Original Message-----

PR #454, opened 2015-12-14
Fix ko bindings escape issue on Activity page

### LINKED ISSUE #449: KO bindings fail on activity page when tooltip string includes an encoded apostrophe
When the tooltip for the little round ? helpers contains an encoded apostrophe, `knockout-bootstrap` fails to apply the bindings.  The apostrophe is encoded properly, so I think there is a bug with `knockout-bootstrap` (I'll submit a Github issue for it).

The problem is seen on the Activity page with the sample MD skill, the description of which is "Trust me, I'm a doctor".  

A workaround is simply to swap the quotes styles.

ksk100:
@mheggeseth, do you want to check this looks okay to you?  

Closes #449

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
@@ -223,30 +223,18 @@
                                         <span asp-validation-for="SignupModel.AdditionalInfo" class="text-danger"></span>
                                     </div>
                                 </div>
-                                @if (Model.RequiredSkills?.Count > 0)
-                                {
-                                    <div class="form-group">
+                                    <div class="form-group" data-bind="visible: skills.filter(function(skill){ return !skill.IsUserSkill; }).length > 0">
                                         <label class="control-label col-md-4">Associate new skills?</label>
                                         <div class="col-md-8">
-                                            <ul class="list-unstyled form-control-static">
-                                                @foreach (var skill in Model.RequiredSkills)
-                                                {
-                                                    if (!Model.UserSkills.Any(us => us.Id == skill.Id))
-                                                    {
-                                                        <li>
-                                                            <input type="checkbox" id="@($"skillId{skill.Id}")" name="SignupModel.AddSkillIds" value="@skill.Id" />
-                                                            <label for="@($"skillId{skill.Id}")">@skill.HierarchicalName</label>
-                                                            @if (!string.IsNullOrWhiteSpace(skill.Description))
-                                                            {
-                                                                <span class="fa fa-question-circle" data-bind="visible: '@skill.Description', tooltip: { title: '@skill.Description', placement: 'top' }" aria-hidden="true"></span>
-                                                            }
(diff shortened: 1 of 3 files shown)


AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
```diff
@@ -223,30 +223,18 @@
                                         <span asp-validation-for="SignupModel.AdditionalInfo" class="text-danger"></span>
                                     </div>
                                 </div>
-                                @if (Model.RequiredSkills?.Count > 0)
-                                {
-                                    <div class="form-group">
+                                    <div class="form-group" data-bind="visible: skills.filter(function(skill){ return !skill.IsUserSkill; }).length > 0">
```
mheggeseth: For cleaner code and easier debugging, make `skills.filter(function(skill){ return !skill.IsUserSkill; })` part of the view model (it does not need to be an observable):

In the `ActivityViewModel` constructor of `activity.js`:
```JavaScript
this.skills = skills.map(RequiredSkill.bind(null, userSkills));
this.unassociatedSkills = this.skills.filter(function(skill){ return !skill.IsUserSkill; }); //added
```

mheggeseth:
One small coding style comment. Otherwise, LGTM.

ksk100:
Nice, thanks.  All done and recommitted.

AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
```diff
@@ -223,30 +223,18 @@
                                         <span asp-validation-for="SignupModel.AdditionalInfo" class="text-danger"></span>
                                     </div>
                                 </div>
-                                @if (Model.RequiredSkills?.Count > 0)
-                                {
-                                    <div class="form-group">
+                                    <div class="form-group" data-bind="visible: unassociatedSkills.length > 0">
```
mheggeseth: This is kind of silly, but `unassociatedSkills.length > 0` can just be `unassociatedSkills.length` because 0 is "falsy".

ksk100: Okay.  It was deliberate, as I think it aids readability and avoids possible problems with edge cases.  For example, I actually had to think about whether the similar use of the `Description` string field on one of the other `visible` data bindings was intentional or not.  And with the latter there is a potential problem when it's `null`, because both `null == true` and `null == false` are `false` so if KO were not testing with `!value` the code wouldn't work; I realise that the `length` of an array won't ever be null, but I thought it was a good habit.  

Anyway, I've removed the `> 0` and repushed.  I'll try to remember that terseness is highly prized :)

mheggeseth: That's a good point. Fortunately, explicitly comparing a value to `true` or `false` doesn't seem to be a common pattern.

mheggeseth:
:shipit:

PR #461, opened 2015-12-28
Changing any uses of "tentant" to "organization"

### LINKED ISSUE #423: Death to Tenants! Long live Organizations!
The initial rev of allReady referred to humanitarian organizations using the site as "tenants" which is a wonky, non-obvious term. It makes more sense to refer to them as "organizations" or "orgs" for short where appropriate.
* At the very least, the word "tenant" should not appear anywhere in the UI. Currently it shows up on `/Admin/Login`.
* As a bonus, change "tenant" code identifiers to "organization". This is easy for local variables, method names, folders, URLs, and other code identifiers that don't map to a database table or column. Changing names of database entities ~~should probably be a~~ is separate issue (#426) and warrants more discussion.

cc @MisterJames

stevejgordon:
Fixes #423


dpaquette:
Looks good to me :shipit:

PR #475, opened 2016-01-05
Fixing #471 with update to the EditCampaignCommandHandler

### LINKED ISSUE #471: Campaign FullDescription text is not persisted after create or update
Instructions to reproduce:
* Navigate to Admin..Campaigns
* Click Create New
* Fill out campaign fields, including placing some text in the FullDescription tinyMCE rich text editor
* Click Create to submit the new campaign, this returns you to the Campaign Details page
* Click "Edit" link (small pencil icon next to campaign name)
* On campaign edit page, the text of the FullDescription is now empty.
* Values entered in FullDescription field do not appear in dbo.Campaigns.FullDescription column, i.e. the text is never persisted.

stevejgordon:
Fixes #471 with slight update to the EditCampaignCommandHandler which was not passing in the FullDescription to the EF model.

I've added a couple of starter tests to verify a basic new and edit via this command. One thing I haven't worked out is line 23 of EditCampaignCommandHandlerTests:

Assert.Equal(4, result); // TODO - Why is this four not three since it's the 3rd record in the in memory database and should get an autogenerated ID of 3?

Based on my seed data and the logic this should be coming back with a new Id of 3. If anyone can see why this isn't the case I'd like to know and maybe I can update the PR?

This handler does need some more tests to verify things such as saving the location but in the interest of fixing the immediate bug quickly I've not added them in this PR and propose to add more tests relating to #59 when I get a bit of spare time.


stevejgordon:
Brain now more engaged I've amended the test logic to simply check for an ID greater than 0 being returned which shows a successful add has taken place.

dpaquette:
Yeah checking for >0 makes sense @stevejgordon 

This looks good. :shipit:

PR #477, opened 2016-01-06
Adding test for anti forgery token attribute

stevejgordon:
A quick lunchtime PR! While reviewing commit 229ece0aab95e7527009d2136674793910ffaffc from @Sobieck00 I saw the inclusion of a check for the Anti Forgery Token and thought about putting it into some of the tests I've written.

I then looked around for a solution to doing this in one place and found http://codiply.com/blog/test-for-omission-of-validateantiforgerytoken-attribute-in-asp-net-mvc

I've included this code in AllReady and it seems to work well. It picked up one failure for the ActivityController where this attribute was missing so I've fixed that up also. I consider this related to #59 

If this looks sensible to others then perhaps we can remove any per controller/action checks?

Note: This may be WIP depending on feedback but if others are happy then I'd consider this done and ready to merge.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/GlobalControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/GlobalControllerTests.cs
@@ -1,0 +1,43 @@
+﻿using AllReady.Controllers;
+using Microsoft.AspNet.Mvc;
+using System;
+using System.Linq;
+using Xunit;
+
+namespace AllReady.UnitTest.Controllers
+{
+    public class GlobalControllerTests
+    {
+        /// <summary>
+        /// A test to reflect across all controllers and valide all post requests also have the ValidateAntiForgeryToken attribute
+        /// </summary>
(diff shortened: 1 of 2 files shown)


dpaquette:
@stevejgordon have you tested to make sure Activity Signup still works after you added the attribute?  I think that controller action is called via Ajax and it might not pass a valid anti forgery test with the request.

stevejgordon:
@dpaquette Yes, I did test it and the modal seemed to submit without issue on my end. The form on the modal does include the @Html.AntiForgeryToken() as well.

dpaquette:
Ah I see where that's being called now. I was worried about this line
https://github.com/HTBox/allReady/blob/master/AllReadyApp/Web-App/AllReady/wwwroot/js/activity.js#L26

but that's the Activity API controller which is different.

This PR looks good to me. It's good to have automated checks like this in place.

:shipit:

Sobieck00:
I think this is a great idea. Do you think that the ValidateAntiForgeryToken tests I wrote have value after this is in place? Should they be deleted?

stevejgordon:
@Sobieck00 I think this does mean we could remove the test that you added. I see no real harm in having both but it might mean others copy the pattern unnecessarily.

Issue #296, opened 2015-11-09
User profile - Make email editable or remove/add like phone number


Harmonym:
The user should be able to access the profile form after initial set-up to make edits.  Perhaps as part of a drop-down from the username on the top right. 

![image](https://cloud.githubusercontent.com/assets/8798159/11026158/5ba63a74-865a-11e5-83f2-a7c6bd88a167.png)

ksk100:
Are there any aspects of the user's profile which need to be editable, but which aren't already editable via the standard manage screen (the one you get if you click on the user's email)?

tonysurma:
Now that that is all working I think everything is editable that needs to be (some day we have to deal with email changes) but there may be some value in adding a UX addition like a drop down arrow or some indication that your name is clickable / takes you to the page to edit them.

Maybe a small menu might be good and this can be 'Profile' as soon we will have other items for a user as well (like seeing all the tasks they volunteered for in one place)

tonysurma:
Definitely need a way to edit email if I typo it when creating my account then I can never change it.  We can do it with phone number currently.

BillWagner:
@tonysurma How would that scenario (mis-typed email) work when the email addr isn't confirmed? Delete user and re-create? Admin only?

MisterJames:
So, the email address is currently being used as the username. We'll need to have a think through on this to make sure that the impact here is understood.

 1. If the user changes their email, we would need to reconfirm it
 1. We probably shouldn't change the primary email until they've confirmed the new one (to prevent accidental lock outs or mal-intent)
 1. We should likely consider a preferred email address, or a campaign-specific email address

Thoughts?

stevejgordon:
I'm not 100% sure how best to tackle this yet but could have a try in a branch if no one is currently progressing this one?

@MisterJames 1 and 2 makes sense to me and perhaps we use a couple of additional DB fields to track this. For item 3 I'm perhaps not clear on how you see that working though? Are you proposing a user could store multiple emails (similar to PayPal) with one marked as primary?

mheggeseth:
I'm not sure I understand @MisterJames 3rd item either, but we already have a way to specify an alternate email when volunteering. That alternate email so neither confirmed or associated with the user in a permanent way though.

stevejgordon:
Maybe the simplest starting point is just a request to change pw with confirmation before final committal. I'll have a play and get something going which we can adjust as necessary.

stevejgordon:
[...]

PR #484, opened 2016-01-11
Edit Primary Email Address

stevejgordon:
WIP for #296 

I think this is mostly there but wanted to get it up for a quick code review. I've created actions of the ManagerController to allow the option to set up a new email address, confirm it via email and if required resend the confirmation and cancel the request.

I've updated the UI also to display this on the manage page. That might need a little tidy up to improve the layout at some stage but wanted to focus on the functionality for now.

I've yet to write any tests but plan to before we finally merge this, once we're happy with it.

I've also yet to test actual emails are received since I wasn't immediately sure how to setup the email for local dev. Is there any advise available for that. I did add SendGrid credentials but it doesn't seem to use that and instead is an AzureQueue?

As I take it email is working elsewhere I didn't spend much time on this as the email is being pushed to the IEmailSender.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
@@ -334,6 +334,8 @@ namespace AllReady.Controllers
             return View(model);
         }
 
+
+
         #region Helpers
 
         private void AddErrors(IdentityResult result)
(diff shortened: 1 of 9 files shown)


tonysurma:
@dpaquette or @MisterJames can you take a look at this and/or any thoughts on the email sending Steve mentioned?

dpaquette:
@tonysurma I think the implementation looks good. 

@stevejgordon When running locally, the emails are logged to your debug output window in Visual Studio. You can see the contents of the email there and test the links. When deployed it sends to a queue which when processed sends emails via SendGrid.  You don't need to test the actual email sending since that is really a separate piece.  As along as the messages appear in the console you should be ok.

stevejgordon:
Fixes #296

Issue #261, opened 2015-11-08
New User Registration Workflow - Name, e-mail, sms


Harmonym:
Prompt, but don't require, new users to supply additional profile information either after they have set up initial log-in through FB, Twitter, MSFT account.

Request:
Name
E-mail
Phone number
Phone number type (drop-down)

Note: mobile numbers will be used for SMS notifications.

stevejgordon:
Please assign me to this issue and I'll start having a look during this week.

@tonysurma My only question at this point is around the phone number type dropdown. If I look at the app currently where we add a phone number it seem to assume all numbers are mobile since it intends to send a verification SMS? Should there also be a concept of a landline?

tonysurma:
@stevejgordon good question, let me noodle that.  Initial thought is guide to mobile preferred and maybe not a drop down type but a [ ] for mobile can receive txt messages or something like that...

stevejgordon:
@tonysurma Just wondering if the PR resolved this or if we need more work (perhaps new issue/story) to allow phone types?

BillWagner:
Closed per PR #485.

PR #485, opened 2016-01-11
Social login data collection

stevejgordon:
WIP for issue #261 

I've opened this to see if this is the correct approach for the issue at hand. There were a couple of possible ways to achieve this and I've taken what I think is the easiest for now. I've added the Name and Phone Number fields to the external callback page so that these can be optionally entered.

Since the flow from the manage page is the require the SMS code to be provided before the number is added I've tweaked the UI a little to put up a UI prompt when the phone number is not yet confirmed in /Manage

This may not be the preferred approach since the phone number is in the DB at this stage before confirmation, although the confirmed flag is not set. We could adjust the flow if preferred to instead send the user directly to the confirm code page as if they had added a new phone number but for now I kept the path of least resistance to redirect to the redirectUrl.

I've also added a verify link on the Manage page which takes the user to the screen to supply the code. I've also added a resend code button as that might be useful.

Opinions welcomed :-)

This will conflict with changes in PR #484 as well so whichever is last to be completed I'll need to rebase and fixup the conflict since I've edited a few of the same files in each PR.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
@@ -174,7 +174,7 @@ namespace AllReady.Controllers
             // Generate the token and send it
             var user = GetCurrentUser();
             var code = await _userManager.GenerateChangePhoneNumberTokenAsync(user, model.PhoneNumber);
-            await _smsSender.SendSmsAsync(model.PhoneNumber, "Your security code is: " + code);
+            await _smsSender.SendSmsAsync(model.PhoneNumber, "Your allReady account security code is: " + code);
             return RedirectToAction(nameof(VerifyPhoneNumber), new { PhoneNumber = model.PhoneNumber });
         }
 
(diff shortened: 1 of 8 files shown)


tonysurma:
I will take a look at the 'approach' as you said but can someone else do a code review, thanks and thanks @stevejgordon !

AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
```diff
@@ -176,6 +177,19 @@ public IActionResult AddPhoneNumber()
...
 
+        // POST: /Account/ResendPhoneNumberConfirmation
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<IActionResult> ResendPhoneNumberConfirmation(string phoneNumber)
+        {
+            var user = GetCurrentUser();
+            var code = await _userManager.GenerateChangePhoneNumberTokenAsync(user, phoneNumber);
+
+            await _smsSender.SendSmsAsync(phoneNumber, "Your security code is: " + code);
```
dpaquette: Should the message include the name of the app?  Something like "Your allReady account security code is: "

stevejgordon:
@dpaquette That sounds very sensible to me. I'll get a commit in ASAP to update that. I'll also rebase off of the latest master as well now some of the other PR's are merged.

stevejgordon:
@dpaquette Message updated as discussed. Logic updated slightly to fix a bug and also rebased off of the current master.

AllReadyApp/Web-App/AllReady/config.json
```diff
@@ -34,8 +34,8 @@
   },
   "Authentication": {
     "Facebook": {
-      "AppId": "[facebookappId]",
-      "AppSecret": "[facebookappsecret]"
+      "AppId": "167952183568521",
+      "AppSecret": "a8053ceca7eefbaa4821ca531929f339"
```
robrich: Reversed by 1d0636c

PR #486, opened 2016-01-12
Hiding volunteer link for a closed activity

### LINKED ISSUE #478: Volunteer available for closed activities
When testing the activity process I noticed that when viewing the details for a closed activity within a campaign I could still volunteer.

Is this correct as I'd expect that once closed no new volunteers can signup? 

If this is confirmed as a bug I propose we can simply hide the volunteer link for any closed activities.

stevejgordon:
Fixes #478


tonysurma:
anyone able to do a code review of this?  Thanks.

mheggeseth:
Can you print that the activity is closed other than showing nothing. Also, I think we should still show the "thanks for being a volunteer" message for logged in users who were volunteers.

stevejgordon:
@mheggeseth Sure. I've added a new commit per your feedback. Thanks.

mheggeseth:
The unenroll link should probably be hidden for closed activities.

stevejgordon:
@mheggeseth Updated to hide un-enroll

mheggeseth:
Looks good. Don't know why the scenario tests are failing, but it doesn't seem to have to do with this code. Maybe push a non-functional change to force a check to see if it passes.

stevejgordon:
@mheggeseth Didn't seem to work

PR #488, opened 2016-01-13
Refactoring so that Child property is not used on the Skill model

### LINKED ISSUE #487: Activity View Crashes with Skills Having Parent Skills
The activity view (`Views/Activity/Activity.cshtml`) crashes if any of the activity's required skills have a parent skill or if any of the user's associated skills have a parent skill.

@stevejgordon This appears to be the result of adding the `Children` property to the `Skill` class. It causes a circular reference exception when trying to JSON serialize. Is the property strictly needed or can you get `SkillDeleteQueryHandlerAsync` to work without it?

stevejgordon:
Fixes #487


stevejgordon:
See comment in #487 before merging as there's two ways we could fix this.

mheggeseth:
Looks good :shipit:

PR #493, opened 2016-01-21
When a Registered User Unvolunteers, notify organizer of any tasks that have been unassigned

### LINKED ISSUE #473: When a Registered User Unvolunteers, notify organizer of any tasks that have been unassigned
Split from #183 to create a more manage-able pull request, please see discussion there.

mmoore99:
Fixes #473.

An email will now be sent to the campaign's primary contact, as defined in **Campaign.CampaignContacts**, when a user unenrolls from an activity.  The attached file contains an example of the email that is generated.

[email.txt](https://github.com/HTBox/allReady/files/99578/email.txt)

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Notifications/GetUserUnenrollsNotificationModel.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Notifications/GetUserUnenrollsNotificationModel.cs
@@ -1,0 +1,139 @@
+﻿using System;
+using System.Collections.Generic;
+using AllReady.Features.Notifications;
+using AllReady.Models;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace AllReady.UnitTest.Notifications
+{
+    public class GetUserUnenrollsNotificationModel : InMemoryContextTest
+    {
+        private ApplicationUser _user1;
+        private Organization _htb;
+        private Campaign _firePrev;
+        private Activity _queenAnne;
+        private Contact _contact1;
+        private AllReadyTask _task1;
+
+        protected override void LoadTestData()
+        {
+            var context = ServiceProvider.GetService<AllReadyContext>();
+            _user1 = new ApplicationUser
+            {
+                UserName = "johndoe@example.com",
+                Name = "John Doe",
+                Email = "johndoe@example.com"
+            };
+            _contact1 = new Contact
+            {
+                Id = 1,
+                FirstName = "Jerry",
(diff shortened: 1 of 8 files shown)


AllReadyApp/Web-App/AllReady/Features/Notifications/NotifyAdminForUserUnenrolls.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (string.IsNullOrWhiteSpace(campaignContact?.Contact.Email)) return;
+
+            var signup = model.UsersSignedUp.FirstOrDefault(s => s.User.Id == notification.UserId);
+
+            if (signup == null) return;
+
+            var activityLink = $"View activity: {_options.Value.SiteBaseUrl}Admin/Activity/Details/{model.ActivityId}";
+            var subject = $"A volunteer has unenrolled from {model.ActivityName}";
+
+            var message = new StringBuilder();
```
MisterJames: This will work for now. @BillWagner and @tonysurma we should talk about getting the templates idea on sendgrid going on the standup this weekend.

AllReadyApp/Web-App/AllReady.UnitTest/Notifications/GetUserUnenrollsNotificationModel.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        [Fact]
+        public void ModelCanBeCreatedFomExistingActivity()
+        {
+            var context = ServiceProvider.GetService<AllReadyContext>();
+            var query = new UserUnenrollsNotificationQuery { ActivityId = 1 };
+            var handler = new UserUnenrollsNotificationQueryHandler(context);
+            var result = handler.Handle(query);
+            Assert.NotNull(result);
+            Assert.True(_queenAnne.Id == result.ActivityId, "ActivityId does not match");
```
MisterJames: I'm not sure we need _all_ these asserts, as at this point the properties aren't really transformed, they are just returned from the in-mem store. How about the null check, and that the collections contain the correct count? Otherwise, we're just really testing language and library features (like object initialization and the in-mem store).

MisterJames:
Good PR @mmoore99. Would you mind simplifying the checks on that test a little and we should be good to merge.  Thanks!

mmoore99:
@MisterJames I simplified the checks on the test as requested. Let me know if there is anything else needed.

MisterJames:
:shipit: Thanks again!

PR #499, opened 2016-01-26
Campaign lock and unlock

### LINKED ISSUE #39: 5.1.1.4.1	Allow AllReady admins to unpublish campaigns published by organization admins
[later functionality or in V1 if possible would be to ban org admin accounts]

stevejgordon:
Fixes #39 

I've taken the approach to add a bool to the campaign model to mark it as locked.
SiteAdmins can lock/unlock any campaign from the details page using the button I've added.
OrgAdmins will see a message stating the campaign has been locked.
In the admin campaign lister I have added a warning icon to locked campaigns
In the public lister, locked campaigns are not shown
API updated to also not display locked campaigns

Hopefully that makes sense and meets the requirement?


mmoore99:
@stevejgordon since the home page has been modified to display a list of current/upcoming campaigns, the Home/Index action will need to be modified to ignore locked campaigns.

'''C#
return View(_dataAccess.Campaigns.Where(c => c.EndDateTime.UtcDateTime.Date > DateTime.UtcNow.Date).ToViewModel().OrderBy(vm => vm.EndDate).ToList());
```

**Comment by stevejgordon** [comment]
@mmoore99 Thanks Mike. Good catch.

**Comment by tonysurma** [comment]
@MisterJames can someone look over this PR before I merge?

File: AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Details.cshtml [hunk 1]
```diff
@@ -11,7 +13,31 @@
...
-        <div class="col-md-9"><h2>@Model.Name <a asp-controller="Campaign" asp-action="Edit" asp-route-id="@Model.Id" class="btn btn-default" title="Edit"><i class="fa fa-edit"></i></a> <a asp-controller="Campaign" asp-action="Delete" asp-route-area="Admin" asp-route-id="@Model.Id" class="btn btn-default"><i class="glyphicon glyphicon-trash"></i></a></h2></div>
+        <div class="col-md-9">
+            <h2 class="inline-form">
+                @Model.Name
+            </h2>
+            <a asp-controller="Campaign" asp-action="Edit" asp-route-id="@Model.Id" class="btn btn-default inline-form" title="Edit"><i class="fa fa-edit"></i></a>
+            <a asp-controller="Campaign" asp-action="Delete" asp-route-area="Admin" asp-route-id="@Model.Id" class="btn btn-default inline-form"><i class="glyphicon glyphicon-trash"></i></a>
+            @if (User.IsUserType(UserType.SiteAdmin))
+            {
+                <form asp-controller="Campaign" asp-route-area="Admin" asp-action="LockUnlock" class="inline-form">
```
mheggeseth: Why does this have to be a form? Seems like a lot of extra code. You are already validating that the logged-in user is a SiteAdmin.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
```diff
@@ -147,9 +147,24 @@ public IActionResult DeleteConfirmed(int id)
             {
                 return HttpUnauthorized();
             }
+
             _bus.Send(new DeleteCampaignCommand { CampaignId = id });            
             return RedirectToAction("Index", new { area = "Admin" });
         }
 
+        // POST: Campaign/Delete/5
```
mheggeseth: Comment has wrong URL

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Campaigns/LockUnlockCampaignCommandHandlerTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            var campaign = Context.Campaigns.FirstOrDefault(c => c.Name == "Locked Campaign");
+            handler.Handle(new LockUnlockCampaignCommand { CampaignId = campaign.Id });
+            var result = Context.Campaigns.FirstOrDefault(c => c.Name == "Locked Campaign");
+
+            // Assert
+            Assert.False(result.Locked); // Campaign should now be unlocked
+        }
+
+        [Fact]
+        public void UnlockedCampaignIsUnlocked()
```
mheggeseth: Rename to `UnlockedCampaignIsLocked`. That's what you are testing.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Campaigns/LockUnlockCampaignCommandHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        }
+        protected override void HandleCore(LockUnlockCampaignCommand message)
+        {
+            var campaign = 
+                _context.Campaigns.SingleOrDefault(c => c.Id == message.CampaignId);
+
+            if (campaign != null)
+            {
+                if (campaign.Locked)
```
mheggeseth: Change to `campaign.Locked = !campaign.Locked;`

AllReadyApp/Web-App/AllReady/Controllers/CampaignController.cs
```diff
@@ -53,19 +52,22 @@ public IActionResult LocationMap(int id)
...
         {
             return _dataAccess.Campaigns
+                .Where(c => c.Locked == false)
                 .Select(x => new CampaignViewModel(x));
         }
 
         // GET api/values/5
         [HttpGet("{id}")]
-        public CampaignViewModel Get(int id)
+        public ActionResult Get(int id)
```
mheggeseth: Where is this used? I can't figure it out.

AllReadyApp/Web-App/AllReady/Controllers/HomeController.cs
```diff
@@ -17,7 +17,7 @@ public HomeController(IAllReadyDataAccess dataAccess)
 
         public IActionResult Index()
         {
-            return View(_dataAccess.Campaigns.Where(c => c.EndDateTime.UtcDateTime.Date > DateTime.UtcNow.Date).ToViewModel().OrderBy(vm => vm.EndDate).ToList());
+            return View(_dataAccess.Campaigns.Where(c => c.EndDateTime.UtcDateTime.Date > DateTime.UtcNow.Date && c.Locked == false).ToViewModel().OrderBy(vm => vm.EndDate).ToList());
```
mheggeseth: Use `!c.Locked` instead of `c.Locked == false`

mheggeseth:
Couple other issues:
* Activities in locked campaigns are still publicly accessible.
* The UI should do a bit better job explaining what "locked" means.

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Details.cshtml
stevejgordon: On this one it's not presenting any other page or input via a http get so I felt it best practice to make this require a http post to the action via a form. If we make it just a hyperlink we potentially should add a page to confirm that we want to lock or unlock the campaign which felt unnecessary.

stevejgordon:
@mheggeseth Thanks for your feedback. I'll get the changes made ASAP.

AllReadyApp/Web-App/AllReady/Controllers/CampaignController.cs
stevejgordon: This is something I noticed when testing the API. Previously there was a HttpNotFound(); line which seems to do nothing since it wasn't being actually returned. However to return it the controller needs to be an action result rather than just the model. I may have misunderstood something here about how the MVC Core controllers work for API responses though?

If this is truly an issue we may need to check the other API endpoints as well.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Campaigns/LockUnlockCampaignCommandHandler.cs
stevejgordon: @mheggeseth Sorry Mike. I'm a little confused by this suggestion. Can you expand?

mheggeseth: @stevejgordon Sorry, I tend to leave comments on the first line of a block instead of the last forgetting how GitHub likes to show them. What I mean is that
``` CSharp
if (campaign.Locked)
{
  campaign.Locked = false;
}
else
{
  campaign.Locked = true;
}
```
is functionally equivalent to and more succinctly written as
```
campaign.Locked = !campaign.Locked;
```

stevejgordon: Thanks @mheggeseth

stevejgordon:
@mheggeseth I worked on this before work this morning and have most of it ready. I want to do a little more before updating the PR. Did you have any thoughts on what we should display in the UI to explain the locked status?

mheggeseth:
I was thinking a simple tooltip something like "This campaign has been locked by a site administrator. Locked campaigns are not visible to the public or to volunteers."

Also, just reading this phrasing makes me think a site admin should enter a comment/reason for locking the campaign. Otherwise, the org admin is in the dark and doesn't know if/what they should change about the campaign. This could be a separate PR.

stevejgordon:
@mheggeseth I've pushed an update which address the above feedback. I've left the form for now unless you think we should just use a hyperlink and post action? I plan to open a new issue later for your thoughts on a lock comment etc.

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.Activity.cs
```diff
@@ -20,6 +20,7 @@ IEnumerable<Activity> IAllReadyDataAccess.Activities
                                 .Include(a => a.Tasks)
                                 .Include(a => a.RequiredSkills)
                                 .Include(a => a.UsersSignedUp)
+                                .Where(a => !a.Campaign.Locked)
```
mheggeseth: I realize this only has one referrer in `ActivityApiController` but it doesn't feel quite right to exclude locked activities _before_ returning from the repository class. I'd rather see those locked activities filtered out in `ActivityApiController`.

mheggeseth:
I had one comment on some code that you added to `IAllReadyDataAccess.Activities`. Everything else looks good.

stevejgordon:
@mheggeseth Thanks again for the feedback. Good point and I've updated the PR.

mheggeseth:
Looks good. :shipit:

PR #509, opened 2016-02-02
Campaign Social Sharing

### LINKED ISSUE #27: 5.1.1.1.5	Allow citizens to share campaign information via email, txt, twitter and facebook

stevejgordon:
WIP to fix #27 

I've tried to find a simple way to share the campaign via existing social api links. We could integrate the standard twitter/facebook buttons but for Facebook I think that now requires an AppId. I used the existing icon buttons + added one new one for email.

NOTE: Facebook link doesn't work for localhost urls since Facebook tries to call the page to get the meta info. However on a live site it should work.

I've used the ViewBag for now to create and save the absolute URL. Not sure if there's a better way in the ASP.NET Core api. I don't really like using ViewBag In most cases, although this didn't feel like a bad fit. However I could update the view model we use instead if preferred.

For email I chose for now to use a mailto link since the sender is then able to send from their mailbox rather than the email potentially being deemed unsolicited from already. We could however use a modal form and send grid approach if preferred,

@tonysurma - The only item not addressed with this is sharing via text. Was the intent there to open a dialog to supply an SMS number and message and then send the message?


tonysurma:
For now I am good with everything but txt, can someone take a look at this to verify we are good to ship after it is rebased on recent changes?  Thanks

tonysurma:
@stevejgordon can you rebase and @dpaquette or @mheggeseth can you review?

dpaquette:
I can review this one after it has been rebased

dpaquette:
I think this one looks good. I am wondering if the logic to generate the AbsoluteUrl belongs more in the view than in the Controller but I could easily see arguments for both sides there. 

Assuming the rebase is good, I say :shipit:

tonysurma:
thanks @dpaquette
@stevejgordon can you rebase and I'll merge?  Thanks again

stevejgordon:
@tonysurma Sure, I'll get to a point where I can commit something I'm working on currently and then rebase this branch.

PR #514, opened 2016-02-07
Add "ActivityType" property to "Activity" model to support new Activity workflows

mmoore99:
Fixes #505.

## Note: This PR may need to be merged into a feature branch instead of Master

This PR  is a pre-requisite for issues #506, #507 & #508 and provides the following:
- [x] Add "ActivityTypes" enum with entries for the different activity types (ActivityManaged, TaskManaged, DeploymentManaged).  Note that per @tonysurma in #504 these names may be changed.
- [x] Add "ActivityType" property to "Activity" model
- [x] Create database migration script
- [x] Add "ActivityType" to  admin add/edit dialog for Activity 

If this were merged into "master" the only functionality change in the app visible to the user would be the "ActivityType" field appearing in the admin add/edit for Activity.  This property is not yet acted on by the rest of the app.  If this needs to be merged into a feature branch and I need to submit a different PR, please advise.

Issue #507, opened 2016-02-01
Add support for "Type T: Task Manged" activities


mmoore99:
This is sub issue of #504 

More detail to follow.

mmoore99:
For this type of activity the following is true:

- Tasks are shown to volunteers
- Volunteers volunteer for Tasks
- Admins can re-assign tasks to volunteers
- Re-assignment communication is only sent to Volunteers when it results in change outside of assigned timeframe and they can accept/reject task assignment
- Reminder sent ahead of Activity with option to respond as 'no longer able to volunteer'

mmoore99:
I am currently on the code for this issue.

mmoore99:
Submitted PR #523

mmoore99:
@tonysurma @OhMcGoo when setting up a `TaskManaged` activity:

- Should we prevent entry of the `Volunteers Required` field in the activity in order to prevent confusion since the `Volunteers Required` will be specified for each task?

- Same question for `Start Date` and `End Date`? Rather entering these dates manually we could calculate them from the dates in the tasks.

- It would also seem that there is not a need to enter `Required Skills` for the activity itself, since they will be entered for each task.  Is that true?  If so, would it be beneficial to provide the option to enter   `Required Skills` that would be automatically replicated to all tasks in the activity.  For example, if we had the following activity/task structure and had a large number of tasks:
  - Activity: Install fire detectors in Queen Ann District
    - Task: Install - 2300 Main Street
    - Task: Install - 3250 Marsh Drive
    - Task: Install - 1234 Norris Canyon Drive

- In this case all of the tasks might have a skills required of "Electrician" and "Driver".

I would appreciate your feedback on this, as I am in the process of trying to finalize this issue.

Thanks.

OhMcGoo:
@mmoore99, @tonysurma: See my answers inline.

Should we prevent entry of the Volunteers Required field in the activity in order to prevent confusion since the Volunteers Required will be specified for each task?

**I agree that Volunteers Required should be set at the Task level, however, we need to ensure that an individual Volunteer can be assigned more than one task during an Activity. For example, a Volunteer may be assigned to the task of delivering supplies to a staging area pre-Event. They, then, could asked to join an installation team later.**

Same question for Start Date and End Date? Rather entering these dates manually we could calculate them from the dates in the tasks.

**I think a *bookend* date range should be set at the Activity level. Then, Task-specific date ranges can be set within bookend dates.**

[...]

Issue #502, opened 2016-01-30
Scenario V2: Volunteer volunteers for Campaign


tonysurma:
_These issues are scenarios documented for use of the application.  Likely will be the 'parent' of multiple sub issues to verify, add or fix functionality to support them._

# **Volunteer Volunteers for Campaign**
Initiation: Volunteer has navigated to campaign at activity or task level where volunteering is an option (per campaign type)
Features:
- [ ] Volunteer link is present and worded with clarity _not just 'volunteer' but something like 'Volunteer for this Campaign'_
- [ ] If Volunteer is not logged in they are prompted to log in
- [ ] If Volunteer is not registered they are prompted to register
- [ ] Volunteer fills in fields presented (which are prepopulated from profile as possible)
- [ ] Volunteer self identifies having skills presented by checking checkboxes in list
- [ ] Skills the Volunteer have previously identified and that are in this list are already checked
- [ ] After volunteering a message is displayed on that campaign page indicating they are registered as a volunteer for this campaign and an event organizer will be in touch

mmoore99:
@MisterJames @tonysurma With the changes being introduced to the Activity workflows in #504 (Volunteer & Org Admin Workflow Differs By Activity Type), I would propose that the workflow items shown here should be merged into in the individual issue items for each of the activity types (see #506 ) and that this issue be closed to prevent confusion.

mmoore99:
**This is covered by PR's #518 and #523**

tonysurma:
I need to review what all is complete per the PRs and what is open per new requirements if anything

mmoore99:
@tonysurma the workflow described in this issue was handled in #506 & #507. This issue can be closed.

PR #518, opened 2016-02-13
Add support for "Type A: Activity Manged" activities

### LINKED ISSUE #506: Add support for "Type A: Activity Manged" activities
This is sub issue of #504 

The following represents an initial pass at attempting to flush out the requirements for this issue.  Comments, and answers to questions, would be appreciated.

### For this type of Activity the following is true as defined in #504 :
- Tasks are not shown to volunteers
- Volunteers volunteer for Activities
- Admins assign tasks to volunteers
- NO assignment communication is sent to Volunteers and thus do not get option to accept/reject
- Reminder sent ahead of Activity with option to respond as 'no longer able to volunteer'

### Desired workflow as defined in #502

The checked items are supported by the existing code.
- [ ] Volunteer link is present and worded with clarity not just 'volunteer' but something like 'Volunteer for this Campaign'
- [x] If Volunteer is not logged in they are prompted to log in
- [x] If Volunteer is not registered they are prompted to register
- [x] Volunteer fills in fields presented (which are prepopulated from profile as possible)
- [x] Volunteer self identifies having skills presented by checking checkboxes in list
- [x] Skills the Volunteer have previously identified and that are in this list are already checked
- [ ] After volunteering a message is displayed on that campaign page indicating they are registered as a volunteer for this campaign and an event organizer will be in touch
- [ ] After volunteering the Volunteer link is replaced with a "No Longer Able to Volunteer" link 

### Work To Be Done
- [ ] On activity detail page eliminate display of tasks for this type of activity
- [ ] On activity detail page make 'Volunteer'  link more prominent (ie rename to something  like 'Volunteer for this Campaign')
- [ ] On activity detail page modify confirmation message after volunteering to indicate that they are registered as a volunteer for this campaign and an event organizer will be in touch
- [ ] Create endpoint to process volunteer cancellation from "No Longer Able to Volunteer" link in reminder email

### Questions/Issues
- Should the app restrict the number of volunteers allowed for an activity to the number specified in the "NumberOfVolunteersRequired" property of the activity?  Similarly, should the app restrict the number of volunteers that an admin is allowed to assign to a task to the "NumberOfVolunteersRequired" property of the task?
- Should the list of activities show the number of volunteers needed/number of volunteers signed up?  Should there be a way to filter the list to only show activities that still need volunteers?
- No communication is to sent to volunteers when an admin assigns the volunteer to a task.  However, is it correct that a confirmation email would still be sent to the volunteer upon volunteering or unvolunteering from an activity as per PR #500 for #372?
- Shouldn't the requirement of  "Reminder sent ahead of Activity with option to respond as no longer able to volunteer" be handled in  #258 which addresses the issue of automatic reminder notifications and #389 which addreses a notification framework?
- Currently the app allows admins to assign/unassign volunteers to tasks, but there does not seem to be a way for an admin to "unvolunteer" a volunteer from the activity itself. Is this a requirement?
- Currently when an admin assigns a task to the volunteer the task signup is set to a "Assigned" status that changes to "Accepted" or "Rejected" when the volunteer responds. Since, for this type of activity, the volunteer does not get the option to accept or reject should the status of the task signup be set to "Accepted" upon assignment by the admin or remain as "Assigned"?

mmoore99:
Fixes #506

**This PR should be able to be merged into "Master" as it does not  introduce any breaking changes.  Unless an Activity is specifically assigned an "ActivityType" of "ActivityManaged" the original workflow for avtivities and tasks remains intact.**

**Note: this PR depends on PR #514**

This PR implements the following changes to the functionality of the Activity Detail Page:

-  No task information is displayed for an "Activity Managed" activity
- The "Volunteer" link was changed to a button and made more prominent (**see Screenshot 1**)
- Client side validation has been implemented for the activity signup dialog using Knockout (**see Screenshot 2**)
- A confirmation message is now displayed after a successfull signup (**see Screenshot 3**)
- A confirmation message is now displayed after unenrolling from an activity (**see Screenshot 4**)
- The posting of the signup activity dialog is now done through an Ajax call to the ActivityApi conntroller, thereby eliminating the need for a page refresh after signing up.  While, this may not be a specific requirement for this issue, it is needed to support functionality for issue #507 where the user will be picking from a list of tasks on the page.

This PR also fixes a bug in the existing code when a user unenrolls from an Activity where task(s) have been assigned to the user. The code was attempting to delete "Tasks" instead of "TaskSignups" and also was causing system exception to be thrown.

**Screenshot 1**
screenclip 1

**Screenshot 2**
screenclip 2

**Screenshot 3**
screenclip 3

**Screenshot 4**
screenclip 4

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
@@ -145,7 +145,7 @@ namespace AllReady.Controllers
         [ValidateAntiForgeryToken]
         [HttpPost("signup")]
         [Authorize]
-        public IActionResult RegisterActivity(ActivitySignupViewModel signupModel)
+        public async Task<IActionResult> RegisterActivity(ActivitySignupViewModel signupModel)
         {
             if (signupModel == null)
             {
@@ -159,7 +159,7 @@ namespace AllReady.Controllers
                 return Json(new { errors = ModelState.GetErrorMessages() });
             }
 
-            _bus.SendAsync(new ActivitySignupCommand() { ActivitySignup = signupModel });
+            await _bus.SendAsync(new ActivitySignupCommand() { ActivitySignup = signupModel });
             return new HttpStatusCodeResult((int)HttpStatusCode.OK);
         }
 
--- a/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
@@ -160,7 +160,7 @@ namespace AllReady.Controllers
             }
 
             _bus.Send(new ActivitySignupCommand() { ActivitySignup = signupModel });
-            return new {Status = "success"};
+            return new HttpStatusCodeResult((int)HttpStatusCode.OK);
         }
 
         [HttpDelete("{id}/signup")]
--- a/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
@@ -145,7 +145,7 @@ namespace AllReady.Controllers
         [ValidateAntiForgeryToken]
         [HttpPost("signup")]
         [Authorize]
-        public async Task<IActionResult> RegisterActivity(ActivitySignupViewModel signupModel)
+        public object RegisterActivity(ActivitySignupViewModel signupModel)
         {
             if (signupModel == null)
             {
@@ -160,7 +160,7 @@ namespace AllReady.Controllers
             }
 
             _bus.Send(new ActivitySignupCommand() { ActivitySignup = signupModel });
-            return new HttpStatusCodeResult((int)HttpStatusCode.OK);
+            return new {Status = "success"};
         }
 
         [HttpDelete("{id}/signup")]
--- a/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
@@ -176,7 +176,7 @@ namespace AllReady.Controllers
             //Notify admins & volunteer
             await _bus.PublishAsync(new UserUnenrolls {ActivityId = signedUp.Activity.Id, UserId = signedUp.User.Id});
 
-            await _allReadyDataAccess.DeleteActivitySignupAsync(signedUp.Id);
+            await _allReadyDataAccess.DeleteActivityAndTaskSignupsAsync(signedUp.Id);
             return new HttpStatusCodeResult((int)HttpStatusCode.OK);
         }
 
--- a/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
@@ -145,7 +145,7 @@ namespace AllReady.Controllers
         [ValidateAntiForgeryToken]
         [HttpPost("signup")]
         [Authorize]
-        public object RegisterActivity(ActivitySignupViewModel signupModel)
+        public IActionResult RegisterActivity(ActivitySignupViewModel signupModel)
         {
             if (signupModel == null)
             {
--- a/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
@@ -11,6 +11,8 @@ using System.Linq;
 using System.Net;
 using System.Security.Claims;
 using System.Threading.Tasks;
+using AllReady.Extensions;
+using AllReady.Features.Activity;
 using AllReady.Features.Notifications;
 using MediatR;
 
@@ -140,33 +142,24 @@ namespace AllReady.Controllers
             }
         }
 
-        [HttpPost("{id}/signup")]
-        [Authorize] 
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> RegisterActivity(int id)
+        [HttpPost("signup")]
+        [Authorize]
+        public async Task<IActionResult> RegisterActivity(ActivitySignupViewModel signupModel)
         {
-            var user = _allReadyDataAccess.GetUser(User.GetUserId());
-
-            var activity = _allReadyDataAccess.GetActivity(id);
-
-            if (activity == null)
+            if (signupModel == null)
             {
-                return HttpNotFound();
+                return HttpBadRequest();
             }
 
-            if (activity.UsersSignedUp == null)
+            if (!ModelState.IsValid)
             {
-                activity.UsersSignedUp = new List<ActivitySignup>();
+                // this condition should never be hit because client side validation is being performed
+                // but just to cover the bases, if this does happen send the erros to the client
+                return Json(new { errors = ModelState.GetErrorMessages() });
             }
 
-            activity.UsersSignedUp.Add(new ActivitySignup
-            {
-                Activity = activity,
-                User = user,
-                SignupDateTime = DateTime.UtcNow
-            });
-
-            await _allReadyDataAccess.UpdateActivity(activity);
+            _bus.Send(new ActivitySignupCommand() { ActivitySignup = signupModel });
             return new HttpStatusCodeResult((int)HttpStatusCode.OK);
         }
 
--- a/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
@@ -159,7 +159,7 @@ namespace AllReady.Controllers
                 return Json(new { errors = ModelState.GetErrorMessages() });
             }
 
-            _bus.Send(new ActivitySignupCommand() { ActivitySignup = signupModel });
+            _bus.SendAsync(new ActivitySignupCommand() { ActivitySignup = signupModel });
             return new HttpStatusCodeResult((int)HttpStatusCode.OK);
         }
 
--- a/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
+++ b/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
@@ -41,18 +41,18 @@ namespace AllReady.Models
 
         Task IAllReadyDataAccess.DeleteActivityAndTaskSignupsAsync(int activitySignupId)
         {
-            var activity = _dbContext.ActivitySignup.SingleOrDefault(c => c.Id == activitySignupId);
+            var activitySignup = _dbContext.ActivitySignup.SingleOrDefault(c => c.Id == activitySignupId);
 
-            if (activity == null)
+            if (activitySignup == null)
             {
                 return null;
             }
             
-            _dbContext.ActivitySignup.Remove(activity);
+            _dbContext.ActivitySignup.Remove(activitySignup);
 
             _dbContext.TaskSignups.RemoveRange(_dbContext.TaskSignups
-                .Where(e => e.Task.Activity.Id == activity.Activity.Id)
-                .Where(e => e.User.Id == activity.User.Id));
+                .Where(e => e.Task.Activity.Id == activitySignup.Activity.Id)
+                .Where(e => e.User.Id == activitySignup.User.Id));
 
             return _dbContext.SaveChangesAsync();
         }
--- a/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
+++ b/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
@@ -39,24 +39,31 @@ namespace AllReady.Models
             return _dbContext.SaveChangesAsync();
         }
 
-        Task IAllReadyDataAccess.DeleteActivitySignupAsync(int activitySignupId)
+        Task IAllReadyDataAccess.DeleteActivityAndTaskSignupsAsync(int activitySignupId)
         {
             var activity = _dbContext.ActivitySignup.SingleOrDefault(c => c.Id == activitySignupId);
 
-            if (activity != null)
+            if (activity == null) return null;
             {
                 _dbContext.ActivitySignup.Remove(activity);
+
                 var signupIds = _dbContext.TaskSignups
                     .Where(e => e.Task.Activity.Id == activity.Activity.Id)
                     .Where(e => e.User.Id == activity.User.Id)
                     .Select(e => e.Id);
 
-                var tasks = signupIds.Select(id => (this as IAllReadyDataAccess).DeleteTaskAsync(id));
-                Task.WaitAll(tasks.ToArray());
-                
+
+                foreach (var signupId in signupIds)
+                {
+                    var taskSignup = _dbContext.TaskSignups.SingleOrDefault(c => c.Id == signupId);
+                    if (taskSignup != null)
+                    {
+                        _dbContext.TaskSignups.Remove(taskSignup);
+                    }
+                }
+                    
                 return _dbContext.SaveChangesAsync();
             }
-            return null;
         }
 
         Task IAllReadyDataAccess.UpdateActivitySignupAsync(ActivitySignup value)
--- a/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
+++ b/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
@@ -43,27 +43,30 @@ namespace AllReady.Models
         {
             var activity = _dbContext.ActivitySignup.SingleOrDefault(c => c.Id == activitySignupId);
 
-            if (activity == null) return null;
+            if (activity == null)
             {
-                _dbContext.ActivitySignup.Remove(activity);
-
-                var signupIds = _dbContext.TaskSignups
-                    .Where(e => e.Task.Activity.Id == activity.Activity.Id)
-                    .Where(e => e.User.Id == activity.User.Id)
-                    .Select(e => e.Id);
-
-
-                foreach (var signupId in signupIds)
-                {
-                    var taskSignup = _dbContext.TaskSignups.SingleOrDefault(c => c.Id == signupId);
-                    if (taskSignup != null)
-                    {
-                        _dbContext.TaskSignups.Remove(taskSignup);
-                    }
-                }
-                    
-                return _dbContext.SaveChangesAsync();
+                return null;
             }
+            
+            _dbContext.ActivitySignup.Remove(activity);
+
+            var signupIds = _dbContext.TaskSignups
+                .Where(e => e.Task.Activity.Id == activity.Activity.Id)
+                .Where(e => e.User.Id == activity.User.Id)
+                .Select(e => e.Id);
+
+
+            foreach (var signupId in signupIds)
+            {
+                var taskSignup = _dbContext.TaskSignups.SingleOrDefault(c => c.Id == signupId);
+                if (taskSignup != null)
+                {
+                    _dbContext.TaskSignups.Remove(taskSignup);
+                }
+            }
+                    
+            return _dbContext.SaveChangesAsync();
+            
         }
 
         Task IAllReadyDataAccess.UpdateActivitySignupAsync(ActivitySignup value)
--- a/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
+++ b/AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
@@ -50,23 +50,11 @@ namespace AllReady.Models
             
             _dbContext.ActivitySignup.Remove(activity);
 
-            var signupIds = _dbContext.TaskSignups
+            _dbContext.TaskSignups.RemoveRange(_dbContext.TaskSignups
                 .Where(e => e.Task.Activity.Id == activity.Activity.Id)
-                .Where(e => e.User.Id == activity.User.Id)
-                .Select(e => e.Id);
+                .Where(e => e.User.Id == activity.User.Id));
 
-
-            foreach (var signupId in signupIds)
-            {
-                var taskSignup = _dbContext.TaskSignups.SingleOrDefault(c => c.Id == signupId);
-                if (taskSignup != null)
-                {
-                    _dbContext.TaskSignups.Remove(taskSignup);
-                }
-            }
-                    
             return _dbContext.SaveChangesAsync();
-            
         }
 
         Task IAllReadyDataAccess.UpdateActivitySignupAsync(ActivitySignup value)
--- a/AllReadyApp/Web-App/AllReady/Startup.cs
+++ b/AllReadyApp/Web-App/AllReady/Startup.cs
@@ -19,7 +19,6 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.PlatformAbstractions;
 using AllReady.Security;
-using Newtonsoft.Json.Serialization;
 
 namespace AllReady
 {
--- a/AllReadyApp/Web-App/AllReady/Startup.cs
+++ b/AllReadyApp/Web-App/AllReady/Startup.cs
@@ -19,6 +19,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.PlatformAbstractions;
 using AllReady.Security;
+using Newtonsoft.Json.Serialization;
 
 namespace AllReady
 {
--- a/AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
@@ -126,7 +126,7 @@
                     if (!Model.IsClosed)
                      {
                          <h2>
-                             <a asp-action="Login" asp-controller="Account" asp-route-ReturnUrl="/Activity/@Model.Id" class="btn btn-default  btn-volunteer">Volunteer</a>
+                             <a asp-action="Login" asp-controller="Account" asp-route-ReturnUrl="/Activity/@Model.Id" class="btn btn-default  btn-volunteer">Volunteer<br /> for this Activity</a>
                          </h2>
                      }
                      else
@@ -139,7 +139,7 @@
                     if (!Model.IsClosed)
                      {
                          <h2>
-                             <button class="btn btn-default btn-volunteer" data-bind="visible: !enrolled(), click: signupForActivity" style="display: none">Volunteer</button>
+                             <button class="btn btn-default btn-volunteer" data-bind="visible: !enrolled(), click: signupForActivity" style="display: none">Volunteer<br /> for this Activity</button>
                          </h2>
                      }
                      else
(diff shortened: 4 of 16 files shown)


MisterJames:
We'll have to sort out the build failure before merging this one. I'll start investigating that now.

MisterJames:
@mmoore99 It looks like the test that is failing is in the neighbourhood of activity signup. Could you please evaluate this further?

mmoore99:
@MisterJames I have investigated this  a little further.  It appears that the test is failing while trying to confirm the ability for the admin to **create an activity** ("Test: The Org Admin can create activies").  Nothing in this PR involves that process.  

However, this PR depends on PR #514, which has not yet been merged.  #514 does make some changes to the Activity create process as it has added the "ActivityType" field.  So, possibly, this error will go away once #514 is merged?  Just sort of guessing here since I am not really familiar with the scenario test process.

mmoore99:
@MisterJames upon further investigation, I think that "Test: The Org Admin can create activities"  may be failing since in #514 there was some validation added to the "Add Activity" form.  The "ActivityType" property was added and is "Required".  Also, the "NumberOfVolunteersRequired " was made "Required". Since the scenario test does not populate either of these fields, the form submission results in validation errors instead of a successful Activity creation.

Is it possible to run the scenario tests locally?

mmoore99:
@MisterJames I pushed a change to #514 in an attempt to fix the scenario test.  Please see comment under that PR.

tonysurma:
along with the comment on 514 I would like to merge this after that but am not sure if the test error - anyone have ideas how to resolve? thanks.

tonysurma:
same here @mheggeseth @BillWagner @MisterJames will merge this after we get a verdict on the other and the fail here.  THanks

mheggeseth:
Looks like this could still use some code review. I'm happy to do it.

mmoore99:
@mheggeseth do you want me to do a rebase on this since the commits from #514 are still showing in this PR?

mheggeseth:
@mmoore99 Sure, go ahead and rebase.

mmoore99:
@mheggeseth rebase completed.

AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
```diff
@@ -120,34 +122,35 @@
...
-                        </h2>
-                    }
-                    else
-                    {
-                        <h2 data-bind="visible: !enrolled()"> This activity is closed</h2>
-                    }
+                    if (!Model.IsClosed)
+                     {
+                         <h2>
+                             <button class="btn btn-primary btn-volunteer" data-bind="visible: !enrolled(), click: signupForActivity" style="display: none">Volunteer<br /> for this Activity</button>
```
mheggeseth: The Volunteer button shouldn't look different depending on whether you are signed in. I think just saying "Volunteer" is sufficient. I would also make the button orange to be consistent with the rest of the action buttons on the site.

Signed in:
image


Not signed in:
image

mmoore99: @mheggeseth I'll change the color as suggested. I agree that the button should be the same whether signed in or not.  Regarding the wording, I made the change in wording specifically in response to @tonysurma's request in #502:

**`Volunteer link is present and worded with clarity not just 'volunteer' but something like 'Volunteer for this Campaign'`**

I'll be glad to change it to whatever is decided.  I'll let you make the call.

AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
```diff
@@ -292,13 +301,14 @@
...
+                        signupModelSeed = Model.SignupModel
                     });
 
-
-        $('#phone-input').mask("(999) 999-9999");
-            
-    </script>
+        </script>
+    <script type="text/javascript" src="~/js/ko.bindings.js"></script>
+    <script type="text/javascript" src="~/lib/knockout.mapping/knockout.mapping.js"></script>
```
mheggeseth: Looks like you forgot to commit the bower.json changes that add `knockout.mapping`.

mmoore99: @mheggeseth the reason for this is that when I went to bower to try to find the  knockout-mapping plugin I found three entries, all of which were forks of Steve Sanderson's "official" plugin.  The three that I found are [here](https://github.com/crissdev/knockout.mapping), [here](https://github.com/LRPalacios/knockout.mapping) and [here](https://github.com/crissdev/knockout.mapping).  Steve Sanderson's version is [here](https://github.com/SteveSanderson/knockout.mapping).  So, instead of using one of the forks i just downloaded Sanderson's.

Do you want me to swtch to one of the forks or just continue with the downloaded version?

mmoore99:
@mheggeseth I committed changes for the color and wording of volunteer button as suggested.  If the wording on the button needs to be changed to something else, let me know.

AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
mheggeseth: You can keep using the downloaded version. You will just have to explicitly add knockout.mapping.js` to the git repo because `.gitignore` is set up to ignore everything in `~/lib`.

AllReadyApp/Web-App/AllReady/Startup.cs
```diff
@@ -19,6 +19,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.PlatformAbstractions;
 using AllReady.Security;
+using Newtonsoft.Json.Serialization;
```
mheggeseth: Don't think this is necessary.

mmoore99:
@mheggeseth I pushed changes to remove unused `using` in startup.cs and added `ko.mapping` files to source control

mheggeseth:
Fix the punctuation here. The "...signed up for this activity" sentence doesn't end and "We hope to see you soon" ends twice.
image

AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
```diff
@@ -7,6 +7,8 @@
     var userTimeZoneId = User.GetTimeZoneId();
 }
 
+<div id="alert_placeholder"></div>
```
mmoore99: @mheggeseth Yes, I agree. I had thought about that at one time, but forgot to come back to it.  I'll work on some changes.

AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
```diff
@@ -140,33 +142,24 @@ public ActionResult GetCheckin(int id)
             }
         }
 
-        [HttpPost("{id}/signup")]
-        [Authorize] 
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> RegisterActivity(int id)
+        [HttpPost("signup")]
+        [Authorize]
+        public async Task<IActionResult> RegisterActivity(ActivitySignupViewModel signupModel)
```
mheggeseth: This doesn't need to be async because you aren't awaiting anything in the body.

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
```diff
@@ -39,24 +39,31 @@ Task IAllReadyDataAccess.AddActivitySignupAsync(ActivitySignup userSignup)
             return _dbContext.SaveChangesAsync();
         }
 
-        Task IAllReadyDataAccess.DeleteActivitySignupAsync(int activitySignupId)
+        Task IAllReadyDataAccess.DeleteActivityAndTaskSignupsAsync(int activitySignupId)
         {
             var activity = _dbContext.ActivitySignup.SingleOrDefault(c => c.Id == activitySignupId);
 
-            if (activity != null)
+            if (activity == null) return null;
```
mheggeseth: Although technically correct, the way the code is formatted makes your intentions somewhat unclear ("do you mean to return null when activity is null or execute the block of code?"). Something like this would be better:

``` CSharp
if (activity == null)
{
  return null; //Purposely on its own line to draw attention to the fact that this is an exit point
}

//The real work...
```

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
```diff
@@ -39,24 +39,31 @@ Task IAllReadyDataAccess.AddActivitySignupAsync(ActivitySignup userSignup)
             return _dbContext.SaveChangesAsync();
         }
 
-        Task IAllReadyDataAccess.DeleteActivitySignupAsync(int activitySignupId)
+        Task IAllReadyDataAccess.DeleteActivityAndTaskSignupsAsync(int activitySignupId)
         {
             var activity = _dbContext.ActivitySignup.SingleOrDefault(c => c.Id == activitySignupId);
```
mheggeseth: Local variable should probably be called `activitySignup` instead of `activity` since that's what it is.

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
```diff
@@ -39,24 +39,31 @@ Task IAllReadyDataAccess.AddActivitySignupAsync(ActivitySignup userSignup)
...
-                var tasks = signupIds.Select(id => (this as IAllReadyDataAccess).DeleteTaskAsync(id));
-                Task.WaitAll(tasks.ToArray());
-                
+
+                foreach (var signupId in signupIds)
+                {
+                    var taskSignup = _dbContext.TaskSignups.SingleOrDefault(c => c.Id == signupId);
+                    if (taskSignup != null)
+                    {
+                        _dbContext.TaskSignups.Remove(taskSignup);
```
mheggeseth: Instead of getting the TaskSignups, getting their IDs, then getting the TaskSignups again to delete them, can't you do something like: 
```
_dbContext.TaskSignups.RemoveRange(_dbContext.TaskSignups
    .Where(e => e.Task.Activity.Id == activity.Activity.Id)
    .Where(e => e.User.Id == activity.User.Id)
);
```

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
mmoore99: @mheggeseth agreed!

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
mmoore99: @mheggeseth yes, that is much cleaner.  Thanks.

AllReadyApp/Web-App/AllReady/wwwroot/js/activity.js
```diff
@@ -1,57 +1,118 @@
...
-                type: "POST",
-                url: '/api/activity/' + activity + '/signup',
-                contentType: "application/json",
-            }).then(function (data) {
-                activityViewModel.enrolled(true);
-            }).fail(function (fail) {
-                console.log(fail);
+        function activitySignupSuccess(signUpViewModel) {
+            self.skills().forEach(function(skill) {
+                if (signUpViewModel.AddSkillIds().indexOf(skill.Id) >= 0){ skill.IsUserSkill(true) }
```
mheggeseth: I can't find any evidence in the code where `AddSkillIds` becomes an observable array. I think it's just a plain array. Additionally, you should put the curly braces on separate lines and put a semicolon after `skill.IsUserSkill(true)`.

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.ActivitySignup.cs
mmoore99: @mheggeseth Agreed.  I had noticed this as well when I was working on PR #523 and made the change there.  I'll make the change here as well.

AllReadyApp/Web-App/AllReady/wwwroot/js/ko.bindings.js
```diff
@@ -0,0 +1,20 @@
...
+            $(element).mask(mask, { autoclear: false });
+        }
+        ko.utils.registerEventHandler(element, "blur", function () {
+            var observable = valueAccessor();
+            observable($(element).val());
+        });
+    },
+    update: function (element, valueAccessor) {
+        var value = ko.utils.unwrapObservable(valueAccessor());
+        $(element).val(value);
```
mheggeseth: So I like that you wrapped the `$().mask` plugin with a binding handler. However, the way you've written it you kind of have to implement your own `value` binding handler at the same time. If you search `knockout.debug.js` for `ko.bindingHandlers['value']` you'll see that you probably don't want to do that. I would instead suggest trying to get the mask plugin to play nicely with KO's standard value binding (if you haven't already tried that). There are a couple ways you could do that:
* `masked` could wrap the `value` binding and call through to `ko.bindingHandlers.value.init/update`. In this case I would suggest naming the binding `valueMasked` to more clearly convey that it is derivative of `value`
* `masked` could be its own binding independent of `value` and try not deal with the value of the `input` at all

Additionally, when you have a binding handler that takes multiple options, I think it's better practice for the bound value to be an object containing each of the options, rather than conveying those options through sibling bindings via allBindingsAccessor`: `masked: { placeholder: "", mask: "" }`

AllReadyApp/Web-App/AllReady/wwwroot/js/activity.js
```diff
@@ -1,57 +1,118 @@
...
+    function showalert(message, alerttype, timeoutInSecs) {
+        $('#alert_placeholder').append('<div id="alertdiv" class="alert ' +  alerttype + ' fade in"><a class="close" data-dismiss="alert">×</a><span>'+message+'</span></div>')
+        if (timeoutInSecs) {
+            setTimeout(function() {
+                $(".alert").alert('close');
+            }, timeoutInSecs * 1000);
+        };
+    };
+
+    SignupViewModel = function (signupModelSeed, unassociatedSkills) {
```
mheggeseth: This function will live in the global namespace unless you do one of these things:
* `var SignupViewModel = function (signupModelSeed, unassociatedSkills)`
*  `function SignupViewModel(signupModelSeed, unassociatedSkills)`

AllReadyApp/Web-App/AllReady/wwwroot/js/activity.js
```diff
@@ -1,57 +1,118 @@
...
+            validate: {
+                required: "'Email' is required",
+                email: ""
+            }
+        });
+
+        self.isValid = ko.computed(function () {
+            var allValidatablesAreValid = true;
+            for (var property in self) {
+                if (self.hasOwnProperty(property) && self[property]["hasError"]) {
```
mheggeseth: Use `typeof self[property].hasError === "function"`

AllReadyApp/Web-App/AllReady/wwwroot/js/activity.js
mmoore99: @mheggeseth `AddSkillIds` is an observable array.  It is created in the ko.mapping statement when the view model is created.  Since AddSkillIds is passed down from the server as an array, the ko.mapping makes it observable.
```
SignupViewModel = function (signupModelSeed, unassociatedSkills) {
        var self = this;
        ko.mapping.fromJS(signupModelSeed, {}, self);
        self.unassociatedSkills = unassociatedSkills;

I'll make the formatting changes as suggested.

File: AllReadyApp/Web-App/AllReady/wwwroot/js/ko.validation.js [hunk 13]
```diff
@@ -17,4 +17,56 @@ ko.observableArray.fn.uniqueValidator = function (property) {
                 return prev;
             }, []).length;
     });
+};
+
+//
+// This is a "poor man's" validation routine that provides specific, limited capabilities (currently only supports required, email, phoneNumber validations).
+// If more a more general & extensive validation capability is needed, consider using Knockout-Validation plugin at https://github.com/Knockout-Contrib/Knockout-Validation
+//
+ko.extenders.validate = function (target, options) {
```
mheggeseth: I like what you did here, but the syntax for using `validate` is a bit clunky. For instance, it's not very obvious what it means to set `phoneNumber` equal to the empty string. It kind of forces you to go read the implementation of `validate`. Additionally, the maintainability of the `validate` function is somewhat suspect. Do you just keep adding if blocks to it for each kind of validation you want to implement? And each of the if blocks has basically the same format and should probably be refactored into a single function.
```
self.PreferredPhoneNumber.extend({
    validate: {
        required: "'Phone Number' is required",
        phoneNumber: ""
    }
});
```

I think it might work better to use the `withValidator` fn in a way that you could then chain validation functions together something like
```
self.PreferredPhoneNumber
    .isRequired({ displayName: "Phone Number"})
    .validatePhoneNumber()
```

`withValidator` would need to be enhanced to be chainable. This isn't too bad. You just check whether `isValid` is already defined and combine its result with the result of `validatorFn` when you re-define `isValid` as a new computed observable. Notice I also changed the base type to be `subscribable` so you can use it with arrays or single values. This also achieves the same short-circuiting that your validate function does because `validateFn` will not execute unless `isValidAlready` returns true.
```
ko.subscribable.fn.withValidator = function (validatorFn) {
    var isValidAlready = this.isValid || function () { return true; } //Default to valid if no validator defined yet
    this.isValid = ko.computed(function () {
        return isValidAlready() && validatorFn(this);
    }, this);
    return this;
};
```

You could also pretty easily add an option to defer validation until the observable changes from its initial value. That seems basically what you are getting at with `showErrors`. If you want to work together on this, I'd be happy to.

AllReadyApp/Web-App/AllReady/Views/Shared/_Layout.cshtml
```diff
@@ -136,7 +136,6 @@
         </script>
     </environment>
     <script src="~/lib/knockout-bootstrap/src/knockout-bootstrap.js"></script>
-    <script src="~/lib/system.js/dist/system.js"></script>
```
mheggeseth: system.js is actually being used on Views\Campaign\Details.cshtml

mheggeseth:
I'm all for using AJAX for the volunteer workflow, but some things about the way it's implemented are perhaps a little more complicated than they need to be. I'm not sure the whole dynamic template rendering thing in `showModal` is necessary. I'm actually not sure `showModal` itself is necessary (or that its benefits are worth its added complexity). Another thing is that the "form" doesn't actually need to be a form anymore because it no longer has a submit action. Do you want to pair on this and see what we can simplify or get rid of?

mmoore99:
@mheggeseth sure, I would be glad to pair with you on this.  I debated as well about whether the `showModal` approach was necessary and would like to discuss it with you.  I am going to be tied up for the rest of the evening, but would have time tomorrow, if that works for you.  Let me know.

mheggeseth:
I'm available tomorrow morning starting at 9 central.

AllReadyApp/Web-App/AllReady/Controllers/ActivityApiController.cs
```diff
@@ -140,34 +142,25 @@ public ActionResult GetCheckin(int id)
...
-            {
-                Activity = activity,
-                User = user,
-                SignupDateTime = DateTime.UtcNow
-            });
-
-            await _allReadyDataAccess.UpdateActivity(activity);
-            return new HttpStatusCodeResult((int)HttpStatusCode.OK);
+            _bus.Send(new ActivitySignupCommand() { ActivitySignup = signupModel });
+            return new {Status = "success"};
```
mheggeseth: Despite the fact that it's not async anymore, I think you should still return an `IActionResult`. Just `return Json(null)` if everything went great.

mmoore99:
@mheggeseth How about 10 central?  I'll Skype you.

mheggeseth:
Good

AllReadyApp/Web-App/AllReady/Views/Activity/Activity.cshtml
mheggeseth: OK, you can change it back to "Volunteer for this activity"

tonysurma:
@mmoore99 and @mheggeseth did you get a chance to chat today?  Thanks.

mmoore99:
@tonysurma @mheggeseth  Yes, we talked today.  I've got some more work to do before we'll be ready to merge.  I'll update when ready.

tonysurma:
Thank you!  You really have been contributing an amazing amount @mmoore99

mmoore99:
@tonysurma You're welcome, I'm having fun.  Unlike your other contributors, I think, I'm an old retired guy that likes to write code.  Keeps the brain functioning :sunglasses:

mmoore99:
@mheggeseth I pushed the changes for the refactored `showModal`

mmoore99:
- @mheggeseth, cc:@tonysurma  I have pushed changes that I believe address all of the outstanding items that we have been discussing on this PR. 

- A couple of notes:
  - On the knockout custom binding for the jquery edit mask I removed the `placeholder` option as it is not needed since we can just use the normal html `placeholder` attribute.

  - @BillWagner  @SeanFeldman I rebased on top of the PR's that were merged last weekend, including the async stuff.  In `ActivitySignupHandler.cs` the call to save database changes was changed to `await _context.SaveChangesAsync();`. When testing, this statement throws the exception below.  I changed the code back to `_context.SaveChanges();` and it does not throw the exception.  So I am going to leave it this way for now until we can determine what is going on.

> Microsoft.Data.Entity.DbContext: Error: An exception occurred in the database while saving changes.
> Microsoft.Data.Entity.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> System.InvalidOperationException: Invalid attempt to call FieldCount when reader is closed.
>    at System.Data.SqlClient.SqlDataReader.get_FieldCount()
>    at Microsoft.Data.Entity.Storage.Internal.UntypedRelationalValueBufferFactory.Create(DbDataReader dataReader)
>    at Microsoft.Data.Entity.Update.AffectedCountModificationCommandBatch.<ConsumeResultSetWithPropagationAsync>d__8.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
>    at Microsoft.Data.Entity.Update.AffectedCountModificationCommandBatch.<ConsumeAsync>d__6.MoveNext()
>    --- End of inner exception stack trace ---
>    at Microsoft.Data.Entity.Update.AffectedCountModificationCommandBatch.<ConsumeAsync>d__6.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
>    at Microsoft.Data.Entity.Update.ReaderModificationCommandBatch.<ExecuteAsync>d__28.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
>    at Microsoft.Data.Entity.Update.Internal.BatchExecutor.<ExecuteAsync>d__1.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
>    at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.<SaveChangesAsync>d__32.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
>    at Microsoft.Data.Entity.DbContext.<SaveChangesAsync>d__28.MoveNext()
> Exception thrown: 'Microsoft.Data.Entity.DbUpdateException' in mscorlib.dll
>

mmoore99:
pushed changes that refactor `showModal` to have fluent api and better encapsulate modal hide function

SeanFeldman:
@mmoore99 not sure what's the issue. `ActivitySignupHandler` doesn't have a test :disappointed: so I manually ran through the scenario (volunteer enrollment into an activity) and it works ok. Is that exception happening all the time for you?

SeanFeldman:
Though I did miss one thing for sure:
```c#
await _context.SaveChangesAsync();
```
should be
```c#
await _context.SaveChangesAsync().ConfigureAwait(false);
```

mmoore99:
@SeanFeldman yes, it is happening all of the time when the async call is used. I tried it using `await _context.SaveChangesAsync().ConfigureAwait(false);` with the same result.  I also tried it with a new database, same result.  It works fine with the non async call.  Of course, I am running on a different code base than you are.  I have this PR plus another one,  #523 , which is coming right behind this one.  I would propose that we leave the non async call in until we get past these two PR's and then we can see where we're at.

SeanFeldman:
@mmoore99 sounds good. The fact that there's no unit test to run against makes it a bit more difficult.

tonysurma:
@mmoore99   please rebase off of master as we now have a fix for the build/test issues!

For reference on rebasing this page helps (replace upstream with HTBox in examples) https://robots.thoughtbot.com/keeping-a-github-fork-updated

mmoore99:
@tonysurma rebase on master complete

tonysurma:
@mheggeseth just dbl checking you are good to go on this and I will merge.  @mmoore99 thanks for the rebase

mheggeseth:
I haven't read all of @mmoore99's new changes but we've been in close contact so I good with this.

SeanFeldman:
@mheggeseth @mmoore99 is the async change got reverted back to be there or still requires attention.

mheggeseth:
@SeanFeldman The proposed code in `ActivitySignupHandler` is as you and @mmoore99 agreed upon yesterday: `_context.SaveChanges()` instead of `await _context.SaveChangesAsync()`.

However, I noticed something peculiar. I wonder whether the real problem here is that the sending of `ActivitySignupCommand` via `_bus.SendAsync()` isn't being "awaited" in `ActivityApiController.RegisterActivity`. I realize you probably _should_ be able to use Mediatr in this way, but I'd be interested to know if it works to change back to `await _context.SaveChangesAsync()` if you also `await _bus.SendAsync(new ActivitySignupCommand())`

SeanFeldman:
Asynchronous operation should always be awaited, unless the task is
returned directly to the caller. Could be the issue indeed.

mmoore99:
@mheggeseth @SeanFeldman.  It appears that @mheggeseth's comments regarding the lack of "await" on `_bus.SendAsync()`  in `ActivityApiController.RegisterActivity`being the problem are correct.  I put the "await" on that call and changed the call back to `await _context.SaveChangesAsync()` `ActivitySignupHandler` and everything worked properly.  Great catch @mheggeseth!  Since this PR has not been merged yet, I'll go ahead and push up these changes.

SeanFeldman:
:+1:

mheggeseth:
:shipit:

mmoore99:
@BillWagner @tonysurma Now that the review of this PR is complete, would it be possible to merge it? Both this PR and #523 (which is under active development) are in a lot of the same code areas and keeping the two branches alive and updated is time consuming.  Thanks.

Issue #524, opened 2016-02-20
Locked campaign appearing on Organization homepage


stevejgordon:
I forgot to filter out the locked campaigns on the org home page http://localhost:48408/Organization/{id} when implementing the locked campaign requirement.

I will have a PR to address this very soon.

stevejgordon:
I will take a few extra minutes to move this action to the command/query pattern whilst I'm there.

tonysurma:
closed in pull request #525

PR #525, opened 2016-02-20
Excluding locked campaigns from the results on the Organization page

stevejgordon:
Fixes issue #524 

- Moved the ShowOrganization action from direct use of context to the command/query pattern
- Implemented a handler which excluded locked campaigns in the OrganizationViewModel
- Added tests for the controller action and the handler


dpaquette:
Looks good to me :shipit:

Issue #503, opened 2016-01-30
Scenario A1: Administrator Creates a Campaign


tonysurma:
_These issues are scenarios documented for use of the application.  Likely will be the 'parent' of multiple sub issues to verify, add or fix functionality to support them._

# **Administrator Creates a Campaign**
Initiation: User registered and logged in as a Organization Administrator 
Features:
- [ ] Administrator can click link to create Campaign from Admin->Campaign and Campaign pages

_**Scenario to be fleshed out more**_

stevejgordon:
@tonysurma Currently creating a campaign from the Admin > Campaign page is already in place. So to meet this scenario as it stands we just need to create a link from the main Campaign page to job to the Add Campaign screen (visible to org and admins only)?

tonysurma:
@stevejgordon yes, the 'non-admin' campaign list page should get that add button with visibility as you stated.  Thanks

stevejgordon:
@tonysurma I think we can close this issue? Looks like it didn't auto close because of the wording of my PR.

tonysurma:
yep, closing this per your PR

PR #527, opened 2016-02-20
Adding create campaign button on the public Campaign page (visible to…

stevejgordon:
Resolves issue #503 

I have added a "Create Campaign" button on the public Campaign page (/Campaign). This button will be displayed if the current user is an org admin or a site admin.


tonysurma:
reviewed and so we can :shipit:

PR #554, opened 2016-02-20
Workaround to Postal Code DB error

### LINKED ISSUE #470: Invalid zip code in Campaign Edit page causes internal server error - remove error
Zip code field needs graceful error recovery on /Admin/Campaign/Edit/### page.

Error message:

A database operation failed while processing the request.

DbUpdateException: An error occurred while updating the entries. See the inner exception for details. 
SqlException: The INSERT statement conflicted with the FOREIGN KEY constraint "FK_Location_PostalCodeGeo_PostalCodePostalCode". The conflict occurred in database "AllReady", table "dbo.PostalCodeGeo", column 'PostalCode'. 
There are pending model changes for AllReadyContext
Scaffold a new migration for these changes and apply them to the database from the command line:

> dnx ef migrations add [migration name] 
> dnx ef database update

stevejgordon:
Fixes #470 

This is an initial workaround to solve the immediate issue of a DB error being thrown up to the UI when trying to save a campaign with a postal code set.

Before saving the campaign I have added a check to validate if the city, state and postal code combination are valid (only triggered if the postal code is supplied in the form) which will return a model error to the user if the postal code is not found.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignControllerTests.cs
@@ -307,7 +307,7 @@ namespace AllReady.UnitTest.Areas.Admin.Controllers
             var tid = organizationId;
             var mockMediator = new Mock<IMediator>();
             mockMediator.Setup(mock => mock.Send(It.IsAny<CampaignSummaryQuery>()))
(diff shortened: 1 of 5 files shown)


tonysurma:
@mheggeseth can I get a review?

mheggeseth:
Looks good. :shipit: I was a little unsure about creating a whole query for a supposedly temporary workaround, but it may be useful to have a query that tells you whether a postal code is valid so I'm good with keeping it.

PR #563, opened 2016-02-20
don't always launch browser when starting to debug

glittle:
Would start new browser on every F5.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Properties/launchSettings.json
+++ b/AllReadyApp/Web-App/AllReady/Properties/launchSettings.json
@@ -10,7 +10,6 @@
   "profiles": {
     "IIS Express": {
       "commandName": "IISExpress",
-      "launchBrowser": true,
       "environmentVariables": {
         "Hosting:Environment": "Development"
       }


mheggeseth:
Can you explain why this is a good change? Isn't is useful that VS launches a browser that takes you directly to the URL? This seems more like a workaround for not closing old browser tabs.

One useful side effect of launching a browser when pressing F5 is that if you set your default browser to IE (not that anyone would do such a thing), VS will automatically attach to the browser for script debugging so you can debug JavaScript in Visual Studio instead of in the F12 tools.

glittle:
It is likely a matter of preference. 

If you start your programming time by directly entering debug mode with F5, this would be good. However, I don't think that is a normal use-case.

I typically start by using the "View in Browser" context menu on the website project. That will start the IISExpress server and open a browser at the default page.  From there, I navigate the site until I'm on a page that I'm working on.

I'd then work on the fix, and refresh that page to see the results. If I have to resort to using Debug mode (which should be infrequent), I want the server to be in debug as I refresh the page I am working on.  With the settings the way they were, *another* tab is automatically opened back at the default page. I need to ignore or close that tab and refresh the page that I'm working on.

So having the Debug mode open a new browser tab back at the default page is almost always a nuisance!

mheggeseth:
@glittle Since this seems to be a personal workflow preference, that suggests it should not be made permanent in the repo. Advanced users will continue to be able to customize their workflow however they like but the repository should be optimized for newbies. Having Visual Studio not launch a browser (which is not the default) when you press F5 will likely make new contributors think they have set up something wrong.

glittle:
Go ahead and reject this... it is not a big deal!  Users can always make such changes if they wish to in their local copy. :)

mheggeseth:
@glittle I'm not able to reject it. Can you close it yourself?

glittle:
I'll give it a try...

PR #582, opened 2016-02-26
Resolve AppVeyor Unit Test Deadlocks

mheggeseth:
I think the unit tests are deadlocking when AppVeyor runs them. I came across this blog post last night (http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html) and it seems to explain pretty well some of the things we are doing.

**TL;DR** If you `.Result` or `.Wait()` a task that wasn't executed/awaited with `.ConfigureAwait(false)` to indicate that it should continue execution in a different context than the captured one, you will get a deadlock. The `.Result/Wait()` is blocking the context waiting for the task to finish and the task is waiting for the context to get freed to it can notify somebody that it's done.

My hunch is that tests are passing on dev machines because they are just faster than AppVeyor and the async task actually completes before `.Result` or `.Wait()` are called.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/ActivityApiControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/ActivityApiControllerTests.cs
@@ -112,7 +112,7 @@ namespace AllReady.UnitTest
                 .WithUser(recordId.ToString());
 
             // Act
-            var result = await controller.UnregisterActivity(recordId);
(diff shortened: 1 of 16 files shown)


mheggeseth:
Well, that didn't work. How about this...

mheggeseth:
:boom:

mheggeseth:
No, sir. :shipit:

:microphone: :arrow_down: :v:

PR #589, opened 2016-03-05
Feature 212 display geo info

### LINKED ISSUE #212: Display location on the activities page for a campaign
Recommend displaying the location (at least the city & state detail in text - or ideally neighborhood if that field is included in the activity setup page for admins) next to the Map link on the activities summary page for a campaign to allow users to identify applicable locations with a minimum number of clicks.

image

MisterJames:
The campaign details page now shows the location of the campaign and of each of the activities.

image

 - Extended the view models
 - Added tests  for VMs
 - Included the location data in the data context

This controller is still using the data access repository, but changing to queries/handlers was outside the scope of this change.

I think that these VMs could be refactored as well as they are being reused a fair bit and many of the properties (and subqueries used to fetch them) are not even being used. This should be addressed in the switch to queries/handlers.

Closes #212

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/ViewModels/CampaignViewModelShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/ViewModels/CampaignViewModelShould.cs
@@ -1,0 +1,64 @@
+﻿using AllReady.ViewModels;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace AllReady.UnitTest.ViewModels
+{
+    public class CampaignViewModelShould
+    {
+        [Fact]
+        public void ReturnCityAndStateWhenBothArePresent()
+        {
+            // arrange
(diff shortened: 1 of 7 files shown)


mheggeseth:
I know you are purposely resisting the temptation to refactor here, but you have added the exact same Summary getter (with identical tests) in two classes. Is there a good reason why that getter couldn't live on the Location class (or an extension thereof)?

tonysurma:
@mheggeseth and @MisterJames is this good to pull in after a rebase?

mheggeseth:
Just need an answer on whether Location Summary can be refactored.

MisterJames:
@mheggeseth Yeah, it was just getting too big. I will refactor both of those out, consolidate, update the tests and pull up the command/query/handler pattern on a seperate PR. Thanks for the review! @tonysurma I'll rebase this evening after work.

MisterJames:
@mheggeseth this should be good to go!

mheggeseth:
OK :shipit:

Issue #595, opened 2016-03-10
ActivityController: add missing unit tests and refactor to mediator


mgmccarthy:
Issues #567 and #59

mgmccarthy:
closing this out as it looks like @mikesigs just submitted a PR that takes care of the refactor to mediator.

PR #596, opened 2016-03-11
Refactored ActivityController to use Mediator

mikesigs:
I moved all/most of the logic from the ActivityController into separate Command+Handler pairs in Features/Activity. There's also some new unit tests.

I also added the Shouldly assertion library and used it in a couple tests.


mikesigs:
Closes #595

mheggeseth:
@mikesigs FYI, the auto-close syntax ("fixes #[issue]") only happens to work if you put in the PR description. I guess this doesn't really matter because you've already closed #595, but you could reopen it just to try out the feature.

MisterJames:
Good stuff, thanks for the PR @mikesigs!

mikesigs:
No problem. Looking forward to contributing more.

PR #606, opened 2016-03-14
Assignee and Status in Admin Activity Task List

### LINKED ISSUE #561: Assignee and Status not showing  on Activity Details page
On the Activity Details page `http://allready-d.azurewebsites.net/Admin/Activity/Details/16` the assignee and status aren't showing even though there was a volunteer assigned to it and the task was completed on the Activity page `http://allready-d.azurewebsites.net/Activity/16` See 2 screenshots below:

image

image

stevejgordon:
Fixes #561 

Updated the query handler to include the related AssignedVolunteers and load the model appropriately as used in other areas of the application.

Updated the view to use the status when a single assignee is set.

I wasn't 100% clear on the expected logic for the HasVolunteered Boolean. For now I've followed it's use elsewhere and set to true when populating the model. If this use is incorrect I can amend.

It feels like it might be good to document some of the this with either XML comments or .md pages as well since some properties can be a little confusing in their expected usage - at least I found it so in this case!


tonysurma:
pinging @mheggeseth @dpaquette @MisterJames for pr review pls

MisterJames:
@stevejgordon This is good for now, and I think you've captured the correct behaviour. #sweet 

:shipit: 

If we notice that this wasn't the intent we can add a defect and throw it up for grabs.

PR #607, opened 2016-03-16
TaskApiController: adding unit tests and refactoring to mediator

### LINKED ISSUE #604: TaskApiController: adding unit tests and refactoring to mediator
@mmoore99 @mikesigs 
Issue #59 and Issue #567

mgmccarthy:
Fixes #604

A couple notes:

First, I added an abstraction called `IDetermineIfATaskIsEditable` that is injected into TaskApiController via constructor injection. The implementation is in the `DetermineIfATaskIsEditable` class. The goal here was to encapsulate, the code that previously was in TaskApiController's `private bool HasTaskEditPermissions(AllReadyTask task)` method.

Breaking this out into a separate class allowed me to unit test whether a user has permissions to edit the task or if the AllReadyTask is in an editable state. 

That freed the Post, Put and Delete methods from having to worry about the setup and outcome of this check, as well as I only had to write the unit test for "is editable" once instead of every time I had to write a unit test against any of these three methods in the controller.

Second, as much as I wanted to, I could not remove the injection of IAllReadyDataAccess from TaskApiController's constructor, even moving 100% to mediator.

The reason is this call in the fist line of the Post method:
`task.ToModel(_allReadyDataAccess);`

'ToModel' does a BUNCH of work turning the TaskViewModel into an AllReadyTask, including making database queires via IAllReadyDataAccess. There are also a larege number of other things going on in here, so much so, that I would say someone needs to take a look at this code, and either start to decompose it, or change some of these rules to live somehwere else. I don't know enough about the app to make those calls.

Two good steps would be:
- removing the injected IAllReadyDataAccess
- writing unit tests around this view model class

that could start the ball rolling.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Features/Tasks/AddTaskCommandHandlerAsyncTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Features/Tasks/AddTaskCommandHandlerAsyncTests.cs
@@ -1,0 +1,22 @@
+﻿using System.Threading.Tasks;
+using AllReady.Features.Tasks;
+using AllReady.Models;
+using Moq;
+using Xunit;
+
+namespace AllReady.UnitTest.Features.Tasks
+{
+    public class AddTaskCommandHandlerAsyncTests
+    {
+        [Fact]
+        public async Task HandleInvokesAddTaskAsyncWithCorrectData()
+        {
+            var message = new AddTaskCommandAsync { AllReadyTask = new AllReadyTask() };
+            var dataAccess = new Mock<IAllReadyDataAccess>();
+            var sut = new AddTaskCommandHandlerAsync(dataAccess.Object);
+            await sut.Handle(message);
+
+            dataAccess.Verify(x => x.AddTaskAsync(message.AllReadyTask), Times.Once);
+        }
+    }
+}
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Tasks/TaskStatusChangeHandlerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Tasks/TaskStatusChangeHandlerTests.cs
@@ -88,15 +88,14 @@ namespace AllReady.UnitTest.Areas.Admin.Features.Tasks
 
             var task = Context.Tasks.First();
             var user = Context.Users.First();
-            var command = new TaskStatusChangeCommand
+            var command = new TaskStatusChangeCommandAsync
             {
                 TaskId = task.Id,
                 UserId = user.Id,
                 TaskStatus = TaskStatus.Accepted
             };
-
-            var handler = new TaskStatusChangeHandler(Context, mediator.Object);
-            await handler.Handle(command);
+            var handler = new TaskStatusChangeHandlerAsync(Context, mediator.Object);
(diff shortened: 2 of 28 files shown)


tonysurma:
pinging @mheggeseth @dpaquette @MisterJames for pr review pls

MisterJames:
Lots of moving parts here, I'm pulling down this PR and starting to work through it. Thanks for the huge effort here, @mgmccarthy.

AllReadyApp/Web-App/AllReady.UnitTest/Features/Tasks/AddTaskCommandHandlerAsyncTests.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Threading.Tasks;
+using AllReady.Features.Tasks;
+using AllReady.Models;
+using Moq;
+using Xunit;
+
+namespace AllReady.UnitTest.Features.Tasks
+{
+    public class AddTaskCommandHandlerAsyncTests
```
MisterJames: Just a note on naming tests...not needed at all, but if you use the class name as part of the test intent description, you get `class AddTaskCommandHandlerShould` and then a test name of `InvokeAddTaskHandlerWithCorrectData` and the two read like a sentence in test suites.

mgmccarthy:
@MisterJames , sounds good to me. I can refactor and submit this change against this current PR if you want...

if you don't want the change for this PR, I'll move to this naming convention for any new work...

where i work now, we actually use static classes to break down the naming of the unit tests... so in your example the test class would look like this:
```
public static AddTaskCommandHandlerAsyncTests
{
    public class AddTaskCommandHandlerShould
    {
        [Fact]
        public async Task InvokeAddTaskHandlerWithCorrectData()
        {
            ...
        }

        public async Task [MORE_TESTS]... etc...
    }
}
```

I am unit NUnit at my job now, so maybe this setup doesn't make sense as xUnit does not seem to have a [SetUp] attribute and a [TestFixture] attribute like NUnit does.

mgmccarthy:
ping @MisterJames. Any movement on merging this PR in?

mgmccarthy:
@MisterJames, would it be easier in the future (since I commit locally so much) to "squash" an PR this big? Not too sure if having all the separate commits makes it harder to review the code than one commit with all the change regardless of my local commits

tonysurma:
@MisterJames how should we move fwd on this one?

mgmccarthy:
@MisterJames @tonysurma. Any plans on accepting this PR? If you need me to make any code changes, just let me know

BillWagner:
@mgmccarthy Thanks for your patience.  Reviews are taking longer than we would like this week. A combination of vacations, build, and new jobs has slowed up many things.

mgmccarthy:
@BillWagner let me know if you want me to slow my rate of PR's against this repository. Just trying to the stubs set up so other devs can finish off these unit tests

BillWagner:
@mgmccarthy Normally This one is just quite large, and since @MisterJames has started reviewing, I'd like to have him weigh in. He should have time shortly.

mgmccarthy:
ping @MisterJames. any movement on this PR?

MisterJames:
@mgmccarthy Apologies as this was a large one with many moving parts. Code wise, things are in pretty good shape, it's the functionality that is a little harder to vet. I am looking at this again this evening.

Issue #340, opened 2015-11-13
Introduce constants for 'magic strings'


BillWagner:
Throughout the code, we use a number of literal string constants.
It would be great to replace those globally with defined string constants. See comments on PR #332 for the start of the conversation.

bcbeatty:
I've used https://github.com/T4MVC/T4MVC in the past for MVC projects. It works great for getting rid of magic strings of controllers and views. Not sure if it's been updated to support tag helpers

joelhulen:
@MisterJames and I had this conversation a couple of times. We discussed creating a static class named something like "EnumsAndStructs" or "WellKnownStrings", something along those lines. The idea is that some of these once-magic strings may be useful in other areas besides just views.

dpaquette:
Don't forget about the new `NameOf` operator in C#6. This can sometimes be used to avoid magic strings for View names in Controller.

Kritner:
Sorry for the spam... I didn't realize that referencing issues on my own fork would hit the issue on origin.  Still learning the flow of git/hub :)

I figured it wouldn't hit the issue log until my PR.

mheggeseth:
Don't worry about it. We don't have established guidelines for when to reference issues in commits. Actually, I think the information is helpful because it shows an issue is being worked on even if a PR is not imminent.

ultrabert:
Hi all

I'm looking for a jump-in issue. Is this still alive?

(And if so, does anyone have a better method of finding all string literals than regular expression "Find In Files")

Cheers,
Brett.

tonysurma:
@ultrabert yes still alive, thx in advance

dangle1:
I'd like to take a stab at this as well. To avoid duplication of effort, I'm letting everyone know I'm working on the ManageController and will be referencing the issue during commits, as per mheggesth's comment from 12/13/15.

dangle1:
Some things to note, 

Not sure how thorough we want to be in converting these magic strings, but Controller names can't take advantage of the nameof operator as a consequence of MVC's convention for expecting the Controller's name without the "Controller". Others have addressed this with extension methods: http://stackoverflow.com/questions/27444121/how-to-use-c-sharp-nameof-with-asp-net-mvc-url-action. 

There's a lot of usage of ViewData throughout the controllers and views, any reason for not using viewmodels instead? If there is going to be extensive use of ViewData, JoelHulen's suggestion of implementing a static class with enums and consts would mitigate typos when keying into ViewData.

[...]

PR #611, opened 2016-03-22
Introduce constants for 'magic strings' in ManageController.

dangle1:
Forgive me if this is an insignificant number of changes for a single pull-request. Lot of firsts for me here: contributing to an open source project, using git and github, so I just wanted to make sure I get the workflow down before I start making bigger changes. 

Please let me know if I should come back with a different pull-request when there is more to review. 

#340

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
@@ -14,7 +14,33 @@ namespace AllReady.Controllers
     [Authorize]
     public class ManageController : Controller
     {
+        //Email Constants
         private const string EMAIL_CONFIRMATION_SUBJECT = "Confirm your allReady account";
+        private const string NEW_EMAIL_CONFIRM = 
+            "Please confirm your new email address for your allReady account by clicking this link: <a href={0}>link</a>. Note that once confirmed your original email address will cease to be valid as your username.";
+        private const string RESEND_EMAIL_CONFIRM = "Please confirm your allReady account by clicking this link: <a href={0}>link</a>";
+        //Error Constants
+        private const string ACCOUNT_SECURITY_CODE = "Your allReady account security code is: ";
+        private const string FAILED_TO_VERIFY_PHONE_NUMBER = "Failed to verify phone number";
+        private const string PASSWORD_INCORRECT = "The password supplied is not correct";
+        private const string EMAIL_ALREADY_REGISTERED = "The email supplied is already registered";
+        //ViewData Constants
+        private const string SHOW_REMOVE_BUTTON = "ShowRemoveButton";
+        private const string STATUS_MESSAGE = "StatusMessage";
+        //Message Constants
+        private const string PASSWORD_CHANGED = "Your password has been changed.";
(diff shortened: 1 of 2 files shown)


mheggeseth:
I :heart: small PRs

mheggeseth:
This is a great start, but I think you could take string abstraction a little further in `ManageConrroller`. There are a few more strings that could use constants.

dangle1:
Okay, something like creating a constant for every string, similar to `EMAIL_CONFIRMATION_SUBJECT` ? 

Or something more robust for things like Controller names? Keep in mind `nameof()` won't work with Controller classes as MVC expects the "Controller" part of the name to be removed when being passed around. 

Otherwise, I can list all the strings throughout `ManageController` as consts fields.

mheggeseth:
I was thinking a constant for each string

dangle1:
Sure thing! Thanks for the feedback and I'll get back to implementing it now.

dangle1:
@mheggeseth each string has been constant-fied. Please let me know what you think.

tonysurma:
Once @mheggeseth does a review I can merge it in.  thx

AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
```diff
@@ -14,7 +14,33 @@ namespace AllReady.Controllers
     [Authorize]
     public class ManageController : Controller
     {
+        //Email Constants
         private const string EMAIL_CONFIRMATION_SUBJECT = "Confirm your allReady account";
+        private const string NEW_EMAIL_CONFIRM = 
+            "Please confirm your new email address for your allReady account by clicking this link: <a href={0}>link</a>. Note that once confirmed your original email address will cease to be valid as your username.";
+        private const string RESEND_EMAIL_CONFIRM = "Please confirm your allReady account by clicking this link: <a href={0}>link</a>";
```
mheggeseth: You should wrap the href values in literal quotes (at least to be consistent with what was there before): `<a href=\"{0}\">link</a>`

dangle1: Sure, thanks for catching that.

dangle1:
@mheggeseth literal quotes added as per your feedback.

mheggeseth:
Looks good :shipit:

mgmccarthy:
@mheggeseth @tonysurma @BillWagner

Why are we changing strings to constants in files that only use the constant once?

IMO, it's easier on the eyes to "read a human readable string" then a BUNCH_OF_CONSTANT_NAMES in a file.. I can see where a constant comes in handy when it's used more than once in a given file, but if it's used ONLY once, what's the benefit?

I understood Issue #340 to be related to getting rid of magic strings for things like using the new `nameof` operator, that would give us compile time checks against items being renamed in the system instead of relying on runtime failures. To me, we get value out of a change like that.

Unless I'm missing something

mheggeseth:
@mgmccarthy I don't feel strongly about it. Personally, I probably wouldn't have made constants for the single-use strings. Having to jump to another place in the file to read the string value is a little annoying but having a constant for every string serves as a good reminder to consider refactoring strings. It also makes it easier to do so when a single-use string becomes multi-use.

Issue #613, opened 2016-03-22
fix methods that return Task to return Task.FromResult(0) instead of null


mgmccarthy:
methods that return a Task should not return null:
```
Task IAllReadyDataAccess.DeleteTaskSignupAsync(int taskSignupId)
{
    var taskSignup = _dbContext.TaskSignups.SingleOrDefault(c => c.Id == taskSignupId);

    if (taskSignup != null)
    {
        _dbContext.TaskSignups.Remove(taskSignup);
        return _dbContext.SaveChangesAsync();
    }
    return null;
}
```

Instead of returning `null`, we should be returning `Task.FromResult(0)`

mgmccarthy:
methods that return a Task should not return null:
```
Task IAllReadyDataAccess.DeleteTaskSignupAsync(int taskSignupId)
{
    var taskSignup = _dbContext.TaskSignups.SingleOrDefault(c => c.Id == taskSignupId);

    if (taskSignup != null)
    {
        _dbContext.TaskSignups.Remove(taskSignup);
        return _dbContext.SaveChangesAsync();
    }
    return null;
}
```

Instead of returning `null`, we should be returning `Task.FromResult(0)`

YannickMeeus:
I've picked this up, PR incoming.

tonysurma:
closed in PR #614

PR #614, opened 2016-03-22
Replaced return statements in DataAccess classes to return Task.FromR…

YannickMeeus:
Bit of refactoring of the data access classes to return Task.FromResult(0) instead of just null when the return type is Task, as per #613


mgmccarthy:
@YannickMeeus that was fast!

I'm not too sure if this also applies to methods returning `async Task` or `async Task<T>`. I'll let you know if that's the case and you can pick the rest of that stuff up. 

Also, if you're looking for something to do next, there are still controllers out there with no unit tests. A simple one to pick up would be OrganizationApiController. More involved; ActivityController ad TaskController. VERY involved: ManageController.

Issue #337, opened 2015-11-13
Creating/updating campaign with a postcode: SqlException


ksk100:
I couldn't find an outstanding issue which explicitly covered this, and wasn't sure if it's covered by any related WIP.

But if you try to add an address with a postcode that doesn't exist in `PostalCodeGeo` you violate an FK constraint:

```
SqlException: The INSERT statement conflicted with the FOREIGN KEY constraint "FK_Location_PostalCodeGeo_PostalCodePostalCode". The conflict occurred in database "AllReady", table "dbo.PostalCodeGeo", column 'PostalCode'.
System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
DbUpdateException: An error occurred while updating the entries. See the inner exception for details.
Microsoft.Data.Entity.Update.AffectedCountModificationCommandBatch.Consume(DbDataReader reader)
```

Relates to #304

bcbeatty:
What's happining is If the PostalCode doesn't exist then a FK violation is thrown. Tied to #304

tonysurma:
closed by PRs since creation

PR #671, opened 2016-04-07
Use existing postal code for location or add new one.

auroraocciduusadmin:
When adding or editing an Organization, the address must contain a
postal code that has been added before (ie. in the PostalCodeGeo table).
The changes to OrganizationEditCommandHandler fetch  the added/edited
postal code OR adds the new postal code if not found. The unit tests in
OrganizationEditCommandHandlerTests test that changes to a location's
city/state are NOT copied over to exiting PostalCodeGeo records but ARE
copied to new ones.


auroraocciduusadmin:
This request is related to issues #337, 304, 306, and 249.
I would think that the "handler" changes belong in AllReady.Areas.Admin.Models.LocationExtensions.UpdateModel(), but the extension method doesn't have a database context.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Organizations/OrganizationEditCommandHandler.cs
```diff
@@ -37,6 +37,17 @@ public int Handle(OrganizationEditCommand message)
 
             if (organization.Location != null)
             {
+                if (!string.IsNullOrWhiteSpace(message.Organization.Location.PostalCode))
+                {
+                    PostalCodeGeo postalCode = _context.PostalCodes.SingleOrDefault(pc => pc.PostalCode.Equals(message.Organization.Location.PostalCode, System.StringComparison.InvariantCultureIgnoreCase));
```
mgmccarthy: a query like this exists in `CheckValidpostcodeQueryHandlerAsync`, but the handler checks the validity of a postal code using more than just the PostalCode itself, it checks the City and State as well. So I'm not too sure if it's just safe to add a postal code when it's missing without also first make sure the combination is valid

auroraocciduusadmin:
@mgmccarthy A Location has a city property and a state property. A PostalCodeGeo also has these properties. Is there an authority to let us know if the city, say, of a location has to match the city of it's PostalCodeGeo? I had assumed that they did NOT have to be the same and that was why there was a City for a location and another for it's PostalCodeGeo. Along the same lines, if I change the City of a location without changing it's postal code, it shouldn't necessarily change the City of the PostalCodeGeo. The postal service can tell you the State and City for a postal code, but where I live, you could address the City as Agincourt, Scarborough, or Toronto and it would still get delivered reliably. Also, the primary key of PostalCodeGeo is it's postal code.
The larger issue, however, is that currently you can't add an Organization with an unrecognized postal code and so the OrganizationEditCommandHandler must be able to add a PostalCodeGeo when it comes accross a new one (unless PostalCodeGeo's can be managed somewhere else and the current behaviour of throwing an exception is correct).

mgmccarthy:
@auroraocciduusadmin, thanks for the explanation. I don't know enough about this area of the system, so my comment was more out of curiosity and trying to learn the system. What you say makes sense, I didn't get the difference between a Location and a PostalCodeGeo.

auroraocciduusadmin:
Actually, I don't either...that's why I was asking about an authority. Is
there anything like a written spec for this project?

mgmccarthy:
@auroraocciduusadmin, I'm not too sure if there is an authority for this either. It seems like there is a friction between allowing the creation of a Location with a PostCodeGeo that is not "validated" and creating any of the three entities that have Location:

Activity
Organization
Campaign

If there was an authority (maybe we'll have a fully validated and seeded PostCodeGeo table as part of the system as whole?), then we could chekc with that authority on the form page where we're allowing the creation of these entities.

I pinged a couple others on my comment so we'll see if anyone weighs in.

auroraocciduusadmin:
Thanks for the help. 
The Organization page includes a drop-down for "country" that lists a lot of countries. I can't imagine seeding the PostalCodeGeo table with the postal codes from all countries.
The code in AllReady.Areas.Admin.Models.LocationExtensions.UpdateModel() creates a new PostalCodeGeo whenever it's called and assigns it the location being edited without any checks. So that leads me to believe that it's by design to add PostalCodeGeo's as needed.
In OrganizationEditCommandHandler.Handle I find:
                _context.Update(organization.Location);
...that doesn't seem to cause the newly created PostalCodeGeo attached to the Location to be inserted. It looks like you have to call _context.PostalCodes.Add() somewhere to get the new one into the database. And if you're going to use Add(), it should be unique...thus the fetch.

MisterJames:
I think we need a decision as to whether or not an organization is allowed to add postal codes, versus a system admin.  Looping in @tonysurma and/or @BillWagner again, as I don't have the answer to that one.

tonysurma:
will review later today

mgmccarthy:
@MisterJames @BillWagner @tonysurma, my two cents on this. Maybe the question we should be asking is:
"Is an Organization still useful to the system in any way if it's created without a postal code (validated or non-validated)"

If the Organization is pretty much useless if it's created without a postal code, then let's make that a "first-class citizen" in the system and enforce that on the UI responsible for creating an Organization.

As to whether or not the postal code is valid, can that be a different use case? For example, maybe we let and Organization get created requiring a postal code and there is a separate validating process for that postal code required on a different UI page. Until the postal code is validated, you can't really do anything with the organization. 

Or maybe the validation of the postal code is an async process that is kicked off after the Org is created, and there is an email sent to the admin of the Org if there is a problem with the validity of the postal code? That way, they need to fix it before the Org becomes usable.

I don't know the "actor" that would be responsible for validating an Org's postal code, if it this is even a route that makes sense. I'm just offering some ideas here. 

Most likely, I don't understand the business side/use cases/requirements of the project enough, so please forgive me if these seem like silly suggestions :wink:

tonysurma:
So there is a high level spec but with organic growth in an OSS project form a big jumpstart you get things that implement assumptions that show their head later like this.

Thanks @auroraocciduusadmin for jumping into this.

I would like to suggest/ask that fix be to remove the PostalCodeGeo table, thus removing foreign keys and thus remove the issue hitting us that causes all the questions in all the associated issues.

You should totally be able to submit an org with whatever postal code you enter (just like we don't validation you spelled 'Chicgao' right) and that org is valid and will work.  As searching happens on location/other attributes you won't show up or show up weird and those can be fixed in the search parts of the system or by the user editing the database because they don't see themselves in search results.

This works for the system 'business' requirements.  Let me know if it causes any big concerns from an implementation point of view as you've seen it @auroraocciduusadmin 

Thanks again

auroraocciduusadmin:
I just assumed that the classes were written with a special use for a postal code object in mind. I don't know what the business requirements are and the change I proposed was just a way for a user to be able to add an organization (or campaign or activity) which can't be done now. So basically, I don't have any special concerns aside from the ability to add an organization.

tonysurma:
Great, so thinking we can accept this then as it removes the issue of the need to be prepopulated and then later we can eval if the whole geocodepostal thing is needed.  @MisterJames can you review If you haven't already but from the thread it looks like this is good to go.  Thx @auroraocciduusadmin

MisterJames:
Yeup, we're good to go here. Thanks @auroraocciduusadmin!

PR #672, opened 2016-04-07
Refactoring IEmailSender out of SiteAdminController

mgmccarthy:
Fixes #660

- using mediator to send mail from this controller by ways of explicitly named messages
- refactored any _dataAccess.GetUser(userId) calls into private GetUser method in SiteAdminController to ready for refactoring to mediator
- update QueryHandler to async as well as removing an unneeded dependency from constructor
- updated all pertinent unit tests/stubs to reflect the refactoring changes

PR #679, opened 2016-04-12
Add tests to register post

darrencauthon:
Adds unit tests to to the ```Register``` actions (both the GET and POST).

**Note** I am not unit testing with the same pattern as used previously... but I'd strongly recommend **NOT** having a single test fixture for the entire controller.  The actions are just too different, so trying to mush them into one common setup makes testing harder.  So instead, I have a test fixture per action.

Also note:  I added a test dependency on an automocking tool:  AutoMoq.  I'd also recommend using it for tests like these, as it drops the number of lines of test code considerably.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
@@ -163,10 +163,28 @@ namespace AllReady.UnitTest.Controllers
             [Fact]
             public void It_should_send_a_confirmation_email()
             {
+                //var token = await _userManager.GenerateEmailConfirmationTokenAsync(user);
+                //var callbackUrl = Url.Action(new UrlActionContext
+                //{
+                //    Action = nameof(ConfirmEmail),
+                //    Controller = "Account",
+                //    Values = new { userId = user.Id, token = token },
+                //    Protocol = HttpContext.Request.Scheme
+                //});
+
+                var callbackUrl = Guid.NewGuid().ToString();
+                urlHelperMock.Setup(x => x.Action(It.Is<UrlActionContext>(
+                    y => y.Action == "ConfirmEmail" &&
+                         y.Controller == "Account" &&
+                         y.Protocol == requestScheme)))
+                         .Returns(callbackUrl);
+
                 Subject.Register(registerViewModel).Wait();
+
                 Mocked<IEmailSender>()
-                    .Verify(x => x.SendEmailAsync(It.IsAny<string>(),
-                        It.IsAny<string>(), It.IsAny<string>()));
+                    .Verify(x => x.SendEmailAsync(registerViewModel.Email,
+                        "Confirm your allReady account",
+                        $"Please confirm your allReady account by clicking this link: <a href=\"{callbackUrl}\">link</a>"));
             }
 
             [Fact]
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
@@ -15,6 +15,7 @@ using Microsoft.Extensions.OptionsModel;
 using Moq;
 using Xunit;
 using MediatR;
+using Microsoft.AspNet.Mvc.ViewFeatures;
 using Shouldly;
 using ClaimTypes = AllReady.Security.ClaimTypes;
 
@@ -154,13 +155,29 @@ namespace AllReady.UnitTest.Controllers
             public void It_should_stay_on_the_registration_page_if_the_user_creation_fails()
             {
                 userManagerMock.Setup(x => x.CreateAsync(It.IsAny<ApplicationUser>(), registerViewModel.Password))
-                    .Returns(Task.FromResult(IdentityResult.Failed(new IdentityError())));
+                    .Returns(Task.FromResult(IdentityResult.Failed(new IdentityError() { Description = Guid.NewGuid().ToString()})));
 
                 var result = Subject.Register(registerViewModel);
 
                 result.Result.ShouldBeOfType(typeof(ViewResult));
 
                 (result.Result as ViewResult).ViewData.Model.ShouldBeSameAs(registerViewModel);
+            }
+
+            [Fact]
+            public void It_should_return_The_account_creation_error_if_the_user_creation_fails()
+            {
+                var description = Guid.NewGuid().ToString();
+                userManagerMock.Setup(x => x.CreateAsync(It.IsAny<ApplicationUser>(), registerViewModel.Password))
+                    .Returns(Task.FromResult(IdentityResult.Failed(new IdentityError() { Description = description})));
+
+                var result = Subject.Register(registerViewModel);
+
+                result.Result.ShouldBeOfType(typeof(ViewResult));
+
+                var viewData = (result.Result as ViewResult).ViewData;
+                viewData.ModelState[""].Errors.Count.ShouldBe(1);
+                viewData.ModelState[""].Errors[0].ErrorMessage.ShouldBe(description);
             }
 
             [Fact]
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
@@ -109,12 +109,6 @@ namespace AllReady.UnitTest.Controllers
             }
 
             [Fact]
-            public void New_test()
-            {
-                Subject.Register(registerViewModel).Wait();
-            }
-
-            [Fact]
             public void It_should_redirect_to_home_page_after_a_success_user_creation()
             {
                 var result = Subject.Register(registerViewModel);
@@ -137,6 +131,32 @@ namespace AllReady.UnitTest.Controllers
                 result.Result.ShouldBeOfType(typeof(ViewResult));
 
                 (result.Result as ViewResult).ViewData.Model.ShouldBeSameAs(registerViewModel);
+            }
+
+            [Fact]
+            public void It_should_stay_on_the_registration_page_if_the_input_is_bad()
+            {
+                var key = Guid.NewGuid().ToString();
+                var errorMessage = Guid.NewGuid().ToString();
+                Subject.ViewData.ModelState.AddModelError(key, errorMessage);
+
+                var result = Subject.Register(registerViewModel);
+
+                result.Result.ShouldBeOfType(typeof(ViewResult));
+                (result.Result as ViewResult).ViewData.Model.ShouldBeSameAs(registerViewModel);
+            }
+
+            [Fact]
+            public void It_should_not_attempt_to_create_a_new_user_when_the_input_is_bad()
+            {
+                var key = Guid.NewGuid().ToString();
+                var errorMessage = Guid.NewGuid().ToString();
+                Subject.ViewData.ModelState.AddModelError(key, errorMessage);
+
+                userManagerMock.Setup(x=>x.CreateAsync(It.IsAny<ApplicationUser>(), It.IsAny<string>()))
+                    .Throws(new Exception("Should not have been called!"));
+
+                Subject.Register(registerViewModel).Wait();
             }
         }
 
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using AllReady.Controllers;
 using AllReady.Models;
 using AllReady.Services;
+using AllReady.UnitTest.Extensions;
 using AutoMoq;
 using AutoMoq.Helpers;
 using Microsoft.AspNet.Http;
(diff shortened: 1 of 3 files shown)


mgmccarthy:
@darrencauthon. The email sending is not testable in what file or file(s)?

darrencauthon:
@mgmccarthy It's in ```AccountController```, the ```Register``` method with the ```HttpPost``` attribute.  It's referencing ```HttpContext```, which is difficult to test.

I saw that there was some type of ```IHttpContextAccessor``` that might be a way out.  I wanted to see how far I could take the tests without being forced to alter the code (even though I did a refactor step afterwards :smile:).  

I could find some way around the ```HttpContext``` issue, but I think the overall problem to fix is the building and firing of the email in the method.  It would probably be better to extract a new class for that task.

Plus, I tried something new last night... I recorded my testing session.  I've been talking to some friends about testing, and testing existing code.  Here's the video of me talking through this pull request: 

https://www.youtube.com/watch?v=Qsl07CxnRFU

mgmccarthy:
@darrencauthon, if the part of the `Register` method with the `HttpPost` attribute that references `HttpContext` is:
```
var callbackUrl = Url.Action(new UrlActionContext { Action = nameof(ConfirmEmail), Controller = "Account", Values = new { userId = user.Id, token = token }, Protocol = HttpContext.Request.Scheme });
```

There is a way to unit test this. Check out AdminControllerTests, lines 163-170:
```
var urlHelper = new Mock<IUrlHelper>();
urlHelper.Setup(x => x.Action(It.IsAny<UrlActionContext>())).Returns(callbackUrl);
...
sut.Url = urlHelper.Object;
```

in this case, the `sut` variable is really a controller:
```
var sut = new AdminController(userManager.Object, null, emailSender.Object, null, null, generalSettings.Object);
```

if you look at the `Controller` base class, you'll see it has a writable `Url` property that returns a `IUrlHelper`:
```
public IUrlHelper Url
{
    get
    {
        if (this._url == null)
        {
            IServiceProvider resolver = this.Resolver;
            this._url = resolver != null ? ServiceProviderExtensions.GetRequiredService<IUrlHelper>(resolver) : (IUrlHelper) null;
        }
        return this._url;
    }
    set
    {
        if (value == null)
            throw new ArgumentNullException("value");
        this._url = value;
    }
}
```

and that's what makes `Url.Action` mockable.

IMO, from what I've read on how to use `IHttpContextAccessor`, it looks like it's meant for injecting an HttpContext into a class that is a non-controller class. Although, I don't think that's a good idea per-se, as you're now tying a non-web class to a piece of a web framework, I can see where it would be useful in factory code or "service" classes.

darrencauthon:
@mgmccarthy Oh yeah, I'm sorry -- Yeah, the ```Url``` bit is testable.  The ```HttpContext``` one is not, and that's the barrier.  I just didn't touch ```Url``` because all of that code was tied to the building of the email.

But looking at it now, my second look... I bet it's not necessary.  I don't know what comes out of ```HttpContext.Request.Scheme``` but if unless it's something that can only be produced by internals, I bet it can be hardcoded to a value that will work.

I understand wanting to push URLs through the URL builder, but I don't find route changes to be very frequent.  It might be easier to just hardcode the URL building.  If the route ever changes, it would be up to the person changing the route to verify that the feature continues to work after the change.  And when it doesn't, they'll just change the hardcoded URL function.

darrencauthon:
I set out to test the methods without changing their behavior.  Technically this means that I'd want to test this code as-is... but as soon as it is tested, I think the email send should be moved out to its own class... even if it's a class defined without the account controller.  Building the details of an email require a lot of specialized email tests within tests for many other things.

mgmccarthy:
@darrencauthon, the sending of the email is in it's own class via the injected _emailSender dependency. What the controller owns is the building of the message to go on the email, and since this involves accessing some HttpContext specific items, that code is ok to stay in the controller.

That being said,I did not fully read your message earlier, so I apologize... it SHOULD have been pretty apparent to me that the direct access to `HttpContext` you were referring to is `HttpContext.Request.Scheme`

To test `HttpContext.Request.Scheme`, there is a `SetFakeHttpRequestSchemeTo` extension method in `ControllerTestHelpers` that will return the `requestScheme` string you pass to the method.

So, to test `Url.Action` that also calls into `HttpContext.Request.Scheme` you're setup could look like this:
```
var urlHelper = new Mock<IUrlHelper>();
var controller= new Controller();
sut.SetFakeHttpRequestSchemeTo(requestScheme);
sut.Url = urlHelper.Object;
...
urlHelper.Verify(...
```

If you don't want to use the extension method, you can work directly with creating a mock of `HttpContext` and `HttpRequst` and then setting up the `HttpRequeset.Scheme` to return a certain value. Assign all of that to `Controller.ActionContext.HttpContext` and you should be set.

Also, I do agree that the usage of `HttpContext.Request.Scheme` seems like overkill here, especially since like you said, that value will rarely, if every, change. My goal in climbing on board this project was to minimize my impact to the code that needed unit tests written. That doesn't mean I agree with how all the existing code is written in the controllers, BUT, once there are good amount of unit tests around them, refactoring will be much easier.

mgmccarthy:
One more thing, I'm in middle of refactoring`IEmailSender` out of the controllers altogether. I have my first PR of many PR's to come waiting to be merged. It's PR #672. The overall goal here is to move to CRQS for controllers across the board. So removing things like `IEmailSender` and `ISmsSender` can reduce controller dependencies and move more code into smaller, discreetly testable classes based on CRQS via MediatR. It also makes commands sent in the system explicit and hides the implementation of those commands from the controller that would be sending them.

Whether or not we'll continue to extend that refactoring to include things like `UserManager<ApplicationUser>` and `SignInManager<ApplicationUser>` remains to be seen. I think that would required a lot of work, and something that the project owners don't necessarily see as very important right now.

darrencauthon:
Do you mean CQRS?

Get Outlook for iOS




On Tue, Apr 12, 2016 at 2:17 PM -0700, "Michael McCarthy" <notifications@github.com> wrote:












One more thing, I'm in middle of refactoringIEmailSender out of the controllers altogether. I have my first PR of many PR's to come waiting to be merged. It's PR #672. The overall goal here is to move to CRQS for controllers across the board. So removing things like IEmailSender and ISmsSender can reduce controller dependencies and move more code into smaller, discreetly testable classes based on CRQS via MediatR. It also makes commands sent in the system explicit and hides the implementation of those commands from the controller that would be sending them.



Whether or not we'll continue to extend that refactoring to include things like UserManager<ApplicationUser> and SignInManager<ApplicationUser> remains to be seen. I think that would required a lot of work, and something that the project owners don't necessarily see as very important right now.

darrencauthon:
@mgmccarthy Thank you for your help in finding how to fake out that scheme stuff.  I think this pull request is complete.

tonysurma:
@mgmccarthy can you review these tests? Thx

MisterJames:
Tests are passing okay here on my side. @mgmccarthy you've helped this one along, any additional comments?

mgmccarthy:
@tonysurma @MisterJames, I might not be able to get my eyes on it for a couple days. If you want to merge it now to keep things going, I'm ok with it. We can always refactor after the PR is merged.

As far as how to do code reviews on GitHub, I could use a little help, as I've never used this process for code reviews before. @MisterJames @BillWagner what do you guys do? Pull the branch down locally, look at the code, run the unit tests, etc...??? or do you just eyeball it on GitHub and accept the PR? Basically, we know if any of the unit tests failed b/c the build server will fail the build and you'll see that in the PR.

I might also need a bit of help on the git command line for how to pull down a specific branch for this repository.

Thanks!

darrencauthon:
@mgmccarthy Here are the git commands you'll want to use to pull this locally:

```
git remote add darrencauthon git@github.com:darrencauthon/allReady.git
git remote update
git checkout -b add_tests_to_register_post darrencauthon/add_tests_to_register_post
```

That will hook you to my repo and checkout my branch.  Then you can review the code locally.

But I've done a TON of code reviews on Github.  I don't think you need to go to this trouble.  Just look over the code and do a "gut check."  Is this wrong?  Will this hurt?  Is the CI server reporting that the tests are passing?  

If I were you, I'd take the PR.  But I wrote the PR, of course.  I haven't changed the production code (I took out my refactors) except for resharper spacing changes.  I added tests that fully vet these two methods, and... in my opinion, I am using a unit testing strategy that could help you or anybody else to test the rest of the app.  I don't think the ```taskFromZeroResult``` approach is going to work. :smile:  I added a new testing dependency in AutoMoq, but... AutoMoq was written specifically to target these sorts of applications.  I've written thousands of unit tests like this... it works!

mgmccarthy:
@darrencauthon, the `TaskFromResultZero`:
```
//delete this line when all unit tests using it have been completed
private static readonly Task<int> TaskFromResultZero = Task.FromResult(0);
```
was added so I could stub out async unit test methods for other developers to write. Without it, the solution won't compile. It's only a placeholder and should be removed once all unit tests in a test class have been completed.

I'll take a look at how you handled it differently in the PR.

darrencauthon:
@mgmccarthy I didn't touch the ```TaskFromResultZero``` stuff... I just ignored them and wrote unit tests based on the code that existed.

What I mean is... I think the ```TaskFromResultZero``` actually **increases the difficulty** in testing this code.  We already know what needs to be tested -- the code that exists.  I think the only person qualified to determine which test methods need to exist is the programmer writing the tests.  Any other input on the tests just gets in the way.

Today, when I run all of the unit tests in the project, I got a HUGE block of tests that yet to be written.  Am I to piece through this huge list and find the tests that I'm writing?  It's much faster and easier to just write a new test fixture and start with new tests.

mgmccarthy:
Wanting input from some of the project owners: @tonysurma @MisterJames @BillWagner.

darrencauthon has a point here. Since I'm the one who took up the task of stubbing out the test names that need to be written for classes with no tests or incomplete tasks, maybe just creating an empty test class (for non-existent unit tests) and then creating an issue for it is a better way to go?

Every contributor has a different way of working, and since we don't really have "code standards" for this project and it is open source, maybe it's best to let each contributor take a crack at it? darrencauthon likes to use AutoMoq. Another contributor introduced Shouldly. People like to use different ways to setup their test. Some people like to CamelCase their test names, some people like to use _

On the flip side, I've put a good amount of effort into the test stubs to make sure we're not missing anything in tests that still need to be written so a contributor does not have to. They can change the name of the tests if they'd like, but the job of tracing through the all the execution path of code that needs tests has already been taken care of and is reflected in the names of the test method stubs.

Thoughts?

@ darrencauthon I'll try to get to this code review in a couple of days, but might be tough as my regular job is starting to get busy again.

darrencauthon:
@mgmccarthy 

>  I've put a good amount of effort into the test stubs to make sure we're not missing anything in tests that still need to be written so a contributor does not have to ... the job of tracing through the all the execution path of code that needs tests has already been taken care of

Just to be clear -- I don't mean it as any slight on you. :smile: If humans could read code and interpret code that well, we wouldn't need to unit test our apps.  We could just look at the code and know.  I'm sure you looked through the execution paths and put care into those test names, and they are probably mostly right... but I'm just saying... no good unit test writer would use them.  Even though

I've had a ton of OSS contributions.  This one wasn't a big deal, and I'll be fine if you rejected it.  But my unsolicited advice... if someone makes the application better, in any way, it's the OSS way to accept it.  Especially when the improvements don't actually change the behavior, making them very safe.  Github culture is not enterprise (yet)... just go with the flow, man! :smile_cat: 

up-chong

mgmccarthy:
@darrencauthon, not taken as a slight in one bit. As a non-project owner, I want some more input into what I'm doing b/c if it's making it harder to write unit tests for contributors, that's not my intent.

mgmccarthy:
@darrencauthon some things to change:

1. the `Register()` GET action method has two attributes on it. Can you please add unit tests for these attributes? There are existing unit tests in the system that do this if you're looking for an example.
2. do the same thing for `Register()` POST attributes.
3. can you please remove any unused `using`s from the top of the `AccountControllerTests` file? We just spent some time going through the whole solution and doing that, and it cleaned stuff up nicely.
4. can you please remove the old unit tests stubs for the tests you wrote? Basically, any unit test stub beginning with `RegisterGet` or `RegisterPost`

@MisterJames, after darrencauthon commits these fixes, merge the PR :shipit:

Thanks for the contribution!

tonysurma:
Hello @darrencauthon and @mgmccarthy - first off let me say thanks to both for the contributions.

Below will be long but I want to communicate the nuance.  In short summary: good with AutoMoq add, good with new tests 'replacing' the placeholders in unique style as offered, want to keep our placeholders for jumpin new users, and want to be careful about adding new tools/dependencies/frameworks so the cognitive load stays small for contributions.

Reading through this and thinking of the project direction, we would say the following below is our approach to best move forward to support the project and engage the broadest audience who can contribute.

Each time we add a dependency or framework we apply the principles described in this [comment discussion](https://github.com/HTBox/allReady/issues/59#issuecomment-187432429). Given, that and since AutoMoq is an extension to Moq already in place I think it makes sense to add & thanks for that @darrencauthon 

As for the approach to unit tests, we want to have an approach that opens up adding tests to the widest audience and agree that moving the project forward is the goal but don't want to make changes that hinder the ability for others to move forward either.  With that, what @mgmccarthy mapped out a set of tests for others that we are labeling 'jumpin' is good for those new to the project or testing but are also totally open to having others enter tests in their style as long as they are discoverable (which all tests will naturally be.)  Also in that case it would be helpful to remove any duplicate test placeholders their tests 'implement' (in this case I think @mgmccarthy identified them above)

Lastly I think the discussion about test mapping to classes/action is great feedback and as we add more placeholders we will work with @mgmccarthy and others to continue to advance the approach.

@MisterJames - once the last few comments are addressed/mitigated lets merge it in.

Altogether, all these contributions are a great addition to a big code base without a lot of test coverage so thanks!

darrencauthon:
Programmer: "Free horses! Get your free gift horses right here!"

Project owner: "Your gift horse looks fine... but I'll need you to give me a few weeks to review it. Can you also polish it's teeth, and... write four more tests to cover attributes I put on the method? Then I would be happy to accept your gift horse."

Issue #705, opened 2016-04-16
6.1.2.1	Entities have new names from user experience discussions – P2


tonysurma:
Campaigns are Campaigns
***Activities are Events***
Tasks are Tasks
***Deployments are Itineraries***

These changes need to be reflected in the UI and secondarily in the system/database for consistency & code clarity (and yes we know we are talking entities named Events and Tasks in C# :smile_cat:)

stevejgordon:
@tonysurma @BillWagner Please assign to me, I'll tackle this next.

stevejgordon:
My plan of attack! I figure I may as well get everything done for this to avoid any confusion for people. Not sure if it's easier as one big PR or smaller chunks? @MisterJames - Your thoughts? I'm happy to do whatever makes reviewing / testing easier.

- [ ] Rename activities in UI elements
- [ ] Rename activities in js elements
- [ ] Rename activities in code
- [ ] Rename activities in database
- [ ] Rename activities in folder structure and files

- [ ] Rename deployments in UI elements
- [ ] Rename deployments in js elements
- [ ] Rename deployments in code
- [ ] Rename deployments in database
- [ ] Rename deployments in folder structure and files

BillWagner:
@stevejgordon Great plan.  My experience is that several smaller PRs is much easier. If @MisterJames feels different, he can override though.

tonysurma:
@MisterJames and I have discussed this several times and definitely smaller 'bite sized' are better.  Thanks

stevejgordon:
@tonysurma This can hopefully be closed now since all 5 PR's have been merged.

PR #718, opened 2016-04-22
Renaming deployments to itineraries

stevejgordon:
Please merge this after PR #717 

Renaming the only two occurrences of the word Deployment to Itinerary in the project.

Relates to issue #705


tonysurma:
merged the other and was going to merge this.  Is seems this PR has the same commit as the other for the UI changes?  Weird is it going to double commit that? @BillWagner you seen anything like this (glancing for a moment during a break so perhaps I would understand it if I stared at it longer...)

stevejgordon:
@tonysurma I think I based this one off of my previous branch to avoid conflicts. I don't think it shouldn't double commit, however I will try rebasing it from the current master branch.

stevejgordon:
@tonysurma That seems to have tidied it up.

PR #722, opened 2016-04-23
Renaming activities to events in all cs code

stevejgordon:
This includes class names, properties, fields and method names. This includes renaming and updating the tests.

This change also includes database model updates and a migration to update dev databases.

**IMPORTANT! This migration involves some dropped tables and is destructive to existing data.**

View folders and names have also been updated to ensure the site functions based on the view location conventions. Other cs files and folders have not been changed in this commit and will form their own commit and PR later to try and limit the complexity of reviewing this PR.

This relates to issue #705


tonysurma:
This is okay - dropping of tables etc. as we are not in active use between now and when the app will be used for the June 11th event.  Thanks for checking, it is key we verify as you did.

tonysurma:
once we get a PR review we can merge and pull in as the table dropping is okay cc @MisterJames

tonysurma:
actually I scanned the 159 files ! changed and thanks for the thorough change set.  Given that review and passing build I am going to accept

PR #723, opened 2016-04-23
Final renaming of activities to events

stevejgordon:
Please merge after PR #722 (may require me to rebase this one before accepting it)

Includes:
File and folder renaming
Final source file updates
SQL file updates
CSS class updates

This final PR fixes issue #705

PR #725, opened 2016-04-25
refactoring IEmailSender and ISmsSender out of AdminController

mgmccarthy:
- removed IEmailSender and ISmsSender from AdminController and replaced with Mediator
- added messsage/handlers for each email/sms sent from controller that is now being sent by Mediator
- updated unit tests for AdminControllerTests
- **TODO**: add unit tests for newly added handlers (this will be a separate PR)

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AdminControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AdminControllerTests.cs
@@ -3,9 +3,10 @@ using System.Linq;
 using System.Threading.Tasks;
 using AllReady.Controllers;
 using AllReady.Extensions;
+using AllReady.Features.Admin;
 using AllReady.Models;
-using AllReady.Services;
 using AllReady.UnitTest.Extensions;
+using MediatR;
 using Microsoft.AspNet.Authorization;
 using Microsoft.AspNet.Identity;
 using Microsoft.AspNet.Mvc;
@@ -83,7 +84,7 @@ namespace AllReady.UnitTest.Controllers
             var userManager = CreateUserManagerMock();
             userManager.Setup(x => x.CreateAsync(It.IsAny<ApplicationUser>(), It.IsAny<string>())).Returns(() => Task.FromResult(IdentityResult.Failed()));
 
-            var sut = new AdminController(userManager.Object, null, null, null, null, generalSettings.Object);
+            var sut = new AdminController(userManager.Object, null, null, null, generalSettings.Object);
 
             await sut.Register(model);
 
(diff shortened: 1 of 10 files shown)


tonysurma:
my review looks good, @MisterJames can you give this a quick once over?

mgmccarthy:
@tonysurma @MisterJames I'd work on merging #723 (Final renaming of activities to events) before merging this. I can easily rebase and fix and naming conflicts, then push. I'll have a separate PR for the rest of the unit tests associated to work done in this PR.

tonysurma:
I merged that PR @mgmccarthy

mgmccarthy:
@tonysurma, looks like we're good on the local rebase off of the renaming's. My guess is the renaming's did not touch any of this code in the PR. Should be good to go whenever a code review can be done.

AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AdminControllerTests.cs
```diff
@@ -107,7 +108,8 @@ public void RegisterHasAllowAnonymousAttribute()
             var userManager = CreateUserManagerMock();
             userManager.Setup(x => x.CreateAsync(It.IsAny<ApplicationUser>(), It.IsAny<string>())).Returns(() => Task.FromResult(IdentityResult.Success));
 
-            var sut = new AdminController(userManager.Object, null, Mock.Of<IEmailSender>(), null, null, generalSettings.Object);
+            //var sut = new AdminController(userManager.Object, null, Mock.Of<IEmailSender>(), null, null, generalSettings.Object);
```
MisterJames: We can probably just nuke this (instead of commenting it out).

MisterJames:
Okay, good PR. In reviewing this I'm thinking I'm going to merge, but we're continuing to work against two pieces of tech debt here: Admin-based account confirmation and messaging propagating it's way through the system (rather than being abstracted out).

I'm going to think on a couple of things:
 1. Unifying the signup/confirmation process (we can always elevate to admin with claims through some other process)
 1. Getting a templating process in place for emails/SMS such that templates and related language are defined external to the source code (but preferably under source control)
 1. Ensuring users can manage notification preferences (SMS, email, digest, DND)
 1. Pushing the code base to use a unified messaging service/command in such a way that the caller doesn't need to know how the user prefers to be contacted.

AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AdminControllerTests.cs
mgmccarthy: i can take care of this in another PR. Sorry, I forgot to uncomment it.

mgmccarthy:
:+1: for all of those, especially the unified message service/command that figures out how the user want to be notified instead of the caller need to know which type of notification to use.

That can be pulled from the users managing their notification process in another page.

If you need help or want to delegate, let me know.

Thanks

Issue #719, opened 2016-04-22
Closed tasks show the volunteer button


stevejgordon:
Presumably this button should be hidden if the task is closed.

Clicking the button shows the volunteer modal and allows the user to volunteer. This should also be checked and blocked in the code.

![image](https://cloud.githubusercontent.com/assets/3669103/14755699/13159660-08dc-11e6-8b8e-a4f62a40f8cc.png)

stevejgordon:
@tonysurma You can assign to me if you want and I'll take a proper look after my renaming work gets done.

stevejgordon:
Update - Nearly done with this one. Got a few tests to right and a little refactoring.

stevejgordon:
@tony My PR message was not quite right to auto close this!

PR #726, opened 2016-04-25
Fixing error where closed tasks can be volunteered for.

stevejgordon:
Fixes issue #719

NOTE: Must merge after PR #722 and PR #723 as it is based on the renaming exercise. I will need to rebase this if/when those are accepted.

Primarily I have hidden the volunteer buttons for tasks which are closed. In addition I have hardened the API and handler to fail if a closed task id is sent.

Changes made:
Added bool isClosed calculated property to the task model (not mapped)
Passed the isClosed value to the TaskViewModel
Used the isClosed property to hide the volunteer buttons for closed tasks
Updated TaskSignupHandler to check for closed tasks and issue error in result
Also added checks for Event and Task not found scenarios
Updated register method for TaskAPI to handle and output correct JSON for the various error responses
Added tests for all changes with expected behaviours

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Features/Tasks/TaskSignupResult.cs
+++ b/AllReadyApp/Web-App/AllReady/Features/Tasks/TaskSignupResult.cs
@@ -4,6 +4,12 @@ namespace AllReady.Features.Tasks
 {
     public class TaskSignupResult
     {
+        // TODO - Move to enum
+        public const string FAILURE_CLOSEDTASK = "failure-taskclosed";
+        public const string SUCCESS = "success";
+        public const string FAILURE_EVENTNOTFOUND = "failure-eventnotfound";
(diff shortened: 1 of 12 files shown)


MisterJames:
@stevejgordon great PR and I love the detail on the changes. Thanks for this. Reviewing now.

AllReadyApp/Web-App/AllReady/Features/Tasks/TaskSignupResult.cs
```diff
@@ -4,6 +4,12 @@ namespace AllReady.Features.Tasks
 {
     public class TaskSignupResult
     {
+        // TODO - Move to enum
```
MisterJames: ....actually, rather than thinking about enums, we're going to have to start to think about resource files.

PR #727, opened 2016-04-27
Posting campaign updates the database

auroraocciduusadmin:
Added a test to verify that when a new campaign is created and posted it
will be saved in the database. Even though the unit tests all pass when I run them, I can't actually use the web app to add a campaign. The test included with this commit will fail.


mgmccarthy:
@auroraocciduusadmin, why do you have to inherit from `InMemoryContextTest` for `CampaignAdminControllerTests`? All dependencies are 100% mockable. Mediator is used everywhere.

In this code:
```
mockMediator.Setup(x => x.SendAsync(It.IsAny<EditCampaignCommand>()))
    .Returns(() => {
        IAsyncRequestHandler<EditCampaignCommand, int> handler = new EditCampaignCommandHandler(Context);
        return handler.Handle((EditCampaignCommand)command);
});
```

you're injecting a Context to the command handler, but you can easily return whatever you want from the mock (in this case, the handler returns an int) and not new up the command handler at all...  mock it instead.

Could these validators be the issue?:
```
CampaignSummaryModelValidator
LocationEditModelValidator
```

looks like they are new'ed up in the controller, so we have to provide mocks of their internals via mediator. A better approach is to move the validators into injectable dependencies into the controller(s) that need them, and mock the outcome of the validator. 

Then the validators become individually testable and the mock setup work lightens for the controllers using them. But that is another conversation ;)

auroraocciduusadmin:
I inherited from InMemoryContextTest so I could have an in-memory Context to work with. The purpose was to test that one could pass a plausible CampaignSummaryModel to the controller and have a campaign added to the database. By checking the context at the end of the test I could determine if it indeed contains the added Campaign. 
That said, I can see now that the test as posted would give a "false positive" because, even if the validation could be passed (it can't) the context would contain the Campaign, but it's location's postal code would be in an "unattached" state. In the database, that would cause a foreign key violation but the test would still pass.
I see that there are already tests in EditCampaign and CampaignSummaryModelValidatorShould and perhaps that is where I should have done the testing?
Finally, I'm afraid I missed your point about injecting the Context: did you mean that I simply should not have used a context for testing? or I shouldn't have inherited from InMemoryContextTest? or something else?

mgmccarthy:
@auroraocciduusadmin, After reading through validators calling validators, it looks like there is postal code work being down in the `LocationEditModelValidator`? This is invoked from the `CampaignSummaryModelValidator`, and this validator is invoked from `AllReady.Areas.Admin.Controllers.CampainController`. Lots going on here.

Seeing how it looks like PostalCode work is being done in `LocationEditModelValidator`, maybe you can start by checking the unit tests there make sense? Then move on to the `CampaignSummaryModelValidator` tests?

That being said, it looks like the main problem, as you said in the first comment associated with this PR is "Even though the unit tests all pass when I run them, I can't actually use the web app to add a campaign." We should fix that actual problem first, and then worry about the unit tests.

As far as inheriting from `InMemoryContextTest` for `CampaignAdminControllerTests`, my point here was that we should not be testing anything with the database from CampaignControllerTests since CampaignController uses mediator to abstract the writing of a Campaign to the database.

Looking through the code, there is an `Edit` post action in `CampainController` that sends an `EditCampaignCommand`. Looking at the `EditCampaignCommandHandler`, this is the class that actually does the work of inserting the Campaign into the database, so this classes' unit tests (if tests exist) should be using an in memory context (you can inherit from `InMemoryContextTest`) and have the tests that the Campaign is being to the context directly.

On a side note, the validators that have been added to this project really should be abstracted and passed into the controller's that need them via DI. That way, the validators become individually unit testable, and the controllers can mock the validator in order to test it's execution flow. But a refactoring like this can be deferred to later though, b/c I believe there are P1 items and bugs to fix right now.

mgmccarthy:
@auroraocciduusadmin, according to the conversation in Issue #671, I THINK the decision was to drop the `PostCodeGeo` table altogether??? But i still see it after wiping my local db and restarting the app locally.

Looks like `LocationEditModelValidator` sends `CheckValidPostcodeQueryHandlerAsync` and it's handler is still querying `context.PostalCodes` and returning false if the postal code is not found in the `PostalCodeGeo` table.

@MisterJames @tonysurma, what's the decision on not allowing an Organization to be created when there is not matching postal code in the PostalCodeGeo table?

tonysurma:
specific to geotable - short version - for this PR leave/do whatever is in there but in short we don't need that link at all and will move on from it in #671 or another set of updates

mgmccarthy:
@tonysurma, so the decision is to not do the postal code validation and allow the Campaign to be created without a match in the PostalCodeGeo table?

tonysurma:
Yes

mgmccarthy:
@tonysurma @auroraocciduusadmin @MisterJames @BillWagner  

I just fiddled with the code for a bit, and it seems just removing this check is not as easy as it first sounded. 

Because the `Location` entity (which contains `PostalCodeGeo`) is a property of `Campaign`, `Organization` and `Event`, when going to create a new one of any of those entities, `LocationExtensions.UpdateModel` is invoked, which has this code:
```
if (!string.IsNullOrWhiteSpace(locationEditModel.PostalCode))
{
    location.PostalCode = new PostalCodeGeo { PostalCode = locationEditModel.PostalCode, City = locationEditModel.City, State = locationEditModel.State };
}
```
By commenting out this code, I can prevent the foreign key exception that is thrown when creating a new `Campaign`, but the `LocationExtensions.UpdateModel`is also called from `EditEventCommandHandler` and `OrganizationEditCommandHandler`, so this check would also be bypassed for creating an `Organization` or an `Event` (which is not what we want).

Interim fixes could be to copy the code we need out of the `UpdateModel` method to `EditCampaignCommandHandler` so we can have control over the `PostalCodeGeo` creation and leave both `Organization` and `Event` using the `UpdateModel`method to enforce the postal code stuff. 

I'm not a big fan of this approach, but it's the first step in breaking out postal code stuff from the `Location` entity, b/c it seems that `Campaign` has different postal code rules it needs to follow than `Event` or `Organization`... or maybe all of them don't need this check? 

Either way, creating/editing a `Campaign` can no longer use `UpdateModel` if we want to preserve the check for `Organization` and `Event` and bypass it for `Campaign`

auroraocciduusadmin:
I agree that "commenting out" the lines you mentioned will prevent the foreign key exception but at the cost of leaving the location without a postal code. Otherwise I think LocationExtensions.UpdateModel is fine the way it is. For the OrganizationEditCommandHandler, the postal code is added to the location after calling UpdateModel...overriding the part you mentioned. Could we do the same thing for Campaign and Activity?

MisterJames:
@auroraocciduusadmin I think we actually drop the FK there and add a different kind of soft-validation, such as a business rule that doesn't allow the organization to go live with their campaigns (or something to that effect) without filling it in. 

Thoughts @tonysurma?

mgmccarthy:
@auroraocciduusadmin, good find on the difference in code in `OrganizationEditCommandHandler`. I can see where the `Location` work is being done in the handler so we don't have this exception being thrown when there is no matching code in the `PostalCodeGeo` table when creating any of these three entities. This would fix the problem.

@MisterJames, I like the idea of dropping the FK more.  `Location` is being treated like a value object, but it really doesn't need to be. and it's write validation is caught up in shared code where is shouldn't be.

Worst comes to worst, the FK association can be dropped and the client code can call/write to `PostalCodeGeo` in a separate LINQ call within the same context instead of saddling the three entities with the same rules for `Location.PostalCodeGeo`.

I'm not really that familiar with working with EF 7 code first where there is one database and all the relationships are taken care in the db and read using `Include`.  I've always used SOA and boundaries, so the relationships are enforced in the code and boundary definitions, not the database, so if any of this sounds bad/not right, sorry.

tonysurma:
@MisterJames I am good with that

auroraocciduusadmin:
@MisterJames , @mgmccarthy, @tonysurma I'm reluctant to comment since I don't know what the intent was when the PostalCodeGeo class was first implemented. Also, I'm just getting my feet wet with this project. 
However, it won't seem to have too much impact, from what I've seen, if Location's PostalCodeGeo is not a FK since it's already a "null" -type in the database.

tonysurma:
@auroraocciduusadmin thank you for getting your feet wet with this project 👣 !  You are free to tear it down (the FK) and I would be thankful for you doing it.

auroraocciduusadmin:
@tonysurma OK. Can you rollback this pull request?

tonysurma:
cc @BillWagner re rollback request

auroraocciduusadmin:
The last commit replaces Location's PostalCode property with a string. Some tests no longer make sense and were removed.

Issue #661, opened 2016-04-06
write unit tests for AccountController


mgmccarthy:
@tonysurma please label with "jump-in". Part of Issue #620

stubs are provided in:
AccountControllerTests.

digitaldrummerj:
@tonysurma Is this issue still valid?  I see a couple of PRs merged back in April and May but no activity since then.

mgmccarthy:
@tonysurma @digitaldrummerj ... there are still skipped tests in AcccountControllerTests so this is still a valid issue

ExoltenOne:
If no one is working on this issue I could take it.

stevejgordon:
Thanks for jumping in @ExoltenOne. If you have any questions as you progress, please just let us know. I have invited you to the repo so you should have an email from GitHub. It's not necessary to begin working but once you have accepted it will allow us to assign you to this issue formally. What you've done here is perfect though to let others know you are working on this issue. Thanks again!

ExoltenOne:
I got the email and obviously accepted the invitation. Please, assign me to this issue.

stevejgordon:
@ExoltenOne You're assigned. 😄

PR #729, opened 2016-04-30
Updating unit tests for Account Controller #661

klabranche:
#661 - write unit tests for account controller.  Have updated 37 of the
tests to run.  Updated all the attribute tests and the simple get view
tests.


MisterJames:
Great work here again @klabranche.

I'm pretty sure we'll be able to simplify this when we amalgamate the authentication bits.

:shipit:

PR #732, opened 2016-05-02
Fixing the error where un-enroll from a task was not working

### LINKED ISSUE #720: Un-enrolling from a task fails
I was able to produce this locally by enrolling to a task and then clicking the un-enroll option.

I see the following js console error

image

stevejgordon:
Fixes #720

Initially the issue was caused by the unenroll taskSignup query not including the required child entities and therefore causing a null ref exception to be thrown - resulting in no actual unenroll occuring. Having fixed this I noticed that the calls to remove the signup records was not behaving as expected and only the Task or Event ids were affected. The records never got removed.

This seems to be due to those related elements not being attached and actually removed. I need to dig further into the behaviour of EF in those cases but to resolve this issue I have modified the flow to involved pulling back the actual entities (taskSignup and potentially eventSignup) which need to be deleted. This ensures they are tracked in the context when calling the remove method. This has added an extra DB call as a result.

I have added tests to validate the behaviour with using in memory context.


tonysurma:
Thx once @MisterJames or others can make a review I'll merge

MisterJames:
:shipit:

PR #745, opened 2016-05-10
deleting DeletUserCommand/Handler and moving code back into SiteAdminController

### LINKED ISSUE #744: delete DeleteUserCommand/Handler and move UserManager<ApplicationUser> call back into SiteAdminController
this Issue is based off the conversation in Issue #567. We will not be using Mediator to abstract calls to `UserManager<ApplicationUser>` or `SignInManager<ApplicationUser>`, b/c it's an unneeded abstraction.

Once the `UserManager<ApplicationUser>` is returned to `SiteAdminController` (where `UserManager<ApplicationUser>` is already injected), I'll fix the unit test to reflect the change.

mgmccarthy:
Fixes #744


AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SiteAdminController.cs
```diff
@@ -61,7 +61,9 @@ public IActionResult Index()
         [ValidateAntiForgeryToken]
         public async Task<IActionResult> ConfirmDeleteUser(string userId)
         {
-            await _mediator.SendAsync(new DeleteUserCommand { UserId = userId });
+            var user = await _userManager.FindByIdAsync(userId);
+            await _userManager.DeleteAsync(user);
```
MisterJames: So...I'm okay doing this, but I think we are still going to need to do a `_mediator.Publish` and punch out an event with the user ID. If we are to 'clean up' in some way when a user is removed (i.e. unassign tasks, remove subscriptions, etc.) we'll need a hook to listen for this type of change. Thoughts

mgmccarthy: @MisterJames, I didn't realize the end game was we wanted to published a `UserDeleted` event. As far as accepting this PR, I have two questions for you:

1. is there anywhere else in the system that is doing the same operation? (aka, deleting a user). If so, then we DO want this deletion code in a handler, and those two controllers that are deleting the user should send a command to do it.
2. If there are two places that delete a user, and we keep the delete call in the current handler, we should publish an event from that command handler.

MisterJames: Hrm...well, we don't want a handler for the sake of a handler, but all that extra cruft has to be taken care of somehow. And our only other option here would be to revert to injecting the DBContext (or repository) into the controller. 

As far as the delete goes, the only occurances 
 - are in Admin 
 - might be in account self-termination (down the road)
 - are likely edge cases

So, IMO I'm okay leaving the delete here for now and we can likely start to raise the event when we get to the point where we're taking care of the extraneous data. For now, there's no keys/integrity defined afaik when we nuke a user, it  just leaves those rows orphaned. Again, since it's an edge case and currently only in admin, I think we're okay.

Sorry for over-thinking it ;)

:shipit:

mgmccarthy: Actually, the PR uses the injected `UserManager<ApplicationUser>` to delete the user, so we would not need to inject any other dependencies.

If we discover going forward we need to perform this operation from multiple places, we can always revert to the handler doing the deletion taking `UserManager<ApplicationUser>` as a dependency. 

Not over thinking it, it's a good convo to have. Thanks for taking the time to explain.

PR #755, opened 2016-05-18
Clean up model validators

### LINKED ISSUE #728: Abstract CampaignSummaryModelValidator, EventDetailModelValidator and LocationEditModelValidator to make them injectable
Right now, we're using the "new" keyword to create these validators in controllers and other validators. This violates SRP as well as DI principles as well as complicates controller unit testing.

UPDATE: now that the `PostalCodeGeo` property on `Location` has been changed to `string PostalCode`, some of these validations can be deleted as they contain empty implementation.

And, here are the changes that will be in the PR:
- deleted `LocationEditModelValidato`r b/c it contains no implementation
- moved code from `CampaignSummaryModelValidato`r back to `CampaignController` b/c it's only used in one place
- deleting `CampaignSummaryModelValidator` and it's unit tests
- updated `CampaignController` unit tests to test for code moved from `CampaignSummaryModelValidator`
- removed constructor and call to `LocationEditModelValidator` from `EventDetailModelValidator`
- updated `EventDetailModelValidatorShould` unit tests to test the actual error message being returned by `EventDetailModelValidator` instead of just the existence of a key name
- added `IValidateEventDetailsModelinterface` which is implemented in `EventDetailModelValidator` and injected the interface into `EventAdminController` instead of using the "new" keyword in action methods that need the validation
- registered `IValidateEventDetailModels` with StartUp code
- unit tests are now contained to validating `EventDetailModels` to the Validator unit test, and the `EventAdminController` is now free of having to provide implementation for a validation that is created in-line in two action methods
- updated all unit tests

mgmccarthy:
Fixes #728

changes in this PR:
- deleted `LocationEditModelValidator` b/c it contains no implementation
- moved code from `CampaignSummaryModelValidator` back to `CampaignController`
- deleting `CampaignSummaryModelValidator` and associated unit tests
- updated `CampaignController` unit tests to test for code moved from `CampaignSummaryModelValidator`
- removed constructor and call to `LocationEditModelValidator` from `EventDetailModelValidator`
- updated `EventDetailModelValidatorShould` unit tests to test the actual error message being returned by `EventDetailModelValidator` instead of just the existence of a key name
- added `IValidateEventDetailModelinterface` which is implemented in `EventDetailModelValidator`.  `IValidateEventDetailModelinterface` is injected into `EventAdminController` instead of using the "new" keyword in action methods that need the validation
- registered `IValidateEventDetailModels` with StartUp code
- unit tests for `EventDetailModelValidator` are now contained in `EventDetailModelValidatorShould` and the `EventAdminController` is now free of having to provide implementation for a validation that is created in-line in two action methods
- updated all unit tests


AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -203,7 +229,7 @@ public void EditPostHasHttpPostAttribute()
         }
 
         [Fact(Skip = "NotImplemented")]
-        public void EditPostHasHValidateAntiForgeryTokenttribute()
+        public void EditPostHasValidateAntiForgeryTokenttribute()
```
MisterJames: lol...you fixed the one spelling mistake but not the other?! What do you have against the letter 'a'?!

MisterJames:
:shipit: 

We can fix the missing A when the test is actually implemented :)

Nice PR.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
mgmccarthy: I can fix real quick. Nothing against the letter 'a' ,I swear! :stuck_out_tongue_winking_eye:

Issue #469, opened 2016-01-05
Organization Name - Should enforce uniqueness


featuresnap:
When creating a new organization (using Admin..Organizations and then Create New), the application allows creation of multiple organizations with the same name. This could lead to confusion among many similar organizations. Recommend addition of a uniqueness constraint at a minimum.

manojattal:
Hello @ksk100, Are you working on this? Otherwise I can take it up!

tonysurma:
@manojattal if you haven't heard anything go for it.  thanks

manojattal:
I need help in testing my changes. I tried to login with admin user and org admin user, but I do not see an option to create organization. Can someone tell me steps to create an organization?

stevejgordon:
Hi @manojattal. If you login as the administrator@example.com user then you should be able to create an organization.

The from the admin menu, choose Organizations

![image](https://cloud.githubusercontent.com/assets/3669103/13462088/762503e4-e07d-11e5-92ad-17d36d70820d.png)

Then at the bottom of that list you can choose create new

![image](https://cloud.githubusercontent.com/assets/3669103/13462106/8be65af2-e07d-11e5-9574-9c3aea75d3ca.png)

I hope that helps?

tonysurma:
@manojattal any update?  Thx

mgmccarthy:
@tonysurma I think this Issue can be closed. Looks like PR #757 was merged.

digitaldrummerj:
@tonysurma bump

PR #757, opened 2016-05-18
 Organization Name - Should enforce uniqueness #469

manojattal:
While adding/editing organization, a duplicate check is being done. An
existing unit test case for edit organization is modified and 2 unit
cases added for add and remove organization.


AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/OrganizationController.cs
```diff
@@ -69,8 +73,21 @@ public IActionResult Edit(OrganizationEditModel organization)
             
             if (ModelState.IsValid)
             {
-                int id = _mediator.Send(new OrganizationEditCommand { Organization = organization });
-                return RedirectToAction("Details", new { id = id, area = "Admin" });
+                var originalOrganization = _dataAccess.GetOrganization(organization.Id);
+                if (originalOrganization != null)
```
MisterJames: Rather than re-introducing a dependency on `IAllReadyDataAccess` here, let's instead run with the Query/Handler pattern to validate.

MisterJames:
@manojattal there is an example in the code here with a query/handler already in place. Would you mind shifting to that to validate the org name? We are trying to eliminate all use of the `AllReadyDataAccess` class throughout the application, so we don't want to reintroduce it here. Thanks!

mgmccarthy:
@MisterJames , perhaps it's better to have another query/handler that actually checks for org name uniqueness, then based on the return value, sends OrganizationEditCommand?

For example:
```
public class OrganizationNameUniqueQuery : IRequest<bool>
{
    public string OrganizationName { get; set;}
}
```

the handler would check for unique name then return true or false

When we send a command, it should have a high chance of succeeding, so fronting any type of model and/or business validation checks (in this case, is org name unique?) should be done in the controller's action method. Especially is there is only one place to create an Organization, which it looks like is the case here.

That way, our handler that is optimized for the "write" side of CQRS is not saddled with a bunch of model and business validations.

Thoughts?

MisterJames:
Yeup, you've got it, I think we're saying the exact same thing.

tonysurma:
@MisterJames and others, what is the status of this?  Thanks

manojattal:
@MisterJames, @tonysurma @mgmccarthy I have done changes in this pull request as per suggested using query/handler. Let me know if any changes required.

mgmccarthy:
@MisterJames @tonysurma do you need a code review of this to merge?

tonysurma:
If @MisterJames could glance at it again I will merge.  Thanks

mgmccarthy:
@tonysurma, looking at the changes in [	b09f337](https://github.com/HTBox/allReady/pull/757/commits/b09f3370eadfca4fc63ed186b44919adfe505465) the correct message/handler code was added, the OrganizationController was updated to use mediator to send, and that unit tests were updated accordingly. it looks good to me.

The only improvement I can think of is to make the action method, mediator call, and handler for the query message async, but that's an optimization we can add later. If we're not using `IAllReadyDataAccess` (which has no async read members) in the handler, and instead using AllReadyDataContext, we should be able to support async

tonysurma:
The build after I incorporated this is failing https://ci.appveyor.com/project/HTBox/allready/build/1.0.0.1017.master

Any ideas @manojattal or @mgmccarthy ? cc @MisterJames 

The issue was:
3759C:\projects\allready\AllReadyApp\Web-App\AllReady.UnitTest\Areas\Admin\Controllers\OrganizationControllerTests.cs(378,49): DNX,Version=v4.5.1 error CS1061: 'IAllReadyDataAccess' does not contain a definition for 'GetOrganization' and no extension method 'GetOrganization' accepting a first argument of type 'IAllReadyDataAccess' could be found (are you missing a using directive or an assembly reference?) 
 3760C:\projects\allready\AllReadyApp\Web-App\AllReady.UnitTest\Areas\Admin\Controllers\OrganizationControllerTests.cs(379,49): DNX,Version=v4.5.1 error CS1061: 'IAllReadyDataAccess' does not contain a definition for 'Organizations' and no extension method 'Organizations' accepting a first argument of type 'IAllReadyDataAccess' could be found (are you missing a using directive or an assembly reference?) 
 3761 
 3762Build failed. 
 3763    0 Warning(s) 
 3764    2 Error(s)

mgmccarthy:
@tonysurma in Commit 3086bd81, @robrich removed the Organization crud members from `IAllReadyDataAccess`. We should be using `AllReadyContext` instead of that interface, b/c it looks like the project is leaning towards using in-memory EF db to run unit tests instead of relying on mocking out members of IAllReadyDataAccess

@manojattal should get latest, rebase and fix the unit test code so it no longer references IAllReadyDataAccess.

tonysurma:
ah nice find @mgmccarthy 

@manojattal any chance you can take that on?

PR #758, opened 2016-05-18
add license, description, and repository to package.json

robrich:
to avoid console warnings during build


MisterJames:
:shipit: 

Thanks @robrich!

PR #760, opened 2016-05-18
Adding link on home page to full campaign list

### LINKED ISSUE #691: 6.1.1.3	Display link to list of campaigns on home page
Below the featured campaign provide a link to the campaign list page which shows campaigns that are current/upcoming.  Remove map link and add in logo/image of campaign on home page

MisterJames:
Fixes #691.

Couple of things:
 - the list of campaigns is still on the home page and it should (perhaps?) remain, in the event that there is no configured feature campaign. Thoughts @tonysurma?
 - The featured campaign has a logo displayed if set in the DB.
 - There appears to be no UI yet to set a featured campaign.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/HomeController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/HomeController.cs
@@ -19,6 +19,8 @@ namespace AllReady.Controllers
         {
             var model = new HomePageViewModel();
 
+            //todo: per #691 the model may need to be updated as we 
+            //      no longer require the list of campaigns
             model.Campaigns = mediator.Send(new CampaignQuery());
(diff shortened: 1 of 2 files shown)


tonysurma:
Okay to leave list, we want a better UI for that eventually

MisterJames:
Beauty, thanks @tonysurma. Should I wrap with an `if` then?
```
@if (Model.HasFeaturedCampaign)
{
 // ... show featured
} else {
 // ... show list
}
```

MisterJames:
Okay, I've added the toggle. If a featured campaign exists, it will display. In the absence of one, it will instead dump out the list of campaigns.

image

tonysurma:
looks good to me :shipit:

Issue #710, opened 2016-04-17
DateTime issue on loading Campaigns page 


sudhakarssd:
Exception on loading campaigns page for admin login. **" ArgumentOutOfRangeException: The UTC time represented when the offset is applied must be between year 0 and 10,000. Parameter name: offset"**
Exception occurred on  /Areas/Admin/Views/Campaign/Index.cshtml at line no. 56
 `@if (item.StartDate != DateTime.MinValue)`

Error is due to difference in local system timezone. It might be a good idea to use ToUniversalTime()
 `@if (item.StartDate != DateTime.MinValue.ToUniversalTime())`

mgmccarthy:
@tonysurma @MisterJames @BillWagner 

what is the approach to Utc vs. local time for allReady? For example, I see that Campaign stores a `StartDateTime` and `EndDateTime` as well as a `TimeZoneId`. Are the `StartDateTime` and `EndDateTime` stored as UTC or local?

After reading this bug, then tracing through how we handle utc based conversions in the system, it seems a bit all over the map. We're using DateTime.Now in views, models, controllers, handlers, and DateTime.UtcNow in other places...

What is the expectation of storing DateTime values in the entities themselves? Are these DateTime values stored as utc or local time?

When reading these values, do we convert to local time based on the stored utc time + timezone values? What about writing the values? My assumption is the creation of entities that have start and end date would be submitted in local time, then we translate to utc before storing it?

tonysurma:
I will let others chime in and add but from a functionality point of view:
All times should be displayed in local times relative to time zone of the campaign (much like flight times should all be displayed local to airport time zone metaphorically).  To make a fine point if Tony in Central time zone looks at campaign in Pacific time zone then it should show times in Pacific time zone not in central.

In my mind that means gathering and storing as UTC and rendering relative to time zone of campaign but up for discussion.  If you find a bunch of issues with that being out of pattern we could create them as issues for 'jump-in'

stevejgordon:
I'm not fully clear on date handling myself but since we are using DateTimeOffset for the storage that should carry the timezone for the date with it and allows for conversion to from UTC time as well. It might be useful to think about any repeated datetime calculations and whether we can provider some kind of abstraction around it and a class to manage any date time related work. I've yet to dig into it in detail though to see what we're doing where.

mgmccarthy:
@tonysurma  @stevejgordon. 

[...]

PR #789, opened 2016-05-21
Adding task model validator

### LINKED ISSUE #782: Itinerary tasks should not span more than 1 day
In relation to the second item from issue #706 and the discussion on the community standup (14th May 2016) the validation rules for tasks need some adjustment to comply with the business rules.

As such a task for an itinerary event should not be allowed to scope outside of a single day.

A validator should be in place to ensure task creation and edits validate this before saving / updating a task occurs.

stevejgordon:
Fixes  #782 

The validator checks the start date and end date properties on the TaskSummaryModel used when creating / editing a task. Dates are validated against the business rules and this includes the addition of validation that an itinerary task does not span more than a day.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/AllReady.UnitTest.xproj
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/AllReady.UnitTest.xproj
@@ -17,8 +17,5 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\..\AllReady.Models\AllReady.Models.csproj" />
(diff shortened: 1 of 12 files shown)


mgmccarthy:
nice PR @stevejgordon! If you want to do the time zone refactoring in steps, we could always create providers to use in the system based on of DateTimeOffset first, then refactor to the abstraction that I mentioned in #710. I know June is a soft launch, so small steps at this point might be better than large steps

stevejgordon:
@mgmccarthy Thanks. On the date part, I agree, given the nearing pilot date probably best to tackle in small pieces. I'm going to try and focus any time I get towards the remaining issues prior to the pilot but can probably find small gaps for helping on the date stuff.

tonysurma:
Thanks.  @MisterJames unless you see an issue in review I will merge this in.

Issue #772, opened 2016-05-18
Unable to delete an Event if it contains any Tasks


VishalMadhvani:
SqlException: The DELETE statement conflicted with the REFERENCE constraint "FK_AllReadyTask_Event_EventId". The conflict occurred in database "AllReady", table "dbo.AllReadyTask", column 'EventId'.
 The statement has been terminated. 

Example Url: ```/Admin/Event/Delete/4```

VishalMadhvani:
This can be closed.

PR #791, opened 2016-05-25
#772 Unable to delete an Event if it contains any Tasks

arst:


tonysurma:
@MisterJames or someone else can I get a second set of eyes on the migration before I merge it in?

Looks good to me and thanks @arst for the contribution

MisterJames:
Yeah, it's noisy because of the way that EF manages changes to FK relationships, but this looks okay to merge. Thanks @arst! :shipit:

PR #805, opened 2016-06-05
Rename EventTypes: Task -> Rally, delete EventManaged

robrich:
TODO: what to do with EventManaged-specific logic?


robrich:
FRAGILE: Don't merge this yet.  I'd like to discuss how we handle the commented-out EventManaged logic before we push this into place.

tonysurma:
For now, delete and we can always bring it back in.

Also i if you prefix the pr name with wip while it's a work in progress that's our convention to know to not merge yet

AllReadyApp/Web-App/AllReady/Models/EventType.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace AllReady.Models
+{
+    public enum EventType
```
MisterJames: yay for non-pluralized enums! :)

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Tasks/AssignTaskCommandHandlerAsync.cs
```diff
@@ -69,7 +69,7 @@ public AssignTaskCommandHandlerAsync(AllReadyContext context, IMediator mediator
                 var taskSignupsToRemove = task.AssignedVolunteers.Where(taskSignup => message.UserIds.All(uid => uid != taskSignup.User.Id)).ToList();
                 taskSignupsToRemove.ForEach(taskSignup => task.AssignedVolunteers.Remove(taskSignup));
 
-                if (campaignEvent.EventType != EventTypes.EventManaged)
+                if (true /*campaignEvent.EventType != EventTypes.EventManaged*/)
```
MisterJames: If deleting the EventManaged per @tonysurma I'm sure we can remove the conditional blocks as appropriate (which I think was related to your query on if we should keep them). Should be safe to remove.

MisterJames:
@robrich I am good with direction here and agree with @tonysurma on the delete note. If you clean up those conditional bits we should be good to go. Also +1 on the depluralization ;)

robrich:
Should be good to go now.  Merge when ready.

tonysurma:
:shipit:

robrich:
How are you rebasing these as you merge?  I've not found such a button on GitHub's site.  Awesome, btw.

robrich:
Will this cause stale data issues for any events currently stored in the db as `EventType=1`?

tonysurma:
@robrich if you can verify no sample data is injected (I doubt it as you would have found it already) I will verify database (on my list of things to do)

robrich:
I patched sample data generation as I made the changes as they're compile time errors.  :D

tonysurma:
yep I figured as such

PR #810, opened 2016-06-07
Implemented some unit tests stubs for Admin/CampaignController

auroraocciduusadmin:
Also modified CampaignListQueryHandler to fetch "ManagingOrganization" since it is needed for the returned CampaignSummaryModels.


tonysurma:
@mgmccarthy can you take a peek?

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -23,21 +23,40 @@ public class CampaignAdminControllerTests : InMemoryContextTest
         //delete this line when all unit tests using it have been completed
         private readonly Task taskFromResultZero = Task.FromResult(0);
 
-        [Fact(Skip = "NotImplemented")]
+        [Fact()]
```
mgmccarthy: replace `[Fact()]` with `Fact[]` for the unit tests you implemented in this file

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -23,21 +23,40 @@ public class CampaignAdminControllerTests : InMemoryContextTest
...
         public void IndexSendsCampaignListQueryWithCorrectDataWhenUserIsOrgAdmin()
         {
+            Organization aginAware = AgincourtAwarenessExtended(Context);
+            Organization pwPrepares = ParkwoodsPreparesExtended(Context);
```
mgmccarthy: this line is not needed for test setup and should be deleted

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -23,21 +23,40 @@ public class CampaignAdminControllerTests : InMemoryContextTest
...
+            Context.SaveChanges();
+
+            // Org admin should only see own campaigns
+            CampaignController controller = CampaignControllerWithListQuery(Context, UserType.OrgAdmin.ToString(), aginAware.Id);
+            ViewResult view = (ViewResult)controller.Index();
+            IEnumerable<AllReady.Areas.Admin.Models.CampaignSummaryModel> viewModel = (IEnumerable<AllReady.Areas.Admin.Models.CampaignSummaryModel>)view.ViewData.Model;
```
mgmccarthy: you can delete the fully qualified `AllReady.Areas.Admin.Models.` namespace form the font of CampaignSummaryModel

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -23,21 +23,40 @@ public class CampaignAdminControllerTests : InMemoryContextTest
...
+            Assert.Equal(viewModel.Where(t => t.OrganizationId == aginAware.Id).Count(), aginAware.Campaigns.Count());
```
mgmccarthy: this line can use .Count directly instead of .Where and then .Count:
`Assert.Equal(viewModel.Count(t => t.OrganizationId == aginAware.Id), aginAware.Campaigns.Count);`

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -23,21 +23,40 @@ public class CampaignAdminControllerTests : InMemoryContextTest
...
         }
-            
-        [Fact(Skip = "NotImplemented")]
+
+
+        [Fact()]
         public void IndexSendsCampaignListQueryWithCorrectDataWhenUserIsNotOrgAdmin()
         {
+            Organization aginAware = AgincourtAwarenessExtended(Context);
+            Organization pwPrepares = ParkwoodsPreparesExtended(Context);
```
mgmccarthy: the above two lines are not used in the unit test setup and should be deleted

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -23,21 +23,40 @@ public class CampaignAdminControllerTests : InMemoryContextTest
...
+            Context.SaveChanges();
+
+            // Site admin should see all campaigns
+            CampaignController controller = CampaignControllerWithListQuery(Context, UserType.SiteAdmin.ToString(), null);
+            ViewResult view = (ViewResult)controller.Index();      
+            IEnumerable<AllReady.Areas.Admin.Models.CampaignSummaryModel> viewModel = (IEnumerable<AllReady.Areas.Admin.Models.CampaignSummaryModel>)view.ViewData.Model;
```
mgmccarthy: delete fully qualified namespace `AllReady.Areas.Admin.Models.` text

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
auroraocciduusadmin: I believe those lines are needed because they insert some organizations and campaigns into the context. The test then tries to see if only the campaigns that match the organization are returned.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
auroraocciduusadmin: The two lines are needed because they set up the context with data that the controller will fetch.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
auroraocciduusadmin: ...also, "aginAware" is used in the Assert() and also in the Claims for the controller.

mgmccarthy:
@auroraocciduusadmin, first, thank you for the contribution. For some of these unit tests I would focus the test just on what the function does (aka mock mediator and verify the send is called) as opposed to capturing the entire flow from action method invocation to what happens in the handler and the database. 

For example, for the `IndexSendsCampaignListQueryWithCorrectDataWhenUserIsOrgAdmin` unit test, we only really need to test that the mediator sent CampaignListQuery with the correct data on it. This data is determined by whether or not a user is an OrgAdmin or not.

I can help guide that if you are interested.

mgmccarthy:
@auroraocciduusadmin, you're right on the money with the last commit. Some other recommendations:

- `EditPostInsertsCampaign` should be updated to assert that  `EditCampaignCommand` is sent with the correct data via mediator mock (similar to `IndexSendsCampaignListQueryWithCorrectDataWhenUserIsOrgAdmin`). We don't need DbContext for this test
- `IndexReturnsCorrectDataWhenUserIsOrgAdmin` and `IndexReturnsCorrectDataWhenUserIsNotOrgAdmin` can be changed to assert the correct `CampaignSummaryModel` is being returned from a mocked mediator.  We don't need to use dbcontext to test these two actions. If you look at `CampaignListQueryHandlerTests`, you'll see the condition of the user being an OrgAdmin or not and what data is being returned from that handle method is already being tested.

mgmccarthy:
@auroraocciduusadmin, nice work! Thanks for the contribution. @tonysurma you can accept this PR.

PR #815, opened 2016-06-07
Write git hash to version.json in deploy.cmd too, close #314

### LINKED ISSUE #314: some indication of which commit current build is from?

robrich:


tonysurma:
if it builds I will :shipit: thx @robrich

tonysurma:
It's so beautiful 😂 thank you!

snip_20160607105154

PR #821, opened 2016-06-07
Updating Request model to allow relationship with event.

### LINKED ISSUE #813: Request should map to Event
The Request DB model needs a navigational property to link it to an event.

A request can link to a single event. An event can have many requests.

### LINKED ISSUE #814: Latitude property is misspelt on Request
Lattitude currently has three t's but should have two - i.e. Latitude

stevejgordon:
Fixes #813 and fixes #814

This has been added as optional in case we do not know the event id when request is added via api. This could allow unassigned requests to be managed and linked to events via the application if necessary. Also as this is an optional relationship we ensure that on deletion of an event we won't cascade deletion of requests. They will instead be orphaned so we retain their history.

@MisterJames If this is not the behaviour we want then we can adjust this mapping.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Migrations/20160607185847_RequestToEventMapping.Designer.cs
+++ b/AllReadyApp/Web-App/AllReady/Migrations/20160607185847_RequestToEventMapping.Designer.cs
@@ -1,0 +1,794 @@
+using System;
+using Microsoft.Data.Entity;
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Migrations;
+using AllReady.Models;
+
+namespace AllReady.Migrations
+{
+    [DbContext(typeof(AllReadyContext))]
+    [Migration("20160607185847_RequestToEventMapping")]
+    partial class RequestToEventMapping
+    {
+        protected override void BuildTargetModel(ModelBuilder modelBuilder)
+        {
+            modelBuilder
+                .HasAnnotation("ProductVersion", "7.0.0-rc1-16348")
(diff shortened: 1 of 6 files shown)


tonysurma:
@stevejgordon @MisterJames should I merge?

stevejgordon:
@tonysurma I think if you can merge as is for now. I can then continue with anything that relies on this mapping. We can always revise the mapping and db rules if we decide on a different approach. It would be best though that if we do change it, we try to do it before we have live data.

tonysurma:
Build errors @stevejgordon ?

3630C:\projects\allready\AllReadyApp\Web-App\AllReady.UnitTest\Areas\Admin\Features\Events\EventDetailQueryHandlerTests.cs(88,17): DNX,Version=v4.5.1 error CS0117: 'Request' does not contain a definition for 'Lattitude' 
 3631C:\projects\allready\AllReadyApp\Web-App\AllReady.UnitTest\Areas\Admin\Features\Events\EventDetailQueryHandlerTests.cs(97,17): DNX,Version=v4.5.1 error CS0117: 'Request' does not contain a definition for 'Lattitude' 
 3632C:\projects\allready\AllReadyApp\Web-App\AllReady.UnitTest\Areas\Admin\Features\Events\EventDetailQueryHandlerTests.cs(106,17): DNX,Version=v4.5.1 error CS0117: 'Request' does not contain a definition for 'Lattitude'

stevejgordon:
@tonysurma Let me pull in and check. Is it best to put in a new PR which fixes this?

stevejgordon:
@tonysurma PR submitted to fix this.

tonysurma:
yes a PR 👍 and thanks for doing it already

stevejgordon:
No problem. Not sure why I didn't get / notice the failure after I rebased.

tonysurma:
me neither and the build on the PR passed as well

Issue #787, opened 2016-05-20
finish writing unit tests TaskStatusChangeHandler


mgmccarthy:
@tonysurma please label with "jump-in". Part of Issue #620

rmcc13:
If no one is working on this, I would like to pick it up.

mgmccarthy:
@rmcc13. I'll try to get a to a code review of this tonight. Since I'm not a project owner, I don't have the ability to set Assignees, so if you're working on a PR, you can include a comment on this issue that you're currently working on it. Thanks

rmcc13:
OK sounds good, I'm working on this.

rmcc13:
Here is an overview of what I am working on. The last commit included unit tests for success verifying expected behavior. Each test made two assertions: 1) Make sure the `TaskSignup` object has the expected `TaskStatus` value and 2) make sure the handler's `Handle` method published a `TaskSignupStatusChanged` notification once, and only once.

#### Current tests from last commit:

- Setting `TaskStatus` to "Assigned"
- Transitioning `TaskStatus` from "Assigned" to "Accepted"
- Transitioning `TaskStatus` from "Assigned" to "Rejected"
- Transitioning `TaskStatus` from "Accepted" to "Completed"
- Transitioning `TaskStatus` from "Accepted" to "CanNotComplete"

#### Here is what I am planning to work on next:

###### Tests for success

- [ ] Transition `TaskStatus` from "CanNotComplete" to "Accepted"
- [ ] Transition `TaskStatus` from "Completed" to "Accepted"
- [ ] Transition `TaskStatus` from "Assigned" to "Completed"
- [ ] Transition `TaskStatus` from "Assigned" to "CanNotComplete"

###### Tests for failure

- [ ] Transitioning `TaskStatus` from "Accepted" or "Rejected" to "Accepted" should throw an exception
- [ ] Transitioning `TaskStatus` from "Accepted", "Rejected", "Completed" or "CanNotComplete" to "Rejected" should throw an exception
- [ ] Transitioning `TaskStatus` from "Rejected", "Completed" or "CanNotComplete" to "Completed" should throw an exception
- [ ] Transitioning "Rejected", "Completed", or "CanNotComplete" to "CanNotComplete" should throw an exception
- [ ] If the `Task` or `TaskSignup` objects do not exist, an exception should be thrown

*Note* These tests all enforce the current implementation, I'm not sure if all these tests are necessary, but I plan on working on these unless others say they aren't necessary.

[...]

PR #836, opened 2016-06-08
Basic behavior tests

rmcc13:
Here is an overview of what I have written tests for so far. I have more comments in #787 

- Setting TaskStatus to "Assigned"
- Transitioning TaskStatus from "Assigned" to "Accepted"
- Transitioning TaskStatus from "Assigned" to "Rejected"
- Transitioning TaskStatus from "Accepted" to "Completed"
- Transitioning TaskStatus from "Accepted" to "CanNotComplete"


mgmccarthy:
@rmcc13, the contribution looks good. A couple suggestions:

- in every test, we're testing the taskSignup.Status being equal t othe TaskStatus.Completed.ToString():
`taskSignup.Status.ShouldBe(TaskStatus.Completed.ToString());`
The code is also setting the incoming message's TaskStatusDescription to the taskSignup.StatusDescription:
`taskSignup.StatusDescription = message.TaskStatusDescription;`
so you could assert that value in each of your unit tests

- if you look at `UpdateMyTasksHandlerAsync` and it's associated unit tests, you can see by using this line in the handler:
`public Func<DateTime> DateTimeUtcNow = () => DateTime.UtcNow;`
you can supply a DateTime in your unit test to the handler and also test `taskSignup.StatusDateTimeUtc` equals the correct DateTime.Utc.

this way, we're explicitly testing all of the properies set on taskSignup that is being returned from `TaskStatusChangeHandlerAsync`.

rmcc13:
OK, thanks for the suggestions. I will look into this over the weekend.

rmcc13:
@mgmccarthy I'm not quite sure if I follow your comment 100%. In your first point, you mentioned all tests were were testing `taskSignup.Status.ShouldBe(TaskStatus.Completed.ToString())`. I'm not sure if you meant we are only testing when tasks are completed, or if you were just using that line of code as an example. I'm assuming the latter. I did however change that test to compare against the command object
`taskSignup.Status.ShouldBe(command.TaskStatus.ToString())`

I am now testing all TaskSignup properties that could be changed in the Handle method. I changed the status test to use the command object, because all other tests are comparing against the command object, and I think the intent is more explicit as we are expecting the status to now be the value we passed to the Handle method.

mgmccarthy:
@rmcc13, looks good. Thanks for all the work. @tonysurma, you can accept this PR when get a chance

PR #837, opened 2016-06-09
Add team member to itinerary

### LINKED ISSUE #835: Add ability to add team members to itinerary
We need to be able to select from potential team members and add them to an itinerary. The following rules must apply.

They must be assigned (and accepted) to a task from the same event. That task must occur on the same date as the itinerary is scheduled for. Only users who are not part of an itinerary for that day already should be shown.

@tonysurma I'm working on this currently.

stevejgordon:
Closes #835 

Added the ability to quick add a team member to an itinerary from a list of potential team members.
To make adding a team member a quick process (less clicks) this is done from a form directly on the Itinerary Details page.

If any potential team members exist they are shown in a dropdown list above the list of assigned team members.

image

If no potential team members exist the form is not shown

image

**A potential team member is defined as follows:**
A person signed up to (and with accepted status) a task from the same event which is scheduled for the same date as the itinerary. We only show task signups that are not already linked to an itinerary. This assumes that the task signup and assignment logic will restrict a user from being assigned to two tasks within the same timeframe.

@tonysurma - The point above warrants some consideration and review as I'm not sure this rule is fully enforced? *Q1* For an itinerary event it would seem sensible to restrict signups by a user to a single task on a given date. However I'm not sure how that may affect all use cases.

What if we have an AM and PM session and want to assign a user to both? Since an itinerary is date bound only I'm not sure how we might want to handle the business rules. *Q2*

It feels better that would should check assignment to tasks for conflicts since in an edge case someone may try to volunteer for tasks from two different events/campaigns on the same day and time which we should also try to prevent by calculating their existing busy times. But that route opens further questions from me!!

cc/ @OhMcGoo 

NOTE:
The action handling this post is a little basic at the moment. If it fails for any reason it simple reloads the Itinerary details currently. There is an edge case where someone may be assigned by a different organizer to another itinerary between the list of potential team members being created and the user adding them. In that case our Mediator handler will catch and prevent a double assignment, however there will be no feedback on the UI. I have left a todo for this and will open an issue to enhance with knockout and improved messaging.

NOTE: No tests added for the AddTeamMemberCommandHandlerAsync at this point due to time. I will raise an issue for this to be addressed in a future PR.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/ItineraryAdminControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/ItineraryAdminControllerTests.cs
@@ -344,6 +344,68 @@ namespace AllReady.UnitTest.Areas.Admin.Controllers
             Assert.IsType<BadRequestResult>(result);
         }
 
+        [Fact]
+        public void AddTeamMemberHasHttpPostAttribute()
+        {
+            var sut = new ItineraryController(Mock.Of<IMediator>(), MockSuccessValidation().Object);
+            var attribute = sut.GetAttributesOn(x => x.AddTeamMember(It.IsAny<int>(), It.IsAny<int>())).OfType<HttpPostAttribute>().SingleOrDefault();
+            Assert.NotNull(attribute);
+        }
+
+        [Fact]
+        public void AddTeamMemberHasRouteAttributeWithCorrectRoute()
+        {
+            var sut = new ItineraryController(Mock.Of<IMediator>(), MockSuccessValidation().Object);
+            var routeAttribute = sut.GetAttributesOn(x => x.AddTeamMember(It.IsAny<int>(), It.IsAny<int>())).OfType<RouteAttribute>().SingleOrDefault();
+            Assert.NotNull(routeAttribute);
+            Assert.Equal(routeAttribute.Template, "Admin/Itinerary/AddTeamMember");
+        }
+
+        [Fact]
+        public void AddTeamMemberHasValidateAntiForgeryAttribute()
+        {
+            var sut = new ItineraryController(Mock.Of<IMediator>(), MockSuccessValidation().Object);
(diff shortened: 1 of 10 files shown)


tonysurma:
I added two tags to relate to the questions above.  So:

For *Q1 "For an itinerary event it would seem sensible to restrict signups by a user to a single task on a given date"* The answer is as follows: Yes that is sensible, no that isn't enforced right now during sign up and that is *ok* as if that happens for now the organizer will deal with it and so we are ok.  That being said in the future (*TODO*) I should add an issue for this as a user should be warned if they are 'double booking' themselves.

For *Q2 "What if we have an AM and PM session and want to assign a user to both? Since an itinerary is date bound only I'm not sure how we might want to handle the business rules"* The answer is as follows: If I understand it, we are currently only allowing one assignment per day is that right?  Either way that is okay for now and we will need to get clarification from folks like @OhMcGoo and @adedonato  if they do multiple sessions in a day

adedonato:
For Q2 "What if we have an AM and PM session and want to assign a user to both? Since an itinerary is date bound only I'm not sure how we might want to handle the business rules": We don't do multiple sessions in a day so only allowing one assignment per day will be fine.

tonysurma:
Thanks @adedonato

mgmccarthy:
@stevejgordon... above you say:
"In that case our Mediator handler will catch and prevent a double assignment, however there will be no feedback on the UI."

can we front this check in the controller right before sending the AddTeamMemberCommand? That way, we're not saddling our command handlers with validation logic, and it should be easy to report the double assignment scenario back to the UI.

stevejgordon:
@mgmccarthy I can see what you're saying and that makes sense. I think (casting my mind back) that I was expecting to do more with the data during the Add command so didn't want to query twice. Looking there now though I can't see any need for it to be there specifically. 

So it probably makes sense to have something like a bool CanUserBeAssignedToItinerary query which checks this and gets call from the controller first. Perhaps rather than a simple true/false we have a simple success/failure object with a message to explain more about the issue. They could already be assigned to that itinerary or another itinerary occurring on the same day, or perhaps they volunteered from the event.

PR #855, opened 2016-06-14
Finish test coverage for TaskStatusChangeHandlerAsync

rmcc13:
I believe this pull request finishes #787 but you may want to review it first.


AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Tasks/TaskStatusChangeHandlerTests.cs
```diff
@@ -3,102 +3,204 @@
...
-        protected override void LoadTestData()
-        {
-            var context = ServiceProvider.GetService<AllReadyContext>();
+    {
+
+        Mock<IMediator> mediator;
+        TaskStatusChangeHandlerAsync handler;
+
+        public TaskStatusChangeHandlerTests()
+            : base()
```
mgmccarthy: remove `:base()`

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Tasks/TaskStatusChangeHandlerTests.cs
```diff
@@ -96,41 +96,57 @@ protected override void LoadTestData()
...
-            // Make sure the task signup has the expected status
-            taskSignup.Status.ShouldBe(TaskStatus.Assigned.ToString());
+            // Make sure the task signup properties are the expected values
+            // based on the command.
+            taskSignup.Status.ShouldBe(command.TaskStatus.ToString());
+            taskSignup.Task.Id.ShouldBe(command.TaskId);
+            taskSignup.User.Id.ShouldBe(command.UserId);
+            taskSignup.StatusDescription.ShouldBe(command.TaskStatusDescription);
+            // the datetimes should be within 1 second of each other
+            taskSignup.StatusDateTimeUtc.ShouldBe(dateTime, TimeSpan.FromSeconds(1));
```
mgmccarthy: take a look at how DateTime is tested in `UpdateMyTasksHandlerAsync` and `UpdateMyTasksHandlerAsyncShould`.  We use a func to pass in what we need for the unit test, and at runtime, the func resolves itself. Up to you if you want to test dates this way. This is pretty much how I do it, but there is not "standard" way for the code for AllReady

mgmccarthy:
@rmcc13, I've made some individual annotations on different commit that are part of this PR. One thing you can do is remove the comments from the unit test code. You code is clear enough that other devs should understand what you're testing and why you're testing it.

Great job on getting full test coverage on this handler. There is a lot going on in there, and now we have test coverage.

tonysurma:
@rmcc13 let me know if you want to make any changes before I accept this PR?

mgmccarthy:
@tonysurma, we can always implement any of my suggestions later. What @rmcc13 wrote works, and it gets this handler to full test coverage so we can accept the PR as-is.

tonysurma:
okay I am accepting now.  thanks @rmcc13

tonysurma:
@rmcc13 and @mgmccarthy looks like we have a failing unit test after the merge.  Can youi take a peek?

https://ci.appveyor.com/project/HTBox/allready/build/1.0.0.1102.master/tests

mgmccarthy:
@tonysurma, back at work behind and impenetrable proxy... so I can't fix it now, but I know why it's failing. 

For a quick fix, just comment out any assertions make on the DateTIme stuff in the unit test that's failing. 

For a better fix, DateTime should be tested like it is in  UpdateMyTasksHandlerAsync and UpdateMyTasksHandlerAsyncShould

tonysurma:
appears transient as it is now working in the current build.  DateTime stuff is fun... not

rmcc13:
@tonysurma @mgmccarthy Hey guys, sorry I haven't been responding today, been at work all day. I kind of assumed the Github Android App would give me notifications, guess not. I agree with the base constructor, I can remove that and the comments. I guess I just don't understand what the benefit of having a delegate for returning a UTC DateTime. It looks like it is only added for the benefit of sharing a date time between the application code and unit test code for testing date times, I could be wrong, but if that's right, I don't think that is a good strategy, and could potentially be harmful. Even with code reviews and developers doing their due diligence, this leaves the possibility for someone to do in application code what is being done in the test code and now there's inaccurate data in the system.

I think a better strategy would be to come up with some agreeable threshold for unit testing, definitely something larger than what I chose (1 second) that caused the test to fail. We could write a helper function or extension methods and tell people to use that to compare dates. And if, later down the line, it is decided the threshold should be increased or decreased, you only have to change it in one place as long as people use that.

mgmccarthy:
@rmcc13, I disagree with establishing a threshold, b/c that threshold is outside of our control, no matter how amicable we make the threshold. The AppVeyor build machine can be under different loads at any time and we have no control over that.

I also disagree with the service based approach, as we already have access to `DateTime.Now` and `DateTime.Utc` now as programmers, which are themselves abstractions on how to grab those values at runtime. We want to avoid needless abstractions (or abstractions of abstractions)

However, I do agree that a delegate might not be the best way to go here. I am also open to a larger conversation about how to go about testing dates and times in unit tests. For instance, do we really need to unit test simple usages of DateTime.UtcNow?

Where I work we currently use the delegate-based approach, and I first discovered its usage while reading a blog Ayende wrote about the challenges of testing dates and times:
https://ayende.com/blog/3408/dealing-with-time-in-tests

For now, I'm going to remove the test for DateTime so we can keep moving forward here. @rmcc13, thanks for sharing your thoughts on this and for the great work on the unit tests for `TaskStatusChangeHandlerAsync`.

@BillWagner @MisterJames @stevejgordon any feedback about this?

Issue #840, opened 2016-06-09
Validate if new users are required to complete their name


stevejgordon:
This is required so we can display their name on itineraries and within the application.

@tonysurma - A couple of questions around this

1. Currently name is a single field. Would this be better if split to forename and surname?
2. Should we prevent task signups until someone has provided their name? A downside would be a slighter bigger barrier to entry for volunteers. An alternative is we promote completing their profile but don't' enforce it. We'd render empty names for those we don't have.

tonysurma:
1. Yes
2. I think we need to enforce name or we prompt but don't enforce 

@OhMcGoo, @adedonato what do you think?

adedonato:
1. Yes, I think splitting the name into forename and surname field would be best. 
2. I think we should enforce names.

tonysurma:
Thanks @adedonato !

stevejgordon:
Thanks @adedonato and @tonysurma - For a regular signup this should be quite simple as we can expose the name fields on the register form as required.

I need to double check the signup flow from Facebook etc where this may not always have this data available. I think the flow brings them to a page to complete their account so this could be quick simple. I need to setup my dev machine with Facebook to test that though.

Currently there is a warning for a user who does not have their profile completed in the top right, however a use can still volunteer without setting their name.

The profile warning also occurs if no phone number is set. This seems sensible since this is a key tool for contacting a volunteer during an event. Setting a phone number requires a verification of a code sent to the phone.

Two options
1. We include the forename, surname fields in the volunteer box, if they are not set in their profile. This could update their profile behind the scenes. This is fine for the name but less slick for the phone since we require it to be validated with the code.

2. We provide a warning that they must complete their profile first and a link to the manage profile screen? We then allow the normal management to take place and once complete they can volunteer. This is easier with the current code and structure as it requires less work to change. It might be slightly more painful for the volunteer though.

Followup:
Since we pretty much expect a phone number for a user, should we include that as required on the register form? Not a good idea if you could have volunteers without a mobile phone who may register. Although, in that case the management of them may breakdown later anyway?

tonysurma:
Could we do Option 1 and redirect to the validation screen if they provide a phone or is that too crazy?

Alternatively Option 2 could work but it feels 'less slick' than Option 1 and more painful as you said.

Lastly yes, mobile phone number should be required.  at this point that shouldn't be a huge impediment for near turn scenarios and we can always adapt/relax this in the future

stevejgordon:
[...]

PR #856, opened 2016-06-14
Split User "name" property

### LINKED ISSUE #846: Split User "name" property
We currently have a name property on the user. This should be broken up into Forename and Surname and any uses of name updated.

Any forms where the user can add their name will need to be adjusted also.

mgmccarthy:
Fixes #846


mgmccarthy:
@stevejgordon, I'm realizing now that I could have updated my newly created migration instead of creating three separate migrations as part of this refactoring (doh)... I'm still pretty new to EF migrations.

If there is a way for me to clean that up, I will. If not, I know for next time.

stevejgordon:
@mgmccarthy You can unwind a migration or migrations and then create a single new one for all changes. What you need to do is:

1. run a database update specifying a migration name of the migration prior to your changes. So in your case I think it would be dnx ef update database "RequestToEventMapping". This unwinds the changes on your database.
2. next you need to remove (once for each migration) the migrations you've added using dnx ef migrations remove. This updates your snapshot and deletes the migration. Your models remain unchanged though. After doing this three times you should have no migrations of your own.
3. You can then create a new migration which will compare your model to the snapshot and build a single migration to make all relevant changes.

That should work ok. Any questions, let me know.

It's nearly lunchtime for me and I was going to try and pull in your code to my local repo and do my first code review on it. I can wait though until you have any final tweaks added.

It's up to you but since they'll end up with a few commits, some of which override work done in the prior commit I usually soft reset my branch to before I started working and then create a new single commit that includes the final view. It's not totally necessary (and be careful if you do it, not hard reset as you'll loose work!) but I do prefer it since you have one commit showing all changes and it's easier for code review. Sometimes I split the functionality commit work and the testing work into it's own commit since that can make it really clear what actual functional files need checking.

stevejgordon:
@mgmccarthy I've managed to pull those down during lunch. I've not looked over in detail but initially all looks good. You have a todo in the ApplicationUser class asking if firstname and lastname should be required.

I think it's valid that at database level they are not required since we don't always have that data when creating the user from social login so we do need a record to be creatable without them. In the UI wherever we collect the data though I think required is correct based on feedback I had in #840

A think a result of #840 once my further points are answered we should be presenting these fields on the confirmation after social login and enforcing their completion.

I'll try and review more of this a little later.

tonysurma:
I commented on #840 thanks

mgmccarthy:
@tonysurma, my latest commit against this PR has been "building" now in AppVeyor for over a half hour... is it stuck?

tonysurma:
it has been restarted

mgmccarthy:
@stevejgordon when you get a chance, let me know if this looks good. Thanks for the help with the EF7 migration stuff.

stevejgordon:
Look good to me. Thanks Mike. 

If I were being picky you might want to change NotifyAdminForTaskSignupStatusChange and use your computed property Name instead of the FirstName and LastName parts in line 50?

Otherwise I think we can ship it

tonysurma:
Going to ship it and @mgmccarthy can add to it later.  Thanks for the contribution

mgmccarthy:
@stevejgordon @tonysurma I'll take care of that tonight when I get home.

Issue #874, opened 2016-06-16
Notify volunteer if assigned/unassigned to an itinerary


stevejgordon:
We should notify a volunteer if they are assigned or unassigned to an itinerary.

There are already some classes handling notifications within the current code so perhaps we can base this code off of those / reuse?

mgmccarthy:
@stevejgordon, if you want me to have a crack at this, I will. I'm assuming we want to know if we should reuse the current handling classes though... would that be something @MisterJames @BillWagner  or @tonysurma could give us guidance on?

tonysurma:
I would reuse and @MisterJames should be your contact for that

stevejgordon:
@mgmccarthy Yes please have a go at this one. I've not looked at the existing ones in detail myself but  can do if you need any advice as you dive into it.

mgmccarthy:
@stevejgordon, I'm assuming the place in the system where a voluteer is assigned or unassigned a to an itinerary is the IntineraryController's AddTeamMember action method?

stevejgordon:
Assignment occurs from AddTeamMember, that's correct. I haven't managed to put in a remove team member flow yet. It's on my list!

mgmccarthy:
@stevejgordon, after poking around a bit, I'm a little confused about what some of the variable names actually mean here. For example, we take a `selectedTeamMember` in the `AddTeamMember` action method on `ItineraryCotroller`, yet that gets set to a `TaskSignupId` on the `AddTeamMemberCommand`...

so how, in the context of `AddTeamMemberCommandHandlerAsync`, would I get a list of volunteers that have been assigned to an Itinerary? 

I see that the `TaskSignup` entity has a User property, so I'm assuming I would send to one volunteer at a time? In other works, the `AddTeamMemberCommandHandlerAsync` handler's transaction is scoped to adding one volunteer(aka, "user") at a time to an Itinerary?

sorry about the questions, I don't fully yet have a firm grip on the whole domain, especially the newly added Itinerary stuff

mgmccarthy:
@tonysurma any idea of what the subject/body of this SMS/Email notification should contain?

stevejgordon:
@mgmccarthy No worries on the questions at all. This piece of functionality crosses a few areas so it's not super clear what's happening.

[...]

PR #638, opened 2016-03-30
refactor ActivityDetailForNotificationQueryHandler to async

mgmccarthy:
Fixes #637

PR #892, opened 2016-06-20
Notify volunteer when assigned or unassigned to an intinerary

mgmccarthy:
My first commit for Issue #874. Check the "to-do" list there for completion as this PR will be one or many against Issue #874


AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Organizations/OrganizationIdQuery.cs
```diff
@@ -1,10 +0,0 @@
-﻿using AllReady.Models;
-using MediatR;
-
-namespace AllReady.Areas.Admin.Features.Organizations
-{
-    public class OrganizationIdQuery : IAsyncRequest<int>
-    {
-        public int? ItineraryId { get; set; }
```
stevejgordon: @mgmccarthy I should probably explain what my intentions where for this query and handler as it looks like you've adjusted/renamed? I was thinking we could extend the query message to have nullable ints for all main lookups (itineraryid, eventid etc) and then a single handler which queries based on the id with a value. Thinking it through, separate specific queries and handlers might be slightly simplier and limited to single responsibility, although we could end up with many very repetitive handler, which if a domain model change occurs; all would need updating.

My intension with the handler was to get to a single initial query and just applying the appropriate where clause in the handler as required. I didn't get that far with it to date.

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public VolunteerAssignedToItineraryHandler(AllReadyContext context, IMediator mediator)
+        {
+            _context = context;
+            _mediator = mediator;
+        }
+
+        public async Task Handle(VolunteerAssignedToItinerary notification)
+        {
+            var taskSignup = await _context.TaskSignups.SingleAsync(ts => ts.Id == notification.TaskSignupId).ConfigureAwait(false);
```
stevejgordon: You might need to do an include of the User here to ensure the related data is loaded by EF.

I've managed to pull in your PR but for some reason the PreferredEmail and SMS is not loading when I debug through. I can't see an obvious reason why not but will look a bit deeper later on.

stevejgordon:
Looking good @mgmccarthy! The query for the task signup is not returning the expected data on my local test at the moment, but I'm not entirely sure why. I'll need to spend more time with it later. The approach looks good though. Some specific commends added above.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Organizations/OrganizationIdQuery.cs
mgmccarthy: @stevejgordon I can revert for now and and name as you originally did. The renaming of the message/handler was b/c the the original name of OrganizationIdQuery was confusing to me at first.Now, when I look at my changes, the initial message/handler name is fine, and my renaming doesn't really make sense. I don't know what i was thinking :confused: 

As far as extending the query message to use nullable ints for all the main look-ups, it would help in re-usability, but could be very confusing to the developer trying to utilize the message. A command should have one logical handler in the system. Although this design still adheres to that, the developer still needs to look at the conditional code in the handler to understand the ramifications of including or dis-cluding values from the properties on a given message. This obfuscates the intent of the message and violates SRP in the handler.

IMO, I'd rather work with/write unit tests for three smaller message/handler classes that all do almost the same thing, then one larger handler that I need to figure out what property on that command should be nullable based on which execution path I want to run in the handler.

Either way, I'll let you make this decision, b/c you understand these UC's at a higher level than I do, and understand how they work with the rest of the system better than I do.

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
mgmccarthy: @stevejgordon, thanks for the feedback. I still couldn't set up my system to run through an entire smoke test of this handler... I could only get so far, but I still decided to submit the PR so I could share my code. I'll keep trying to do a fresh setup following your instructions (it's probably something I'm missing on my end) so I can get an end to end test on this.

Did you want me to include User and update the PR or do you want to have a crack at it first?

stevejgordon: Probably should get the User include added as I'm fairly sure we'll need that. I'll have to try and work out why on my run through I'm not seeing all of the data.

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+            _mediator = mediator;
+        }
+
+        public async Task Handle(VolunteerAssignedToItinerary notification)
+        {
+            var taskSignup = await _context.TaskSignups
+                .AsNoTracking()
+                .Include(x => x.User)
+                .SingleAsync(ts => ts.Id == notification.TaskSignupId)
+                .ConfigureAwait(false);
```
dpaquette: Why ConfigureAwait(false)? I don't think this is necessary.

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+            if (!string.IsNullOrWhiteSpace(emailAddress))
+            {
+                notifyVolunteersViewModel.Subject = subject;
+                notifyVolunteersViewModel.EmailRecipients.Add(emailAddress);
+                notifyVolunteersViewModel.EmailMessage = message;
+                notifyVolunteersViewModel.HtmlMessage = message;
+            }
+
+            await _mediator.SendAsync(new NotifyVolunteersCommand { ViewModel = notifyVolunteersViewModel }).ConfigureAwait(false);
```
dpaquette: Again, ConfigureAwait(false) as mentioned above.

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
mgmccarthy: @dpaquette, it is my understanding when you don't need the original calling context back after invoking an await call, you should explicitly set ConfigureAwait(false) for best performance.

An example of where you would not want to use ConfigureAwait(false) is in a UI thread, that would need to pick it's context back up again after the awaited call is finished to it can return a result to the UI. 

For example. I would not want to use .ConfigureAwait(false) in an MVC controller action.

Sean Feldman gave me some guidance early on in this project with using .ConfigureAwait(false) in handler code. You can see some of his feedback here on one of my PR's. He corrects me to use .ConfigureAwait(false) both on async invocations of the mediator and db context:
https://github.com/HTBox/allReady/pull/638#issuecomment-204093377

That being said:
1. I might not have interpreted Sean's guidance correctly.
2. with .NET Core, there might be a better/different way to do this
3. I don't understand the intricacies of ConfigureAwait yet, and I should not be using it until I understand what I'm doing ;)

dpaquette: It seems I'm with you on point 3. Time to do some more research.

tonysurma:
responded to @mgmccarthy on the issue regarding text for email/sms

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
mgmccarthy: @dpaquette, you mean that I'm using ConfigureAwait incorrectly?

dpaquette: No, if @SeanFeldman told you to do it this way then you're doing it correctly. I'm just saying I don't fully understand it and need to do some research.

mgmccarthy: @dpaquette, here are a couple links that helped me to get my head around it:
- http://stackoverflow.com/questions/13489065/best-practice-to-call-configureawait-for-all-server-side-code
- http://blog.stephencleary.com/2012/02/async-and-await.html
- http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

But Sean is really the one who helped me understand it best. He basically said (and I'm paraphrasing): 
"on any invocations of an async method that you await, if the place the method is being invoked from is NOT a UI-bound thread (MVC/WPF/Wndows Forms, etc...), use .ConfigureAwait(false)"

SeanFeldman:
@dpaquette @mgmccarthy it's all about the questions if you care / need to get to the same context you've started from or not. 

MSFT had to make a choice, either make it by default return to the same context (aka `.ConfigureAwait(true)`) or not (aka `.ConfigureAwait(false)`). With the first option as the default, the user code that has anything to do with UI can run as-is w/o `.ConfigureAwait(true)` sprinkled all over the code, but make library developers unhappy. In case the second options would be made the default, library developers would be happy, but UI devs would swear on a daily basis. Microsoft decided that it's better not to piss off UI developers (WPF/ASP), so library devs are now "blessing" `.ConfigureAwait(false)` every day.

That's a bootleg version of the story :wink:

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+            var taskSignup = await _context.TaskSignups
+                .AsNoTracking()
+                .Include(x => x.User)
+                .SingleAsync(ts => ts.Id == notification.TaskSignupId)
+                .ConfigureAwait(false);
+
+            var emailAddress = !string.IsNullOrWhiteSpace(taskSignup.PreferredEmail) ? taskSignup.PreferredEmail : taskSignup.User?.Email;
+            var sms = !string.IsNullOrWhiteSpace(taskSignup.PreferredPhoneNumber) ? taskSignup.PreferredPhoneNumber : taskSignup.User?.PhoneNumber;
+
+            if (string.IsNullOrWhiteSpace(emailAddress) && string.IsNullOrWhiteSpace(sms)) return;
```
SeanFeldman: Would it reveal the intention better if `TaskSignup` would expose `.HasAMethodOfCommunication`?
Then the logic of determining if email address & sms along with validations would be encapsulated in the `TaskSignup`.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ItineraryController.cs
```diff
@@ -128,7 +127,7 @@ public ItineraryController(IMediator mediator, IItineraryEditModelValidator itin
         [Route("Admin/Itinerary/{id}/[Action]")]
         public async Task<IActionResult> SelectRequests(int id)
         {
-            var orgId = await _mediator.SendAsync(new OrganizationIdQuery { ItineraryId = id });
+            var orgId = await GetOrganizationIdBy(id);
```
SeanFeldman: Noticed that getting organization ID and verifying that it's a valid one or that the user can administer the given organization appears multiple times in several actions. Question: would it justify to convert that code into custom action filter?

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Itineraries/AddTeamMemberCommandHandlerAsync.cs
```diff
@@ -21,11 +22,8 @@ public AddTeamMemberCommandHandlerAsync(AllReadyContext context, IMediator media
         {
             var itinerary = await _context.Itineraries
                 .Where(x => x.Id == message.ItineraryId)
-                .Select(x => new
-                {
-                    EventId = x.EventId,
-                    Date = x.Date
-                }).SingleOrDefaultAsync();
+                .Select(x => new { x.EventId, x.Date })
+                .SingleOrDefaultAsync();
```
SeanFeldman: `.ConfigureAwait(false)` :smiling_imp:

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Organizations/OrganizationIdQueryHandlerAsync.cs
```diff
@@ -12,23 +12,16 @@ public class OrganizationIdQueryHandlerAsync : IAsyncRequestHandler<Organization
...
-            }
-            else
-            {
-                return 0;
-            }
+            return await _context.Itineraries.AsNoTracking()
+                .Include(i => i.Event.Campaign)
+                .Where(i => i.Id == message.ItineraryId)
+                .Select(i => i.Event.Campaign.ManagingOrganizationId)
+                .FirstOrDefaultAsync();
```
SeanFeldman: `.ConfigureAwait(false)`

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Itineraries/AddTeamMemberCommandHandlerAsync.cs
```diff
@@ -58,8 +56,10 @@ public AddTeamMemberCommandHandlerAsync(AllReadyContext context, IMediator media
                 _context.TaskSignups.Attach(taskSignup);
                 _context.Entry(taskSignup).Property(x => x.ItineraryId).IsModified = true;
                 await _context.SaveChangesAsync();
+
```
SeanFeldman: `.ConfigureAwait(false)` for `await _context.SaveChangesAsync();` and `_mediator.SendAsync`

dpaquette:
@SeanFeldman thanks for quick lesson on `ConfigureAwait()`. I'm going to pick your brain about this a little more when I see on you Thursday.

SeanFeldman:
@dpaquette will be happy to answer if can... (don't trust @SeanFeldman`s answers, he can bluff his way through).
 
This can perhaps help a little with a mixed project like this, where UI and library code are mixed together.  https://github.com/distantcam/ConfigureAwait with some AOP magic can get the pesky `ConfigureAwait` out of the way.

AllReadyApp/Web-App/AllReady/Features/Notifications/VolunteerAssignedToItineraryHandler.cs
mgmccarthy: @SeanFeldman, yes it would. As a matter of fact, after this soft release to June to the Red Cross, there is an item on the backlog that would encapsulate this logic for all messages, as well as let users set their communication preferences (SMS/emai, etc...) so we're not shuttling around messages that expose all communication types as properties and put the work on the consumer to look up and populate what needs to be sent.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ItineraryController.cs
mgmccarthy: @SeanFeldman, not only in sever actions (which I've now encapsulated in a private method with latest version of code, my PR still waiting to be merged), but it happens across several controllers.

ActionFilter might make sense here b/c of usage across controllers.

mgmccarthy:
@tonysurma @MisterJames @stevejgordon @BillWagner 

I think Fody looks like the way to go here, especially with our push towards using MediatR handlers. This could greatly simply making sure we're optimizing our async/await calls on non-UI threads.

Also can help, b/c it's very easy to forget to put .ConfigureAwait(false) on all code.

tonysurma:
@mgmccarthy I will take lead from the others like you, @stevejgordon @MisterJames @BillWagner @dpaquette on use of the library.  Thanks.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ItineraryController.cs
tonysurma: @mgmccarthy you mention a PR still waiting to be merged but I don't see one...

mgmccarthy: @tonysurma PR #892. It should NOT be merged yet, b/c I have more work to do... unless you want multiple PR's against the Issue #874. I just pushed some .ConfigureAwait(false) changes to this PR that sean feldman found below.

Let me know if it's easier for you to have it broken up into separate PR's for this issue

tonysurma: No this is fine.  You had just said _"(which I've now encapsulated in a private method with latest version of code, my PR still waiting to be merged)"_ and I thought you were referring to a different PR.  Apologies for my confusion

tonysurma:
per issue discussion going to merge this while @mgmccarthy continues to work on notifications functionality and @stevejgordon on itinerary items

PR #898, opened 2016-06-22
Adding a very basic volunteer dashboard

### LINKED ISSUE #376: Create "My Events" Page For Volunteers
Create a listing of events for which the logged-in user has volunteered for.

From #373.

stevejgordon:
Closes #376 

This is a very basic initial volunteer dashboard. I'm committing in it current basic state so that we can review what this page really needs to do and break out small issues to extend and enhance it as required.

NOTE: There are no tests currently but the basic mechanics are there.

image

Added a dashboard link for logged in users which takes them to this listing of events (where they have volunteered for tasks). Does a basic breakout of the events in past, current and future.

@tonysurma I think this is enough for now. I've done basic manual testing of it but will spend more time on it in the near future.

cc/ @mgmccarthy - As discussed earlier, this is what I had sitting around. Better to get something committed I think so we can see it and then refine as we move forward.


mgmccarthy:
@stevejgordon is on :fire: !!!

send in the :fire_engine:

thanks for committing early Steve. Having something basic to look at will help in the conversation going forward.

PR #899, opened 2016-06-23
Unassign request from an itinerary

### LINKED ISSUE #877: Ability to unassign requests
We need to be able to unassign a request from the the Itinerary management page.

This will remove the assignment mapping and change the status.

In a future feature this should also update the itinerary / request history.

stevejgordon:
Added basic v1 functionality to allow requests to be unassigned from an itinerary.

@tonysurma Need to review the business rules as we may need to further enhance this. At the moment unless a request is marked completed I allow it to be removed from an itinerary. We may want to prevent removal if the itinerary date has passed but I'm not sure if that would be correct in all use cases.

Closes #877 

NOTE: No tests in the interests of getting functionality into the site for testing


tonysurma:
For now those rules are good, thx

PR #923, opened 2016-06-23
Changing default end date for event to match start date

### LINKED ISSUE #918: Create event
In create event, during create second date box should default to first date box date

stevejgordon:
Closes #918

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventAdminController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventAdminController.cs
@@ -79,7 +79,7 @@ namespace AllReady.Areas.Admin.Controllers
                 OrganizationId = campaign.OrganizationId,
                 OrganizationName = campaign.OrganizationName,
                 StartDateTime = DateTime.Today.Date,
-                EndDateTime = DateTime.Today.Date.AddMonths(1)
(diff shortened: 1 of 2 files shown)


tonysurma:
marked as WIP in hopes steve/someone can do the other part I added to issue

stevejgordon:
@tonysurma Updated per feedback. Currently on loosing focus from the startdate box I have set the end date to the same value.

Is having a time against an event still valid our would just selecting the date be enough? With tasks then bounding the specific timings for things to occur within that date range? That might be a feature story if it needs to change but thoughts I'd ask.

Currently the field is set as date time so the time element is editable via the browser implementation of the date/time control.

tonysurma:
events need dates and times and yes in tasks should have to fall within those bounds generally - but it gets funky as different folks do different things for tasks like setup and teardown (do you put event 9a-5p and setup 8a-9a or do you then make the event 8a-5p even tho only one tasks goes outside of the 'published' times for the event for example)

VishalMadhvani:
Not sure if event can span multiple days but if not, then Event StartDateTime and EndDateTime can be replaced with EventDate, StartTime and EndTime. This makes the implementation more explicit.

For tasks, the editor can use the EventData, StartTime and EndTime as a default, but the validation should be relaxed so tasks are not restricted to be inside the Event Start and End time. Hence allowing setup or teardown tasks.

PR #925, opened 2016-06-23
Hiding 'My assigned tasks' section when no assigned tasks

### LINKED ISSUE #924: My assigned tasks
On the event details page only show the "My assigned tasks" section if the user actually has some tasks assigned.

image

In this case it should be hidden.

stevejgordon:
Closes #924


stevejgordon:
This may need a rebase after some of the other prior PRs

tonysurma:
@stevejgordon looks like it needs no rebase but is failing build? (I will research more later)

tonysurma:
rerunning the build

PR #945, opened 2016-06-24
Validate if new users are required to complete their name

mgmccarthy:
First PR for #840

I still need to add sending the phone number verification message for signing up via regular AllReady user registration.

For external auth providers, I've only tested with Facebook. Google and Twitter have yet to be tested


mgmccarthy:
@tonysurma, just pushed update to this PR that sends the sms confirm message as part of the AllReady user registration process. The user should not receive the sms code on the phone number they provide at registration and be able to use that code via the UI to verify their phone number.

tonysurma:
I presume that was a typo and it was 'user should NOW receive...' right :)

Thanks for adding that, I will merge in a bit.

mgmccarthy:
@tonysurma, you are correct and I fixed that typo.

Issue #891, opened 2016-06-20
Gaurd clauses in class constructors... do we really need them? (low priority)


mgmccarthy:
Looks like some of our code uses guard clauses for dependencies passed in to our constructors. I don't really see the benefit of this extra code. 

If an instance of the dependency is not provided at runtime, we'll receive and exception anyhow that should be descriptive enough that would allow us to figure out what is wrong.

Same thing for unit testing, if we're writing a test and pass in a null value for a dependency that is not used for that particular test, then we're going to get an exception thrown. But it the particular test does not call any of the items on the contract of the dependency, we should allow a null to be passed.

Thoughts?

dpaquette:
@mgmccarthy can you link to an example of this?

mgmccarthy:
@dpaquette  I've seen it here most recently b/c I'm helping @stevejgordon with a messaging handler that is going to be invoked from a command sent from this controller:
https://github.com/HTBox/allReady/blob/d0923120d66a3711796c2fe73355ee4b705a1465/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/ItineraryController.cs

but I've also seen them elsewhere in controller, ViewModel and sometimes extension code.

just to be clear, I'm not saying one way is right on wrong here, just bringing it up for conversation. If it's not a big deal either way, that's fine.

dpaquette:
I would agree that these are probably not necessary

stevejgordon:
I've personally tended towards checking for the nulls / out of expected range on parameters explicitly on public methods to be sure that nothing within the method may occur before we reach the null and end up with something in a bad state. Happy to go with the majority as it may be overkill in some of the cases.

tonysurma:
pinging @MisterJames and @BillWagner for thoughts as well

BillWagner:
I'm generally a fan of "failing fast" to help developers using an API find their mistakes quickly. I think that leads to adding the guard clauses. (As @mgmccarthy points out, sometimes it's just extra code with no benefit.)

PR #947, opened 2016-06-24
Implemented stubs for CampaignAdminControllerTests

auroraocciduusadmin:


tonysurma:
Thanks @auroraocciduusadmin for your contributions again!

@mgmccarthy can you take a peek at these before I merge?

mgmccarthy:
@tonysurma, I might not be able to get to it until Sunday or Monday.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -82,29 +88,63 @@ public void IndexReturnsCorrectDataWhenUserIsOrgAdmin()
...
+            mockMediator.Setup(x => x.Send(It.Is<CampaignListQuery>(c => c.OrganizationId == null)))
+                .Returns((CampaignListQuery q) => {
+                    // return some models 
+                    List<CampaignSummaryModel> ret = new List<CampaignSummaryModel>();
+                    ret.Add(new CampaignSummaryModel { OrganizationId = OrganizationId });
+                    ret.Add(new CampaignSummaryModel { OrganizationId = OrganizationId + 1 });
+                    return ret;
+                }
+            );
+            var mockImageService = new Mock<IImageService>();
```
MisterJames: When you're testing a method that doesn't use the dependency (like IImageService) you don't have to mock it, you can just pass in null. It makes the test a little lighter, too.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -82,29 +88,63 @@ public void IndexReturnsCorrectDataWhenUserIsOrgAdmin()
...
+            {
+                new Claim(AllReady.Security.ClaimTypes.UserType, UserType.SiteAdmin.ToString()),
+            };
+            controller.SetClaims(claims);
+
+            // All campaigns returned when not OrgAdmin
+            ViewResult view = (ViewResult)controller.Index();
+
+            // verify the fetch was called
+            mockMediator.Verify(mock => mock.Send(It.Is<CampaignListQuery>(c => c.OrganizationId == null)));
```
MisterJames: I think this is good as far as unit tests go. In fact, I would argue that for the controller that if the ViewModel is of the expected type and you've invoked the bus with the correct parameters, then the other assertions (around count) are superfluous. ...and you may _just_ be testing your test.

MisterJames:
@mgmccarthy I started reviewing this one but will wait for your feedback as well.

@auroraocciduusadmin Thanks for your contributions! Your tests are asserting what they should be, and perhaps a little more. In some respects they may be brittle as unit tests, but parts could be extracted as integration tests. That said, I see throughout the project that other tests are going to this level as well, so what you're doing is consistent. We'll likely merge, but we should also be giving you proper guidance here, too. Thanks again for the heavy lifting!

mgmccarthy:
@MisterJames, I agree with all of your feedback. I am watching my 8 month old today solo (my wife's at work), so don't really have that much time to contribute today :baby: 

One thing I think is important to get alignment with the project owners (@tonysurma @BillWagner ) is the fact that we should be passing null for unused dependencies in the constructor of classes instead of new'ing up a mock for unit testing.

One mock is not expensive, but after you clear 10,000 unit tests on classes with multiple dependencies, then 1/4 of the mocks you're creating are for nothing, and you're going to slow down your builds.

This links back to an Issue #891, which is something I brought up about guard clauses. @dpaquette @stevejgordon and @BillWagner all weighed in on the topic and it appears there is no consensus as to whether or not we should be using guard clauses in class constructor's for injected dependencies.

The apporach to unit testing your suggesting (which I agree with), is to pass in null for unneeded dependencies for a given unit test, but, if we start putting guard clauses for injected dependencies looking for null and throwing exceptions, we won't be able to approach unit testing this way.

We'll always need to supply a mock for each injected dependency.

Thoughts?

tonysurma:
Thanks all, I will merge this in - as its good (thank you @auroraocciduusadmin) and the thoughts on unit testing and guards will still be here as well as in the other issue for discussion.

auroraocciduusadmin:
@MisterJames  re "testing the test": 
I often feel that I'm testing the test when using mocks. For example, a test like Campaign.DetailsReturnsCorrectViewModelWhenViewModelIsNotNullAndUserIsOrgAdmin would be expected to ensure that, when passed an ID for a Campaign, we get a view that is passed a CampaignDetailModel with the correct data for that campaign (and user type). But we need a way to have the method return a viewmodel so we can see that it's working correctly. So we look under the hood of CampaignController.Details() and see that it's getting the data from CampaignDetailQueryHandler. So in our test we build a controller, mock the handler, and make sure that the handler gets called in the test. But how do we know that the "real" handler that got mocked is working? Well, we have tests for CampaignDetailQueryHandler so everything's good. But now we're not really testing the "contract" for CampaignController.Details() which is that we give it a campaign id for a known campaign and get the right CampaignDetailModel passed to a view. But isn't that the same thing? Not really. I shouldn't really need to know how the controller is working in order to test it...I just need to know what it expects as input and what it's expected to return (and expected side-effects). We could pass a CampaignDetailQuery in the CampaignController test and have it pass the test. We could also pass a CampaignDetailQuery to a CampaignDetailQueryHandler test and have it pass that test and (perhaps wrongly) conclude that the controller is working properly...espcially if the two tests are passed different parameters. The only way to really exercise the controller would be to use a "real" Mediator, in-memory database, and the like. So that's my case for "overkill" in unit tests. 
Maybe there's a way to write more "expensive" tests that won't get run after each build (ie. on demand)?

PR #952, opened 2016-06-25
Adding Privacy policy external url

### LINKED ISSUE #908: Add org privacy policy link field
We should include a field which allows a privacy policy link to be provided for an organization.

We should validate that either the link or the privacy policy content are provided when saving an organization.

On the public pages we should render the link if available or a link to the custom privacy policy page on allReady instead

stevejgordon:
Closes #908

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/OrganizationControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/OrganizationControllerTests.cs
@@ -1,6 +1,7 @@
 ﻿using AllReady.Areas.Admin.Controllers;
 using AllReady.Areas.Admin.Features.Organizations;
 using AllReady.Areas.Admin.Models;
+using AllReady.Areas.Admin.Models.Validators;
 using AllReady.Models;
 using MediatR;
 using Microsoft.AspNet.Authorization;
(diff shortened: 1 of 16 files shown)


mgmccarthy:
@stevejgordon, I am in complete alignment with the injectable model validators. They adhere to SRP and are really useful when more than one controller action calls into a private method which tests multiple paths of execution (the injectable validator approach is still a solid way to go even if it's only used by one controller action).

tonysurma:
was going to merge this but had a question, I have since merged a PR with another migration.  Does this still work or if this migration was 'before' that one will it break or not know which to run first?

If you get a moment @stevejgordon can you rebase and see if this works with the PRs I pulled in since you made this?  That will solve a question I have to see what the flow for PRs with migrations needs to be.

Thanks.

MisterJames:
@tonysurma I don't think this should be a problem because the model changes aren't overlapping, but I'll leave it to @stevejgordon to confirm he doesn't want to do a reordering. I'm wondering if the model state would be a problem (there is the one prop on Organization). Steve, if you can't get to it today I'll pull down the PR and give'er a go.

stevejgordon:
@MisterJames Sorry, having a bit of a hectic week. Not sure how the snapshot will behave in cases like this. I will try to revert my migration later and build a new one after rebasing from current master.

tonysurma:
I don't think you need to revert and redo  the  migration and the rebase will have no conflicts 

It is just more an experiment to learn if anything goes wrong for future

If i get  some  time I will try it locally and see what happens

VishalMadhvani:
This might help: https://msdn.microsoft.com/en-us/data/dn481501.aspx

But I think the general gist is if a migration is added to the middle on the pack, anyone that is on the latest database will need to drop and recreate.

On dev environments, we i think we should use the Drop and Recreate database on model change.
Because we have test data being seeded, any data added to the database outside of the seeding code will be lost.
I guess the side effect of this is the seed data will be forced to be continually improved.

mgmccarthy:
we should probably revisit the seeded data and simplify it. There is a lot overhead dragging all the of seeded data forward. Most of it, I never use, and if I do use any of it, it's just usually the first Campaign I see on my screen that I click on.

stevejgordon:
@tonysurma Looks good. I pulled in the latest master code, rebuilt my db, opened this branch, rebased off master (no conflicts) and run database update with no issue. So it looks compatible as far as I can tell.

Tests all pass and the site loads to a basic test of the privacy policy url

Issue #764, opened 2016-05-18
Remove map link and add in logo/image of campaign on home page


MisterJames:
When displaying the list of campaigns, display the image. The "map" link can be removed from the view. Lifted from #691.

digitaldrummerj:
@tonysurma can this issue be closed?  looks like fix has been merged for this issue.

PR #958, opened 2016-06-25
#764 remove map link and show campaign image on home page

VishalMadhvani:
PR for Issue #764


VishalMadhvani:
No restrictions on image size so the user is expected to provide an appropriately sized image.

Issue #967, opened 2016-06-27
SendResetPasswordEmailHandler and SendResetPasswordEmailHandlerAsync do the same thing


mgmccarthy:
I'm currently working on this. Please assign to me

@tonysurma @MisterJames @BillWagner. Can we get a decision on the "verbiage" here and I can refactor out the duplicate code?

`SendResetPasswordEmailHandlerAsync` handles `SendResetPasswordEmail` via this this code:
`await emailSender.SendEmailAsync(message.Email, "Reset allReady Password",  $"Please reset your allReady password by clicking here: <a href=\"{message.CallbackUrl}\">link</a>")`

`SendResetPasswordEmailHandler` handles `SendResetPasswordEmail` via this code:
`await emailSender.SendEmailAsync(message.Email, "Reset Password", $"Please reset your password by clicking here: <a href=\"{message.CallbackUrl}\">link</a>");`

this is the **same exact message** with the difference being `SendResetPasswordEmailHandlerAsync` adds the word "allReady" to the message. In both cases, the objective of sending the message is the same, to reset your password.

not only that, but we have two handlers handling the same request (which is a no-no, b/c "commands" should have only one logical endpoint).

we should centralize the "reset password" functionality to use one command/handler.

mgmccarthy:
I'm currently working on this. @tonysurma, can you please assign to me?

mgmccarthy:
@tonysurma this has been fixed and can be closed

Issue #948, opened 2016-06-24
Get External Auth working with Twitter


mgmccarthy:
@tonysurma please assign to me.

I'm currently working on this.

MisterJames:
closed via #969

PR #969, opened 2016-06-27
Send Email and Sms confirmation messages when authenticating using external providers

### LINKED ISSUE #840: Validate if new users are required to complete their name

### LINKED ISSUE #955: Send Email and Sms confirmation messages when authenticating using external providers
@tonysurma, please assign to me.

this is an Issue that grew out of #840, I've moved my final comment from the last thread on #840 and broke it out into this issue.

@tonysurma, just to clarify the functionality of what was merged in #945. Currently, when authenticating via Facebook, we are NOT sending out a message to confirm email or confirm phone. For regular AllReady registration, we're sending out both messages at the point the user registers.

For Facebook auth, after authenticating, the First, Last, and Email fields are all prepopulated, and the user only has to fill out Phone:
image

after they click register, we redirect them to the home page, where they can click on their manage account link and this is the page they will see:
image

I wanted to outline this for you, b/c regardless of which external auth provider we use, the user still lands on this page after authenticating with the external auth provider:
image

and they're still required to fill out all four fields in order to continue registration with AllReady, **regardless of what information is made available to us for the given auth provider**. Again, Facebook gives us more info to pre-populate those fields, and Twitter gives us hardly anything (at least from what I've been able to figure out so far).

**I'm thinking since all auth providers use this form page, and since all four fields are required, we should still be able to send an email and sms confirmation message when they hit register.**

mgmccarthy:
First, lousy name for my local branch name. Ignore it.

Fixes #840, #955, #976, #950, #949, #948

High level of what's included in this fix.

AccountController.ExternalLoginConfirmation now:

- sends email confirmation
- sends sms confirmation
- adds Security.ClaimTypes.ProfileIncomplete to user
- all three of the above steps were already been done in AccountController.Register

AccountController.ForgotPassword

- added new AllReady.Features.Manage.SendResetPassword message/handler
- ForgotPassword now sends this message instead of invoking email sender
- this is temporary code that can be consolidated once a decision is made on #967
- all unit tests have been updated to reflect changes and stubs have been added for additional code added to ExternalLoginConfirmation

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
@@ -1165,21 +1165,35 @@ namespace AllReady.UnitTest.Controllers
         }
 
         [Fact(Skip = "NotImplemented")]
-        public async Task ExternalLoginCallbackPopulatesCorrectValuesForFirstNameAndLastNameFromNameClaim__WhenNameClaimIsNotNull_AndExternalLoginInfoIsNotNull()
+        public async Task ExternalLoginCallbackPopulatesCorrectValuesForFirstNameAndLastName_WhenLoginProviderIsGoogle_AndExternalLoginInfoIsNotNull()
         {
             //delete this line when starting work on this unit test
             await taskFromResultZero;
         }
 
         [Fact(Skip = "NotImplemented")]
-        public async Task ExternalLoginCallbackPopulatesStringEmptyForFirstNameAndLastNameFromNameClaim__WhenNameClaimIsNull_AndExternalLoginInfoIsNotNull()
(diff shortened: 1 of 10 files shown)


tonysurma:
@dpaquette @BillWagner @MisterJames can I get a second set of eyes on this please?

tonysurma:
just moved to RTM so will need to be rebased and retested, also @MisterJames can you take a peek?

mgmccarthy:
@tonysurma and @misterjames I'll probably get to the rebase Sunday night.

mgmccarthy:
@tonysurma and @misterjames. rebased and ready for your review.

Issue #982, opened 2016-06-30
Resource Based Authorization


stevejgordon:
I spent a bit of time yesterday reading up on Resource Based Authorization which is part of the ASP.NET Core authorization stack.

In short this provides a standardised way to authenticate that a user has permission to perform an action on a specific resource. Within allReady we do this often to ensure that admin users have permission to modify only the objects which belong to their organization. E.g. An org admin can only view/edit an event if it belongs to their organization.

We have a useful extension method on the ClaimsPrinciple which checks if a user can access an object based on the OrganizationId. At the moment we have to rely on this method being called by most of the admin actions, in some cases having to handle code to first query for the OrganizationId based on an event, itinerary or task.

Resource based authorization feels like a good approach to clean the code a little and start to rely on the underlying authorization design within ASP.NET Core. What it means is that we have a few authorization handlers than can accept an object and then perform the relavant checks to ensure the user is authorized. These handlers are regsitered within the DI container and ASP.NET Core will check for appropriate handlers when we call the _authorizationService.AuthorizeAsync method. We just need to inject the IAuthorizationService into our admin controllers.

This feels like a sensible approach as in some cases it will simplify our code. We can move the logic that queries the db (where it needs to) to get an org Id and utilise the existing plumbing to manage the authorization. We can write tests around the auth handlers so that once they are defined we can ensure that they work as expected. Our logic lives in one place and has a single responsibility. Our controller actions will be slightly simplified in some cases as we won't need to use the Mediator to query for the org id. The controller should not really own that activity.

Also I think it should simplify our controller testing as well since I can't see that we'd need to mock up the User claims. We can mock the IAuthorizationService and verify it a) gets called where it should and b) mock the return from AuthorizeAsync to validate the success and failure behaviours.

There is a bit of a caveat for RC1 code in that the handler requires an override on the sync Handle method as well as the async one. This has been fixed in RTM I believe.

I spent a bit of time to put in an example PR where I have created some of the structure for using resource based auth. I have utilised this on a few sample actions to show how it would fit together. I'll submit the PR for easier discussion and review of this issue.

Ultimately, if it's agreed this feels like the proper approach I can flesh out my code a little more and we can begin to implement it on new actions, plus over time tidy up the existing actions.

Please let me know your thoughts.

See https://docs.asp.net/en/latest/security/authorization/resourcebased.html for more info.

mgmccarthy:
[...]

PR #983, opened 2016-06-30
WIP: Initial sample of using resource based authorization

stevejgordon:
Example relating to #982 

NOTE: Do not merge this currently as it's incomplete and just an example at this stage. It would need a little tidy up and tests need to be refactored to support what I've changed. For this example I've just commented out the tests that broke until we can review the concept.

Documentation reference - https://docs.asp.net/en/latest/security/authorization/resourcebased.html

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Security/AuthoizationHandlers/OrganizationOperations.cs
+++ b/AllReadyApp/Web-App/AllReady/Security/AuthoizationHandlers/OrganizationOperations.cs
@@ -1,0 +1,9 @@
+﻿using Microsoft.AspNet.Authorization.Infrastructure;
+
+namespace AllReady.Security.AuthoizationHandlers
+{
+    public static class OrganizationOperations
+    {
+        public static OperationAuthorizationRequirement Manage = new OperationAuthorizationRequirement { Name = "Manage" };
+    }
+}
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/EventAdminControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/EventAdminControllerTests.cs
@@ -30,17 +30,17 @@ namespace AllReady.UnitTest.Areas.Admin.Controllers
             await TaskFromResultZero;
         }
 
-        [Fact]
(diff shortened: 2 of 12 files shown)


MisterJames:
@stevejgordon let's chat about this after the standup next week (Aug 20) to figure out where it lands.

tonysurma:
@MisterJames @stevejgordon thoughts on this?  (trying to get PRs cleaned up or out or back on track)  Thanks

MisterJames:
@stevejgordon I know you've done some additional work on this topic on your blog. Any new insight or learnings you wanted to apply before we pick this up again? I think this is the right solution over the extension method, and as pointed out, provides a better test mechanism with lower friction.

stevejgordon:
@MisterJames @tonysurma Have to try and get my head back into this PR but I think it's still the better way to go and follows the pattern laid down in the Microsoft documentation as well.

I've been doing a fair bit of work for an API recently and in that case I've used custom middleware to validate authentication for a specific path route we have. However in that case it's a simple, one profile to one path mapping and easy to check a user's claims match up. Here we have different levels of resource that need checking back org/campaign/event etc.

Overall I think the approach above is probably easier to understand and apply as we grow the application. I'm happy to get this started up again and perhaps get one or two controllers going through this flow as a starting point. If during that it doesn't feel right we can review again. As we've refactored a lot since I put this in I'm thinking I might create a fresh branch and work there.

AllReadyApp/Web-App/AllReady/Security/AuthoizationHandlers/OrganizationOperations.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.AspNet.Authorization.Infrastructure;
+
+namespace AllReady.Security.AuthoizationHandlers
```
dpaquette: Spelling mistake in namespace here. *AuthorizationHandlers*

dpaquette:
@stevejgordon I am a big fan of this approach in general. I think it really helps to clearly state the resource authorization requirements (e.g. Is this person allowed to edit a campaign?) and makes it much easier to test.

stevejgordon:
I'm going to park this WIP PR since we'll be going through some change for IdentityServer and we can incorporate as necessary during that work

PR #990, opened 2016-07-03
Import Red Cross data via CSV

MisterJames:
- allows basic import with no error handling
- I will work on a couple of needed unit tests

Using the sample file that we got from the Red Cross (offline) this new end point on the admin side of the site will allow us to import requests.

The data provided by the RC works, so I haven't yet implemented much by way of error handling.


MisterJames:
@tonysurma I have created this as a PR now because it don't have any migrations, it is functional, we don't have the RTM passing yet and I am travelling for the next day or so. I'll do what I can to respond to any questions while in transit.  I have a few tests to add, which I'll try to do on my flight, but I'll submit as a separate PR.

tonysurma:
There is one failing hit test but I think it is okay, going to merge and we can fix if it is an issue

PR #998, opened 2016-07-08
Update to RTM, with build scripts

BillWagner:
This is an update of #989, with the build scripts updated as well


BillWagner:
@tonysurma This should be ready, and we can see if the kudu build works.

Note: DO NOT Squash and merge this. If you do, it will squash all @shawnwildermuth 's commits into mine. Leave it as several commits.

tonysurma:
Excellent, will do

tonysurma:
I have done the work to make the azure build work, I am going to accept this and them implement on top of it the changes to deploy.cmd, web.config etc. needed to make it work.  I promise to reference this PR so we can track it all along with the issue it came from.

Issue #724, opened 2016-04-25
Remove Dependency on Microsoft.Extensions.DependencyInjection for any unit tests that test classes that inherit from InMemoryContextTest


mgmccarthy:
In most unit tests that test classes that inherit from `InMemoryContextTest`, the context is retrieved using this code:
```
var context = ServiceProvider.GetService<AllReadyContext>();
```

`InMemoryContextTest` inherits from `InMemoryContextTestBase`, which exposes `AllReadyContext` as public getter `Context`:
```
protected AllReadyContext Context { get; private set; }
```

Therefore, instead of taking a dependency on Microsoft.Extensions.DependencyInjection in unit test code, this line of code:
```
var context = ServiceProvider.GetService<AllReadyContext>();
```

should be changed to this:
```
var context = Context;
```

- cleaner code
- reducing unneeded dependencies
- taking a singular approach to tests inheriting from `InMemoryContextTest`

digitaldrummerj:
@mgmccarthy @tonysurma can this issue be closed?  it looks like a PR was merged that fixes the issue.

PR #999, opened 2016-07-11
Removed Dependency on Microsoft.Extensions.DependencyInjection for unit tests that test classes that inherit from InMemoryContextTest

auroraocciduusadmin:
Borrowed the code from TestBase to add a test fixture using xUnit's Collection Fixtures feature so that the ServiceProvider and AllReadyContext used in some test methods will be created only once. In AddTeamMemberCommandHandlerAsyncTests, instead of inheriting from InMemoryContextTest, the test runner supplies the required context.
Running the tests in my environment showed a small improvement in time so I didn't implement the change in the other test classes that inherit from InMemoryContextTest as well until I see how the change is received.
The test class I started with: AddTeamMemberCommandHandlerAsyncTests previously started each test with an empty database. But since the strategy used here doesn't create a new database for each test, the database has to be cleaned-up at the end of each test (handled by Dispose() in xUnit).  I suspect there is a better way to "tear down" an in-memory database but this was the best I could find so far. "EnsureDeleted" can be used to get rid of the data, but the change tracking persists and interferes with subsequent tests.

In a subsequent commit, changed tests that inherit from InMemoryContextTest so they use the base class' Context instead of creating one with the ServiceProvider (issue #724).


MisterJames:
@auroraocciduusadmin I think generally speaking this looks good, but I also am hesitant because we're introducing a new pattern. Do you have plans as well to convert other similar tests?

As far as the build failure goes, we're just getting the script cleaned up with the move to Core and `dotnet`.

tonysurma:
just moved to RTM so will need to be rebased and retested cc @MisterJames

auroraocciduusadmin:
I did intend to convert other tests, but I was hoping for some feedback before going ahead. I also was fishing for a better way to "tear-down" the database context.

stevejgordon:
There have been a few discussions on slack regarding the test base class, especially as RTM changes resulted in us getting a single InMemoryDb instance rather than a fresh one per test run. This caused quite a few test failures as each test was trying to load some data into the same Db instance which was resulting in conflicts.

Vishal has dug into this and found a solution in #1005.

I'm not convinced we want a shared service provider and context for most tests. This surely results in some issues with data added/expected in one test, potentially affecting another. While creating these per test may be less efficient, it does allow for a fully clean starting point. If data from a prior test corrupts or could affect other tests we could start running into some real issues (similar to what we're seeing post RTM changes).

While the test base helps newcomers get going, it does start to hide some of the behaviour for the setup and could result in incorrect assumptions being made. @mgmccarthy also had some ideas around this.

It certainly feels like we need to standardise the approach to the tests. In some specific sequences of tests, sharing a context and InMemDb might be useful, but generally I feel each should responsible for getting a fresh context and db + setting up any required sample data so the tests stand alone.

mgmccarthy:
@tonysurma @MisterJames @stevejgordon @VishalMadhvani 

my two cents is we shouldn't be carrying the same db context across test methods in a unit test class. I realize that many of our unit tests use an overridable "LoadData" method that can contain setup data at a unit class level.

The problem with this approach is we're putting a LOT of data into an in-mem DbContext, and then running what should be autonomous unit tests using a shared data. Test setup/teardown should be done on a method by method basis to attain a clean starting point for each test.

Currently, you CAN use the existing provider like this, just reach out directly to Context from the base class, and call Context.Database.EnsureDeleted and don't use "LoadData" to load the data, but with the RTM upgrade, there has been churn around how to do this. 

The other thing I'd like to see go away in our existing test base classes is the other setup being done by the ServiceProvider. By this I mean setup being done on anything else but AllReadyContext. You can see this setup being done by TestBase and InMemoryContextTestBase.

I also shared this Gist with the Slack channel as an example of what I am thinking from a "test method base" class. https://gist.github.com/mgmccarthy/5e780c7c9b5c38c258ac5f7031fcb927. But I'm not too sure how pertinent it is anymore after the RTM upgrade. 

It looks like @VishalMadhvani addressed this same issue with #1005

VishalMadhvani:
+1 on using a clean database per test. Otherwise there is an increased likelihood of brittle tests.

Just to clarify, the LoadTestData method on TestBase in RC1 did exactly that, however due to changes in EFCore RC2 and RTM this was no longer the case. #1005 restores the desired behaviour.

On the flip side, there seems to be some interest in sharing a database across tests and hence in the interest of making it easy for people to contribute I wouldn't enforce against it. In fact if the test author is disciplined in only asserting state and not modifying it then this is a viable use case.

 @mgmccarthy I'm happy to see LoadTestData go and a general clean up of "magic" as you have in your gist. I think given that the docs and guidelines haven't been updated, I'd recommend waiting until they are.

mgmccarthy:
@VishalMadhvani is right, we need to support both... per test and per test class. Although we all agree that a shared context is not a good idea across multiple unit tests
1. some developers might be more comfortable working in this fashion
2. if we were to remove context per test class, and "LoadTestData" were to go away, we would have a LOT of unit test setups to re-write in our existing tests.

We can refactor slowly towards context per test as we're in the code base working.

VishalMadhvani:
Not so much a rewrite @mgmccarthy. The code from LoadTestData would just move to the test class constructor, or a LoadTestData method in the test class.

auroraocciduusadmin:
Well, I was a little challenged with git, but I think it looks ok now. I've rolled-back on the test-fixture and now the changes reflect issue #724.

VishalMadhvani:
@auroraocciduusadmin the changes looks good.

Yeah git takes a while to get used to. A few people who contribute to this project have written some good posts on common workflows. Unfortunately I don't have links handy but hopefully someone will point you to them.

Also, the PR title no longer matches the changes you have made so I suggest updating that.

tonysurma:
@VishalMadhvani @mgmccarthy @MisterJames are we good to pull this in?  Thanks @auroraocciduusadmin for the contribution

VishalMadhvani:
@tonysurma yeah looks good

PR #1001, opened 2016-07-15
Replaced 'magic strings' with constants in AccountController #340

nedruk:
Strings replaced:
- view names
- view data
- controller names
- area names

Custom messages left as is.


AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
```diff
@@ -411,5 +411,29 @@ private static void RetrieveFirstAndLastNameFromExternalPrincipal(ExternalLoginI
             firstName = array[0];
             lastName = array[1];
         }
+        
+        //Area Names
+        private const string ADMIN_AREA = "Admin";
+
+        //Controller Names
+        private const string HOME_CONTROLLER = "Home";
```
MisterJames: Can you change these to use `nameof` inline instead of consts?

mgmccarthy: @MisterJames, can we change title of #340 to use `nameof()` instead of magic strings? We should not be using constants anywhere we can avoid them (and I like to avoid them at all costs).

nedruk: nameof(HomeController) results in "HomeController". Do you want me to add some string operation, like nameof(HomeController).Replace("Controller","")? Looks a bit messy, imho. 

Re nameof area, I can't seem to find that implemented. nameof() documentation says it is "used to obtain the simple string name of a variable, type, or member". Not sure any of those can be applied to area, it is defined as a "logical grouping". Please correct me if i'm wrong.

stevejgordon: To add to this I'd read MS guidance in some of their GItHub issues not to rely on nameof for getting the controller or action names. There are attributes which can override the action name which could then differ from the method name, so nameof could return incorrect results.

mgmccarthy: @nedruk, unless the string "HomeController" is used more than once in the class, then you can just use the string... I'd argue that even if it's used more than once, still use "HomeController". We should be avoiding constants unless we absolutely need them.

@stevejgordon, I didn't realize that MS is guiding towards not using nameof() for action names on controller's. Is this b/c of attributes like `[Route("STRING")]`? I'm only asking b/c we're using nameof() all over our code base to resolve action names on controllers instead of using a string with the name of the action.

nedruk:
I came across this conversation on git re using nameof with controller names [https://github.com/aspnet/Mvc/issues/2608](url). Is that the one you were referring to @stevejgordon? 

@team, I feel the need to clarify the requirements to keep working on this issue. I'll outline everything I've grasped so far. Please feel free to correct me if i'm wrong or add something if i messed it out.

1. we do not need to replace 'magic strings' with constants, but rather replace them with nameof wherever it's possible.
2. the majority of 'magic strings' fall in the following buckets:
- action names
- view data
- controller names
- area names
- custom messages
3. nameof doesn't work for controllers, areas and VeiwData, so those should be left as is.
4. the only candidates for replacement are action names.

Bottom line, what needs to be done is to replace all hardcoded action names with nameof. 

Please confirm, so that i could get the ball rolling. Thx.

MisterJames:
@nedruk totally fair comments. We'll discuss on the standup this AM and get some notes in the repo on this.

tonysurma:
After @MisterJames responds this will need to be rebased and retested as we just moved to RTM

stevejgordon:
@nedruk Yeah, that looks like it might be the issue I'd read. While we're not using the ActionName attribute here currently, it could bite us later is someone introduced it somewhere.

An extension to manage to remove the controller part from the result of nameof might an option for controllers in the short term.

MisterJames:
@nedruk This one is a little long-running, but I'd be happy to merge it in if you can get the conflicts resolved. There might be a few touch-ups, but we can deal with that after the merge...nothing that we should block on. Thanks again for your contribution.

nedruk:
@MisterJames , sorry but I've accidentally deleted the related local branch. I've pulled down a local copy of this PR and fixed the conflicts, but can't find a way to push the changes back to PR. I'm very new to GitHub, and not even sure that such a way exists.. Do you want me to create another PR for this issue? Or given the amount of questions that my previous attempt raised, it'd be better to hold off for now? ;)

tonysurma:
@MisterJames wondering if you ever got the notification for @nedruk 's comment above?

MisterJames:
@nedruk Yes, please go ahead and get an new PR in and we'll move it along. Thanks!

PR #1006, opened 2016-07-18
Refactored ContactExtensions.UpdateCampaignContact to fix issues when creating and modifying campaigns.

VishalMadhvani:
Refactored to improve readability and fix EntityFramework related errors.
I think the tests for this method were broken and hence skipped so I'll need to take a look, however I've kicked and punched it manually and all seems good :grin:

Would appreciate an extra pair of eyes on it incase I've missed a use case.


mgmccarthy:
@VishalMadhvani, this is just a suggestion. Since `UpdateCampaignContact` is only used by `EditCampaignCommandHandler`, I think it would be easier/clearer to move that code out of the extension method and into the handler code. Then delete the extension method. That way, we don't have extension methods "hiding" database access code.

VishalMadhvani:
Yep @mgmccarthy agreed. ```EditCampaignCommandHandler``` needs to be reviewed and possibly cleaned up. There is potentially more bugs in there and there may also be more cases where we use extension methods to do stuff. I'll raise an issue for this.

But for now, I'll move the code out of the Extension class and into the Handler class.

PR #1014, opened 2016-07-20
Fixing event creation flow by utilising the AddOrUpdate extension helper

### LINKED ISSUE #1013: Creating an event fails with an EF error
Error is:

DbUpdateConcurrencyException: Database operation expected to affect 1 row(s) but actually affected 0 row(s). Data may have been modified or deleted since entities were loaded. See http://go.microsoft.com/fwlink/?LinkId=527962 for information on understanding and handling optimistic concurrency exceptions.

I've already had a look and it seems to be related to the use of context.Update (rather than Add) for this new untracked entity. I presume this is RC1 to RTM changes in how this works.

I have found an existing Extension method for the context to handle AddOrUpdate which I'll add to the Event as a fix.

stevejgordon:
Fixes #1013

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Events/EditEventCommandHandler.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Events/EditEventCommandHandler.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using AllReady.Areas.Admin.Models;
+using AllReady.Extensions;
 using AllReady.Models;
 using MediatR;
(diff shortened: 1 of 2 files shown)


mgmccarthy:
@stevejgordon @MisterJames 

Is our `AddOrUpdate` operation a destructive operation, or is there EF magic behind `ctx.Update(entity)`; where EF will first query the existing record, figure out what changed, and only update those properties?

I've used .AddOrUpdate from EF 6's Migrations namespace, and that AddOrUpdate is a "destructive" operation. It wont' query first to see how the incoming model differs from the one in the db, and only update the fields it sees as changed. It takes the incoming model as the source of truth and all props are written to the db.

PR #1019, opened 2016-07-20
 Fixing up use of asp-validation-summary per the latest docs

### LINKED ISSUE #1018: Create task page blows up.
Clicking create task on an event causes the following error.

image

👷 Working on it

stevejgordon:
Fixes #1018 

Looking at the latest docs it seems we shouldn't prefix the asp-validation-summary value with the enum name any longer.

ValidationSummary.ModelOnly

becomes

ModelOnly

I've updated the three occurrences I found of this usage.


dpaquette:
Looks good. :shipit:

PR #1025, opened 2016-07-21
WIP: Fix Date and DateTime pickers

VishalMadhvani:
Had a quick look into this and it seems when initialising the datetime picker plugin, it clears the existing date.
I assumed this issue was new post dotnet core RTM update but that now seems unlikely.
The plugin has been used since Nov 2015 I think.

I dont have the dotnet core rc1 tooling installed any more so i'd appreciate if someone can check if this work before the update.

The fix currently works for Events and Tasks dates because they use ".datetimepicker" but not Campaigns because it uses ".datepicker".

Seems like for a ".datetimepicker" we serialise the date using the default formatter, however ".datepicker" uses a custom formatter. I'm not currently sure how all this works across different locales.

Hopefully will have some time to continue looking at it over the weekend.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Views/Shared/_Layout.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Views/Shared/_Layout.cshtml
@@ -153,11 +153,17 @@
                 showClose: false,
                 toolbarPlacement: 'bottom'
             });
+            $(".datepicker").each(function (index, element) {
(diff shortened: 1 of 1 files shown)


VishalMadhvani:
https://github.com/Eonasdan/bootstrap-datetimepicker/issues/1521

VishalMadhvani:
@MisterJames this was still WIP...
It fixed the dates not being populated but there are still locale issues.

PR #1028, opened 2016-07-22
Home Page shows no Campaigns when there are campaigns seeded

### LINKED ISSUE #1027: Home Page shows no Campaigns when there are campaigns seeded
when loading HomeController/Index, the page is displaying with no Campaigns
image

looking at the view model before it's sent back to the view, there are six campaigns that were populated from a query handler:
image

so we should be showing these campaigns. 

Looking closer at \Home\Index.cshtml it looks like we're using knockout to populate the list of campaigns. Checking out the errors via Chrome's debugger, looks like knockout is throwing this error:
image

Not being a javascript or a knockout guy, my initial **guess** is knockout can't "define" `ImageUrl` b/c it's included on each `CampaignViewModel` coming back in the list of campaigns
image

and the `ImageUrl` used on `CampaignSummaryViewModel`.
image

which are both included on the `HomePageViewModel` returned to the view.

That, or there is a problem with the version of the knockout library we're using in conjunction with RTM? (this is also a guess).

A very simple fix for this is to not use knockout. We're showing a simple table of data here, so MVC is the easiest way to make that happen. After a quick refactor to use MVC/razor view to show the data, the page is working again:
image

Should I submit a PR for this by "knocking out"  :facepunch: the knockout code?

mgmccarthy:
Fixes #1027

If this is not the right fix, or there was something else we were relying on knockout for on /Home/Index.cshtml, then feel free to delete this PR.

I am out of town all next week and wanted to get submit PR if it does happen to fix the problem correctly


VishalMadhvani:
👍 looks good.. using knockout here seems completely unnecessary.
The StartDate and EndDate need formatting however.

mgmccarthy:
I'm on vacation this week no no laptop. If someone wants to merge the PR and the fix the formatting it should be a simple change

mgmccarthy:
@VishalMadhvani formatting has been fixed. Looks like C# doesn't support ordinal formatting of days (st, nd, rd, th, etc...) out of the gate, but I don't think it's that big of a deal to not have it there. If it is, I can write something that will produce this type of output

PR #1035, opened 2016-08-01
Updating HomeController to make both mediator actions async on Index action method

### LINKED ISSUE #1034: Updating HomeController to make both mediator actions async on Index action method
I'm currently working on this

mgmccarthy:
Fixes #1034


AllReadyApp/Web-App/AllReady/ViewModels/Home/Index.cs
```diff
@@ -1,11 +1,10 @@
 ﻿using System.Collections.Generic;
-using AllReady.ViewModels.Campaign;
 
 namespace AllReady.ViewModels.Home
 {
-    public class HomePageViewModel
+    public class Index
```
stevejgordon: @mgmccarthy I was glancing through this at lunchtime. Curious on the naming choice for this class. While sometimes a bit redundant I tend to prefer the ViewModel suffix to make it pretty explicit what these classes are.

AllReadyApp/Web-App/AllReady/web.config
```diff
@@ -1,9 +1,9 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <system.webServer>
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath=".\AllReady.exe" arguments="" forwardWindowsAuthToken="false" stdoutLogEnabled="false" />
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" forwardWindowsAuthToken="false" stdoutLogEnabled="false" />
```
stevejgordon: @mgmccarthy I've had to make this change locally to get it building and running. I think this was originally the setting until the RTM work. Do we know if this will still build / run in AppVeyor if this is committed?

stevejgordon:
@mgmccarthy I've not run it up, but looks okay to me after reading it through. I added two specific comments in the changes.

AllReadyApp/Web-App/AllReady/ViewModels/Home/Index.cs
mgmccarthy: @stevejgordon I agree on making the class say `ViewModel`... more specific is better. That being said, the class is located in a ViewModels directory, so there is still clarity to how the class should be used when looking at the solution, but adding `ViewModel` to the class name would be more specific when looking at the class outside the context of the solution.

Some of these refactorings are part of what I'd like to see on a larger scale in this project, where we're building a specific ViewModel around GET/POST action methods instead of dragging one big ViewModel with 50+ properties across ALL actions methods in a controller, or even worse, sharing ViewModels between controllers.

AllReadyApp/Web-App/AllReady/web.config
mgmccarthy: @stevejgordon, yeah, this is a mystery to me. When I run locally, VS 2015 changes this to `.\AllReady.exe`, from the original value of `%LAUNCHER_PATH%`. I'm not too sure why. I've been good an undoing my changes to this file before commiting, but this time I forgot. Seeing how this PR built successfully in AppVeyor, my assumption is it builds fine. As far as it actually running in whatever environment we have set up that is not our local dev environment, I'm not too sure.

AllReadyApp/Web-App/AllReady/ViewModels/Home/Index.cs
stevejgordon: @mgmccarthy Sounds good. I too much prefer smaller, very specific view models. Generic ones just become a pain and bloated with unused properties. 100% behind that refactor.

AllReadyApp/Web-App/AllReady/web.config
tonysurma: For now please don't change/update web.config in your PR.  With the %% stuff it won't work in production on the site.  Not sure yet why post RTM this happens local and is mutually exclusive with running live

tonysurma:
If this looks good to @stevejgordon and web.config changes aren't in there after a rebase we should be able to merge.  Thanks

AllReadyApp/Web-App/AllReady/web.config
mgmccarthy: @tonysurma, sorry.  This snuck into the PR. When I F5 locally, the IDE checked out the web.config file and made this change. When I committed, I didn't realize the change that had been made

Issue #662, opened 2016-04-06
write unit tests for AllReady.Controllers.EventController


mgmccarthy:
@tonysurma please label with "jump-in". Part of Issue #620

stubs are provided in:
AllReady.UnitTest.Controllers.EventControllerTests.

digitaldrummerj:
@tonysurma I see in the PR that has been merged that you put a closes this issue comment in but the issue is still open.  Is it still a valid issue?

mgmccarthy:
@tonysurma @digitaldrummerj the unit tests exist and this Issue can be closed

PR #1040, opened 2016-08-01
Tests for AllReady.Controllers.EventController issue #662

danielepo:
Implemented tests for ChangeStatus, Index and ShowEvent.
Removed stub tests for UpdateMyTask, GetMyTasks and GetMyEvents as the controller no longer has these methods.


danielepo:
I looks like I messed up the branch, will clean it up an re-do the PR

Issue #568, opened 2016-02-20
Show Campaign Impact Goal status?


palittle:
Should the Campaign page show the Campaign Impact Goal status in the green box as it does on the Admin Campaign page? The green box on the Campaign page looks incomplete so I am assuming that it show this.

`http://allready-d.azurewebsites.net/Campaign/7`
![image](https://cloud.githubusercontent.com/assets/17350634/13199768/a6beb0f8-d7eb-11e5-98e0-3019c6dc74f9.png)

`http://allready-d.azurewebsites.net/Admin/Campaign/Details/7`
![image](https://cloud.githubusercontent.com/assets/17350634/13199753/413107fe-d7eb-11e5-901c-7e2e843555ab.png)

danielepo:
this was fixed by PR https://github.com/HTBox/allReady/pull/1047

digitaldrummerj:
@tonysurma looks like this can be closed since it was fixed in the PR.

PR #1047, opened 2016-08-04
Fix System.js missing in Campaign details

danielepo:
The effect of this missing library was that the thermometer was missing in the campaign details view


digitaldrummerj:
fixes #568

PR #1054, opened 2016-08-06
Add Test for EventController

danielepo:
Tests for AllReady.Controllers.EventController: issue #662
Implemented tests for ChangeStatus, Index and ShowEvent.
Removed stub tests for UpdateMyTask, GetMyTasks and GetMyEvents as the controller no longer has these methods.


tonysurma:
closes #662

PR #1090, opened 2016-08-07
Geocode request addresses

### LINKED ISSUE #1077: Geocode addresses on requests as they are imported via CSV
If there isn't a lat, long supplied geocode and add this to the request

- [ ] check for lat long incoming
- [ ] update request model for lat, long
- [ ] geocode address to lat, long
- [ ] persist this as we import

### LINKED ISSUE #1078: Geocode addresses on requests as they are posted to the external API
If there isn't a lat, long supplied geocode and add this to the request

- [ ] check for lat long incoming
- [ ] update request model for lat, long
- [ ] geocode address to lat, long
- [ ] persist this as we import

will reuse work from #1077 as this is same work but from two entry points

mheggeseth:
* Updated CSV import spec to support specifying lat/long in the CSV (api/request already handled it)
* If incoming Request entity does not have latitude and longitude specified, use Google's Geocoder API to resolve lat/long from Address, City, State, and Zip. The API returns a list of addresses of which we will trust that the lat/long of the first address is the one we want.

Fixes #1077 
Fixes #1078 

**Note:** We are currently using Google's Geocoding API anonymously which is fine for now, but it is subject to [usage limits](https://developers.google.com/maps/documentation/geocoding/usage-limits) so we will need our own API key at some point. I'll write this up as a separate issue.


mheggeseth:
@tonysurma This is good to go.

Issue #1084, opened 2016-08-07
Messaging Updates for AllReady Volunteer Communication


tonysurma:
- [ ] Requestor can specify communication preferences (text, voice) for communications
- [ ] Promote text and fall back to voice
- [ ] Upon adding a request to an itinerary send a "scheduled for [date] notification with "Y" response for setting request to confirmed, "N" for setting request to unassigned
- [ ] Daily Web Job for 7 days before scheduled date "scheduled for next week on [date]"
- [ ] If request is confirmed add "'N' if you can no longer make it" and if they respond N mark unassigned
- [ ] If request is not confirmed add "'Y' if you can make it, 'N' if you can't" and set status as above
- [ ] Daily Web Job for day before scheduled date "scheduled for tomorrow on [date]"
- [ ] If request is confirmed add "'N' if you can no longer make it" and if they respond N mark unassigned
- [ ] If request is not confirmed add "'Y' if you can make it, 'N' if you can't" and set status as above
- [ ] Daily Web Job for day of for requests not confirmed message "sorry you couldn't make it, we will reschedule" and mark unassigned
- [ ] On Installer marking previous request in itinerary delivered day of send "See you in a few" message

tonysurma:
ping @OhMcGoo

OhMcGoo:
![allready communications workflow](https://cloud.githubusercontent.com/assets/4147733/17463624/58a55bfa-5c8f-11e6-9f62-1ecff34faad5.jpg)

seanepping:
@tonysurma I am looking into this

mgmccarthy:
@tonysurma  @seanepping 

we could look into using a code-based scheduler for this functionality: http://www.quartz-scheduler.net/

a lot more flexible than kicking off a process from a build in windows scheduler.

tonysurma:
PR #1111 starts implementation of this

mgmccarthy:
@tonysurma, what I'd like to add/fix from that PR is communication preferences should not be on the message, but rather stored as user preferences and the lookup for how to communicate to the user will be done via a read the based on preferences,  decide how to send.

mgmccarthy:
@tonysurma, can you assign this to me? I'd like to start with the first item, then discuss/investigate how to get some of the other things done.

It's a shame this project is not using NServiceBus, b/c Sagas would be a very good way to coordinate the work with the "web jobs" for the communication of confirming requests.

[...]

PR #1111, opened 2016-08-07
WIP - #1084 - Added Initial Notification on adding request to Itinerary - Naive assumption of sms

seanepping:
Need to add communication preferences to requests
The preferences will be used to either send email or text

Start to implement #1084


AllReadyApp/Web-App/AllReady/DataAccess/IAllReadyDataAccess.cs
```diff
@@ -8,72 +9,81 @@ public interface IAllReadyDataAccess
...
         IEnumerable<Event> Events { get; }
-        Event GetEvent(int eventId);
-        int GetManagingOrganizationId(int eventId);
-        IEnumerable<EventSignup> GetEventSignups(string userId);
-        IEnumerable<EventSignup> GetEventSignups(int eventId, string userId);
-        Task UpdateEvent(Event value);
-        IEnumerable<Event> EventsByPostalCode(string postalCode, int distance);
-        IEnumerable<Event> EventsByGeography(double latitude, double longitude, int distance);
-        IEnumerable<Resource> GetResourcesByCategory(string category);
+        Event GetEvent( int eventId );
```
MisterJames: Can you check your IDE settings? It seems like something is adding spacing around parens, which  will change a lot of lines of code and make it difficult to spot the changes for a reviewer. Thanks!

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using System.Linq;
+using System.Threading.Tasks;
```
MisterJames: Are changes needed in this class?

AllReadyApp/Web-App/AllReady/DataAccess/AllReadyDataAccessEF7.Itinerary.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
+
+namespace AllReady.Models
+{
+    public partial class AllReadyDataAccessEF7 : IAllReadyDataAccess
```
MisterJames: We are actually moving away from the `AllReadyDataAccessEF7` class usage. Have you seen the work we're doing with handlers? I'd be happy to point you at some resources or Skype up with you to walk you though implementation if you like.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Itineraries/AddRequestsCommandHandlerAsync.cs
```diff
@@ -4,46 +4,41 @@
...
 namespace AllReady.Areas.Admin.Features.Itineraries
 {
     public class AddRequestsCommandHandlerAsync : IAsyncRequestHandler<AddRequestsCommand, bool>
     {
-        private readonly AllReadyContext _context;
+        private readonly IAllReadyDataAccess _data;
         private readonly IMediator _mediator;
 
-        public AddRequestsCommandHandlerAsync(AllReadyContext context, IMediator mediator)
+        public AddRequestsCommandHandlerAsync( IAllReadyDataAccess data, IMediator mediator )
```
MisterJames: What is the driver to avoid using the context in the handler? I've made another comment about moving away from the `IAllReadyDataAccess` class as well, feel free to explore this in the project and reach out as required. We're happy to help :)

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/NotifyRequestorsCommandHandlerTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using Moq;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace AllReady.UnitTest.Areas.Admin.Features.Requests
+{
+    public class NotifyRequestorsCommandHandlerTests
+    {
+        [Fact]
+        public async Task ShouldSendSmsForEachRequestWithCorrectPhoneNumber()
```
MisterJames: I like the approach here with an array of tests, however, it introduces logic into the test itself. I would recommend having a look at [theories and inline data](https://xunit.github.io/docs/getting-started-desktop.html#write-first-theory), which would likely lead you into something like [property and class data] for theories. 

The idea is: you write a test that accepts parameters, namely the test data and the expected result, then you simply provide the parameters through attributes, properties or instances of classes.

The benefit is that you end up with a test that is clear in what it is doing and your test data can easily grow without modifying the test.

If you are not testing for specific data points here, I would also suggest considering [GenFu](http://genfu.io/) for test/sample data generation as you would not have to new up objects on your own.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Itineraries/AddRequestsCommandHandlerTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            var command = new AddRequestsCommand
+            {
+                ItineraryId = itinerary.Id,
+                RequestIdsToAdd = requests.Select(x => x.RequestId.ToString()).ToList()
+            };
+
+            var iRequestList = new List<ItineraryRequest>();
+
+            foreach (var request in requests)
```
MisterJames: Same comment here re: testing, please move to xUnit theories.

mgmccarthy:
@seanepping, are you still working on this?

mgmccarthy:
@tonysurma do you want me to pick this up? I looked at the linked Issue and it looks like it's a P3 (used to be P1?)...

tonysurma:
@mgmccarthy please do.  It is P3 for October (as import is the priority) but this will elevate in Priority for November and is definitely in scope for release in December.  Thanks

mgmccarthy:
@tonysurma, can you assign #1084 to me? Thanks

mgmccarthy:
@stevejgordon, I'm taking a tour through this PR right now and I've pulled it down locally and attempted a rebase with master. So far, it's looking OK, but I see a singular concept in the system (add a request) represented by two different commands/handlers:

- the existing `AddRequestsCommand` which you wrote along with it's handler
- the newly added `AddRequestCommand` which was added in this PR

I'm not too sure if you knew about this, or had talked to the author of this PR about it. I'm not too sure if adding a request via an external requestor (RequestApiController) vs. adding a request via the application (via IntineraryController) are different, but I"m thinking they probably aren't.

Do you have any domain knowledge on whether these two things are really the same thing, just the source of the request is different?

Thanks

stevejgordon:
@mgmccarthy I think we should end up with a single handler for adding requests. We have three ways to add requests.

1. API call to our endpoint from GetASmokeAlarm
2. Manually via the UI
3. Import via CSV

There may be differences in the validation rules for each and the data required may be slightly different. The ProviderId for example will be null when supplied manually.

mgmccarthy:
This PR can be deleted once PR #1400 is merged

tonysurma:
Included in later pr

PR #1128, opened 2016-08-12
automatically verify email (do not send confirm email message) when registering via external auth providers that provide email

### LINKED ISSUE #965: automatically verify email (do not send confirm email message) when registering via external auth providers that provide email
@tonysurma, please verify this "requirement" looks correct.

broken out from final comment by Tony Surmar on #955. (" If there was a way to detect the email was prepopulated and therefore not send verification for the email that would be even nicer but not required.")

If the external auth provider being used gives us access to the user's email address (example: Facebook gives us the email address, Twitter does not), confirm email address upon completion of registration w/out sending an email confirmation message

mgmccarthy:
Fixes #965

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
@@ -318,7 +318,8 @@ namespace AllReady.Controllers
                 FirstName = externalUserInformation.FirstName,
                 LastName = externalUserInformation.LastName,
                 ReturnUrl = returnUrl,
-                LoginProvider = externalLoginInfo.LoginProvider
+                LoginProvider = externalLoginInfo.LoginProvider,
+                EmailIsVerifiedByExternalLoginProvider = !string.IsNullOrEmpty(externalUserInformation.Email)
             });
         }
 
@@ -399,6 +400,7 @@ namespace AllReady.Controllers
             }
         }
 
+        //TODO: either extract this into a provider pattern or somehow decide which of these conditionals are invoked by which caller and get rid of this method.
         private IActionResult RedirectToLocal(string returnUrl, ApplicationUser user)
         {
             if (Url.IsLocalUrl(returnUrl))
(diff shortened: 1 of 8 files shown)


AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
```diff
@@ -393,6 +400,7 @@ private void AddErrorsToModelState(IdentityResult result)
             }
         }
 
+        //TODO: either extract this into a provider pattern or somehow decide which of these conditionals are invoked by which caller and get rid of this method.
```
MisterJames: There is the new `LocalRedirect` controller base method that could be considered.

MisterJames:
@mgmccarthy should be good to :shipit: after the rebase.

AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
mgmccarthy: @MisterJames.  I did not know about the base method. I'll check it out and make it part of the rebase.

mgmccarthy: @MisterJames, after further investigation, the `LocalRedirect` base method still does not solve the problem I'm trying to get at here.

For any action method that uses the existing `RedirectToLocal` method, this means that the unit test for that action method needs to test all four possible results of the outcome of the method. 

Since the method is called three times in the controller, and since there is no way of knowing whether or not some of the action methods in the controller invoke only some of the conditionals based on other criteria, it makes the AccountControllerTests class a mess.

A quick fix is to extract the `RedirectToLocal` method into its own class, encapsulating the conditional statements. When then write a test class class that tests the conditionals in the method.

This change frees AccountControllerTests from having to write four unit test methods for any action method that calls `RedirectToLocal` to account for the four possible execution paths.

A concrete example of this in AccountControllerTests is:
```
public async Task ExternalLoginCallbackRedirectsToCorrectActionAndController_WhenExternalLoginSignInAsyncResultIsSuccessful_AndExternalLoginInfoIsNotNull_AndUrlIsALocalUrl()
public async Task ExternalLoginCallbackRedirectsToCorrectActionAndController_WhenExternalLoginSignInAsyncResultIsSuccessful_AndExternalLoginInfoIsNotNull_AndUserIsASiteAdmin()
public async Task ExternalLoginCallbackRedirectsToCorrectActionAndControllerWithCorrerctRouteValues__WhenExternalLoginSignInAsyncResultIsSuccessful_AndExternalLoginInfoIsNotNull_AndUserIsAnOrgAdmin()
public async Task ExternalLoginCallbackRedirectsToCorrectActionAndController_WhenExternalLoginSignInAsyncResultIsSuccessful_AndExternalLoginInfoIsNotNull_AndUrlIsNotALocalUrl_AndUserTypeIsNotASiteAdminOrAnOrgAdmin()
```

Moving code from `RedirectToLocal` to a separate class would get rid of all these unneeded tests for all action method in AccountController.

PR #1131, opened 2016-08-13
Mark IAllReadyDataAccess obsolete

### LINKED ISSUE #622: Refactor all read operations on IAllReadyDataAccess to async
Since EF 7 supports async/await ops, we should be able to update all read ops on IAllReadyDataAccess to be async.

### LINKED ISSUE #1130: Mark AllReadyDataAccess methods as obsolete
The goal here is to discourage use of the repository pattern and move towards query/command patterns leveraging the `AllReadyContext` directly.

The `IAllReadyDataAccess` approach pre-dated a number of framework features and patterns that are available today (due to the pre-pre-pre-release status of Core at the time).  Implementing it in such a way is an abstraction over an abstraction that only gives the possible long-shot benefit of being able to switch to a different data store. Which...is highly unlikely.

Using queries, commands and handlers with `AllReadyContext` gives us direct access to the rich features of `DbContext` and EF.

From #622.

MisterJames:
- marked interface and method definitions as obsolete
- added disable for 0618 in project files (web and test)

Fixes #1130
Closes #622

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/project.json
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/project.json
@@ -4,7 +4,8 @@
   "authors": [ "mattgal" ],
 
   "buildOptions": {
-    "warningsAsErrors": true
+    "warningsAsErrors": true,
+    "nowarn": [ "0618" ]
   },
 
   "dependencies": {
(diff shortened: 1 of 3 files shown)


MisterJames:
Just as an FYI for @BillWagner, @dpaquette  or @mgmccarthy on review, it seems that with ASP.NET Core projects that even if you specify "false" as a param in the `Obsolete` attribute, warnings are treated as errors and the build fails. I had to add 0618 to the ignore list at the project level to get the build to pass locally.

mgmccarthy:
@MisterJames is this by design for ASP.Net Core, or do you think this is a bug?

mgmccarthy:
can we close this issue seeing as we're not doing what it says in the title anymore?

mgmccarthy:
@MisterJames, can you please merge this? We just had another PR submitted where IAllReadyDataAccess was used... thanks

PR #1142, opened 2016-08-17
Refactor TaskSummaryViewModel out of TaskController

### LINKED ISSUE #1141: Refactor TaskSummaryViewModel out of TaskController
@tonysurma please assign to me

Trying to align with simplifying our ViewModels (Issue #730), I refactored TaskController to use a ViewModel/Message/Handler per action method where it makes sense. I've also removed a call that was being processed synchronously from an async action method.

Note each handler is optimized for the ViewModel it is building for the UI that will eventually consume it. The only "shared" ViewModel left here is the EditViewModel, which is used for both Creation and Editing purposes (but note I've used comments to denote what properties on the ViewModel are being used and for which purpose).

The refactoring deters dragging a single ViewModel with a bunch of properties through all action methods on a given controller, or even worse, from sharing ViewModels across controllers. This approach aligns with SRP and deters view model sharing as well as writing extension methods that end up doing WAY more than they should be doing, especially when it comes to ViewModels.

TaskSummaryViewModel is still being used across multiple controllers/handlers. Not too sure if that will change, but I wanted to have some example code of how to break up ViewModels/handlers across a given controller.

mgmccarthy:
Fixes #1141


mgmccarthy:
Hmmm... I'm not too sure why this branch (RefactorTaskController) has these commits:
image

These commits are part of PR #1128, which is still waiting to be merged...

I branched from my master branch, and my master showed no changes at the point I created the RefactorTaskController branch...

sorry about that. Maybe there is a way I can get rid of those commits? The easiest thing to do would be to merge PR #1128 and the other commits don't really matter b/c they've already been merged

MisterJames:
@mgmccarthy  I didn't squash on that other branch (meaning, all of those first 9 commits should be over on master FingersCrossed), so hopefully nothing is required here, just letting GitHub catch up and I'll try to merge later this evening.

mgmccarthy:
@MisterJames, ah crap. I just blew away my local master branch and recreated it from /remotes/HTBox/master b/c I thought I had incorrectly, somehow, merged those commits into the master branch.

image

hope I haven't screwed anything up for myself going forward.

Again, my lack of Git knowledge here is pretty bad. Thanks for the clarification.

MisterJames:
No, I've seen this thing where - if the SHAs don't change - then the overlapping commits just wuzzle out of the mix. But this might be only in certain scenarios (we work with forks, so the SHAs change, I think).  Either way, I'll squash and merge, we should be good. Thanks for all the heavy lifting here @mgmccarthy!

Issue #1153, opened 2016-08-19
Introduce IDateTimeOffsetProvider to abstract the conversion of date time offsets.


mgmccarthy:
@tonysurma pleas assign to me.

I am currently working on this.

As part of an ongoing effort to make working with DateTimeOffset's easier, I've introduced a new abstraction, IDateTimeOffsetProvider.

This provider will be injected into code that needs to reach out to DateTimeOffset and supply it a TimeZoneId along with a DateTimeOffset (usually from a ViewModel) to convert to the correct value based on TimeZoneId.

stevejgordon:
Closing after work by @mgmccarthy in #1154 - Any further work to be discussed and addressed by future issues.

Issue #410, opened 2015-12-03
Consolidate DateTimeOffset conversion to a single place


dpaquette:
We are converting DateTimeOffset values to the Campaign timezone in a few different places. Would be nice to avoid the duplication of code here.

dpaquette:
Examples in [TaskController.cs](https://github.com/HTBox/allReady/blob/master/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/TaskController.cs), [ActivityEditCommandHandler.cs](https://github.com/HTBox/allReady/blob/master/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Activities/EditActivityCommandHandler.cs) and [EditTaskCommandHandler.cs](https://github.com/HTBox/allReady/blob/master/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Tasks/EditTaskCommandHandler.cs)

dpaquette:
Assign to me please

mgmccarthy:
we're also doing this work in [TimeTagHelper.cs:](https://github.com/HTBox/allReady/blob/53c7f60facfaaf7d016265087eadc13ac7720ac7/AllReadyApp/Web-App/AllReady/TagHelpers/TimeTagHelper.cs) and [campaign.js](https://github.com/HTBox/allReady/blob/53c7f60facfaaf7d016265087eadc13ac7720ac7/AllReadyApp/Web-App/AllReady/wwwroot/js/campaigns.jsrl), so it's pretty much all over the place at this point.

mgmccarthy:
@dpaquette, I have a VERY simple spike of a "provider" that wraps up a lot of the duplicate code here:
https://github.com/mgmccarthy/allReady/tree/DateTimeOffsetProvider

Again, this is just a baby step towards a better DateTime solution. This spike does not address the duplicate code that has leaked into [TimeTagHelper.cs](https://github.com/HTBox/allReady/blob/53c7f60facfaaf7d016265087eadc13ac7720ac7/AllReadyApp/Web-App/AllReady/TagHelpers/TimeTagHelper.cs) and [campaigns.js](https://github.com/HTBox/allReady/blob/b370c169dc7485067519f6fbe15a563f7acf19c9/AllReadyApp/Web-App/AllReady/wwwroot/js/campaigns.js), but I don't think this code belongs in tag helpers or javascript.

[...]

PR #1154, opened 2016-08-19
Introduce an abstraction for the conversions of DateTimeOffsets

mgmccarthy:
@dpaquette

Addresses #1153 and #410

**this is in no way meant to be the final version, this is an initial take.** 

Here is a quick before and after of the EditTaskCommandHandlerAsync.

before:
```
var timeZoneInfo = TimeZoneInfo.FindSystemTimeZoneById(message.Task.TimeZoneId);
var startDateValue = message.Task.StartDateTime;
var startDateTimeOffset = timeZoneInfo.GetUtcOffset(startDateValue);
task.StartDateTime = new DateTimeOffset(startDateValue.Year, startDateValue.Month, startDateValue.Day, startDateValue.Hour, startDateValue.Minute, 0, startDateTimeOffset);

var endDateValue = message.Task.EndDateTime;
var endDateTimeOffset = timeZoneInfo.GetUtcOffset(endDateValue);
task.EndDateTime = new DateTimeOffset(endDateValue.Year, endDateValue.Month, endDateValue.Day, endDateValue.Hour, endDateValue.Minute, 0, endDateTimeOffset);
```

after:
```
task.StartDateTime = _dateTimeOffsetProvider.GetDateTimeOffsetFor(message.Task.TimeZoneId, message.Task.StartDateTime, message.Task.StartDateTime.Hour, message.Task.StartDateTime.Minute);
task.EndDateTime = _dateTimeOffsetProvider.GetDateTimeOffsetFor(message.Task.TimeZoneId, message.Task.EndDateTime, message.Task.EndDateTime.Hour, message.Task.EndDateTime.Minute);
```

Here are some thoughts/potential improvement/quetions:

- we could add an API for passing in both a start date and and end date along with a time zone id, and use out params to give the converted values back. This can be done instead of invoking the provider twice with the same time zone id and passing in the start, then the end date.
```
DateTimeOffset GetDateTimeOffsetFor(string timeZoneId, DateTimeOffset startDateTimeOffset, DateTimeOffset endDateTimeOffset, out DateTimeOffset convertedStartDate, out DateTimeOffset convertedEndDate);
```

- some of the existing DateTimeOffset calls specified values down to hours, minutes and seconds, and other did not. I'm not too sure what the intent here is, but if we want to support specifiying down to the second, the current API allows you pass those values in. If they're not present, then they default to 0.

- I did not try to do anything with reads. Since the translation on a read is currently scattered throughout the system (see #410), I'm assuming there will have to be more of a coordinated effort around centralizing that functionality. Right now it's a mix of a TimeTagHelper and javascript.  **IMO, javascript should not be doing any translation.**


mgmccarthy:
@MisterJames @BillWagner I would also be interested in getting your feedback on this.

dpaquette:
@mgmccarthy This is a good cleanup of the original code as it reduces a lot of duplication. I wonder what we can do to help convey the intention of the code or help to ensure it is being used properly.

The basic idea is that the DateTime that is passed in to the controller action needs to be adjusted to the timzone of the particular event/task (as opposed to the timezone of the server and the timezone of the client). One idea might be to rename the GetDateTimeOffset to AdjustDateTimeOffset??  Another idea is to do some custom model binding and try to abstract it that way but I'm always reluctant to go down that path.

mgmccarthy:
@dpaquette, I like this statement you just made: "The basic idea is that the DateTime that is passed in to the controller action needs to be adjusted to the time zone of the particular event/task (as opposed to the timezone of the server and the timezone of the client)."

And I'm looking for the type of feedback we you just gave me... which is how should this thing be named so developers understand what it's used for and how to use it.

Did you have any feedback for me as far as making the API take two dates in and return two converted dates? An example would be:
`DateTimeOffset GetDateTimeOffsetFor(string timeZoneId, DateTimeOffset startDateTimeOffset, DateTimeOffset endDateTimeOffset, out DateTimeOffset convertedStartDate, out DateTimeOffset convertedEndDate);`

since most of the time (at least from what I can see), we're always asking for start and end date conversions in the same transaction (handler). This would keep devs from having to use moq's SetupSequence in order to unit test code that uses the dependency.

VishalMadhvani:
@mgmccarthy @dpaquette how about preferring Utc DateTime from the client?

    DateTimeOffset GetDateTimeOffsetFor(string timeZoneId, DateTime dateTimeUtc)

mgmccarthy:
@VishalMadhvani, I'm following @dpaquette's lead here. I think we're always dealing with DateTimeOffsets, even when getting dates from the client... all the ViewModels used for write ops have properties that are of the DateTimeOffset type.

VishalMadhvani:
Fair enough... i'd err on the side of keeping things simple to allow more developers to get involved and contribute... this datetimeoffset stuff is mind boggling!
I appreciate going on a witch hunt to remove all usage of datetimeoffset is maybe a bit silly.. but some kind of halfway point surely is doable.

@mgmccarthy I'm trying to follow what you're doing and it feels like creating such an abstraction is just applying masking tape over a problem. Maybe I'm missing something but even with the work you're doing it's still hard to work with DateTimeOffset. I don't mean to call you out so apologies in advance but surely your time is better spent removing usages of DateTimeOffset over trying to come up with some custom way of handling them.

just my 2 cents...

mgmccarthy:
@VishalMadhvani,  I don't think you're calling me out. I welcome discussions about this type of thing. 

To paraphrase @dpaquette, he talked to someone on the ASP.NET team, and they recommended using DateTimeOffset. Matt Johnson (the guy who wrote NodaTime) recommends to store dates/times in one of two ways:

1. DateTime as UTC
2. DateTimeOffset

Here is one of the slides from his PluralSight video that shows the two ways he recommends storing data:
image

regardless of which type you store, you always need the context of a TimeZoneId for conversions.

Another thing to consider is we already have all the code written to work with DateTimeOffset. If we refactored to DateTime UTC, then we have to do more work.

The abstraction I'm writing will simplify some of the logic a developer needs to understand to get a DateTimeOffset conversion done (granted, the abstarction's API is still being worked on). It also centralizes the conversion so that code is encapsulated and only needs to be tested once, instead of classes that need the conversion reaching out to the static DateTimeOffset class. For the classes that use the abstraction, their unit tests now don't have to deal with the static DateTimeOffset class anymore.

**I do agree, however, that I'm less comfortable/familiar with DateTimeOffsets than I am with DateTime**, but that's probably because I've never worked with DateTimeOffset before. Where I work now, we store DateTime as UTC along with a TimeZoneId, then handle the conversion on write and reads.

@dpaquette, @VishalMadhvani does bring up a  good point that developers might be more comfortable/familiar with DateTime then DateTimeOffset. Do you any thoughts on this?

dpaquette:
I'm just on vacation this week. I understand the points that are raised but need to spend some time relearning the situation here. I know we were using UTC dates and it was even more confusing than it is now. Also, while datetimeoffset was unfamiliar to me when it was recommended, it made more sense when I learned it is just a datetime plus a time zone. It is a poorly named object IMO.

In my opinion, we should accept this pull request as it does improve things by reducing duplication. When I get back I will try to unravel some of the reasoning behind the current approach so we can figure out a path forward.

mgmccarthy:
@dpaquette, just pushed some of your recommendations (the renaming of the method on the provider from `GetDateTimeOffset` to `AdjustDateTimeOffsetFor`)... also added a new method that allows a user to pass in a TimeZoneId and a start and end date and get a converted start and end date back as out params.

I'm still fine entertaining storing as DateTime UTC vs. DateTimeOffset.

Have a good rest of your vacation.

mgmccarthy:
Latest update to the PR renames IDateTimeOffSetProvider to IConvertDateTimeOffset. Namings for implemenation class and injected dependencies changed throughout codebase. all unit tests run.

mgmccarthy:
adding some more fuel to the fire... here Matt Johnson suggests storing DateTime's as UTC:
image

mgmccarthy:
@dpaquette just rebased this. If you can get eyes on it, I'd like to either go with a merge, or make any changes you recommend and then push again. It's going to be tough keeping this rebased with all the changes/PR's happening right now

dpaquette:
@mgmccarthy I will review tonight

dpaquette:
Did a quick review and I think this looks good. It reduces the scattered code which will help with making any changes we need to make to the logic here much easier to implement.  Nice work @mgmccarthy 

@mgmccarthy looks like you have some conflicts so you might need to rebase again. 

Once this is merged, I will review the original design with respect to DateTimeOffset vs DateTime in UTC and see what we can do to improve on it.

mgmccarthy:
@dpaquette, rebased and ready for merge

tonysurma:
@mgmccarthy sorry but needs another rebase and then I will merge.  thank you

mgmccarthy:
@tonysurma, dev machine is smoked, until I get it up and running, I won't be able to rebase

mgmccarthy:
@tonysurma, this has finally been rebased. I'm pretty much back in action at this point. @dpaquette thanks for the help earlier today on Slack. the rebase was a little tricky, but I think I got everything correct.

tonysurma:
@mgmccarthy are we good to go on this then?

mgmccarthy:
@tonysurma, yep, good to go. @dpaquette and I will probably meet and discuss it, but it's the first step in the direction of abstracting some of the involved code in doing the DateTimeOffset conversions

PR #1161, opened 2016-08-20
Move CDN reference above local references

robrich:


robrich:
Closed in favor of https://github.com/HTBox/allReady/pull/1162 because apparently naming a branch `head` makes git grumpy.  :D

Issue #1082, opened 2016-08-07
Cordova - convert to a true ionic project


digitaldrummerj:
Right now the project is a Cordova Tools project that has ionic added into it but it is not a true ionic project.  The ionic framework has a great cli that provides functionality like converting sass to css, theming support, serve with live reloading.  

TACO is fully compatible with an ionic framework project that is generated from the ionic cli.

digitaldrummerj:
I am working on this

Issue #1081, opened 2016-08-07
Cordova - upgrade to the latest Cordova and Ionic versions


digitaldrummerj:
For Cordova the project is 5.1.1 and the latest version is 6.3

For Ionic is it using 1.0.1 and the latest version is 1.3,1

digitaldrummerj:
I am working on this

Issue #1076, opened 2016-08-07
Cordova Build - unable to find the sax module


digitaldrummerj:
Build Error with Cordova Project

```
Error : BLD00401 :
Could not find module 'sax'. 
Please Go to Tools --> Options --> Tools for Apache Cordova --> Cordova Tools --> Clear Cordova Cache 
and try building again.	
```	

Error troubleshooting link:

[https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(BLD401);k(TargetOS-8.1)&rd=true](https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(BLD401);k(TargetOS-8.1)&rd=true)

digitaldrummerj:
I am working on this

Issue #1091, opened 2016-08-07
Update Mobile App for Current Functionality Requirements


tonysurma:
- [ ] Volunteer has mobile app to view volunteer dashboard (similar to website)
- [ ] Volunteer can see itinerary they are scheduled on next/today
- [ ] For today's itinerary volunteer can view requests, order of request and details of each item
- [ ] Volunteer can mark request complete on today's itinerary
- [ ] Volunteer can mark request incomplete on today's itinerary with note

tonysurma:
- [ ] Volunteer can view and edit profile data on mobile app

digitaldrummerj:
@tonysurma should the list of functionality be split into individual issues?

tonysurma:
thanks @digitaldrummerj

PR #1179, opened 2016-08-27
mobile app converted to true ionic project.

digitaldrummerj:
Removed login page since the login was not used with any of the
functionality.

converted to a true ionic project that can be used either from visual studio or the command line.  

fixes https://github.com/HTBox/allReady/issues/1082
fixes https://github.com/HTBox/allReady/issues/1081
fixes https://github.com/HTBox/allReady/issues/1076

https://github.com/HTBox/allReady/issues/1091


tonysurma:
@digitaldrummerj is this good to go (assuming build works etc.)?  Thanks

tonysurma:
also anything special a VS2015 community edition Update 3 user would need to know to open/work on this? (I assume install cordova stuff that comes when you click cordova in install, but anything special to check/uncheck?)  Thanks again

digitaldrummerj:
The biggest thing is to install the npm packages that ionic requires by running.  All of the Visual Studio stuff is still the same after that.  

```
npm install -g cordova ionic gulp bower
```

digitaldrummerj:
@tonysurma this is good to go.  It has the same functionality as the current mobile app which is to view events, event details, and check in with the QR code scan.  I removed the login screen since none of the existing functionality required it.

Issue #894, opened 2016-06-21
Remove PreferredEmail and PreferredPhoneNumber on TaskSignup/EventSignUp


mgmccarthy:
this results in handlers needing to first check if PreferredEmail/PreferredPhoneNumber is null, and if it is, to include User into the query in order to get at ApplicationUser's Email and PhoneNumber values (which also could be null).

We go through an email vaildation for ApplicationUser email (there is EmailConfirmed field) to validate the user's email, but do we also validate email's/phone numbers added to TaskSignup?

mgmccarthy:
@tonysurma, any time to look at why we're supporting a Preffered Email and Phone number on Tasks when a user is forced to provide both to use when they register with the site? I don't think users can volunteer for a task unless they're registered with the site, correct?

mgmccarthy:
@tonysurma @MisterJames 

the reason I'm asking this is b/c we are not requiring either:
1. that the user confirm their email when registering with the site before they start to volunteer
2. preferred email need to have legit email when volunteering

this is resulting in a lot of code in communication-based handlers (aka, handlers that send email and/or sms texts) further down the pipeline that has to check first if the user's email has a value, then check if the preferred email for the given event has a value, and if neither do, then to silently not send a message.

I good example of this type of check is in NotifyVolunteerForUserUnenrolls. There is a bunch of code that deals with all the checks I just mentioned before getting to the task of building the message and sending it:
![image](https://cloud.githubusercontent.com/assets/2428993/18005214/8df09866-6b66-11e6-9af1-3b240b871098.png)

it would be nice to not have to check for the existence of a user's email. Things like

- not allowing the user to volunteer until they confirm their email with us
- not allowing the user to voluntter if they have not confirmed their email with us by makign Preferred Email a required UI field

could get rid of a lot of this type of "null/whitespace/existance" checking code that has propogated into a lot of our controllers and handlers.

thoughts?

tonysurma:
preferred was a previous concept. High level thoughts:
a) we should require that they complete the email and phone prior to volunteering but to think on workflow there so we don't potentially get user drop off
b) we can drop preferred altogether from UI and model and show current email/phone number when you volunteer and link/allow edits if that has changed since you registered
c) that should simplify it down to you have to have emial & phone to volunteer and the only ones are the ones on your account

mgmccarthy:
my feeback on:

[...]

PR #1182, opened 2016-08-27
Remove AllReadyDataAccess

robrich:


robrich:
and now also AllReadyDataAccess.Tasks

robrich:
and now also AllReadyDataAccess.Request

tonysurma:
@robrich I think this needs a rebase to pick up the unit test fixes like the other PRs.  (sound right @MisterJames ?)

robrich:
and now also AllReadyDataAccess.Event, EventSignup, and ApplicationUser ... which removes AllReadyDataAccess completely.

robrich:
and rebased as well.

MisterJames:
Thanks @robrich!! This is a hyuoooge change and it looks good so far. I'm going to pull down a copy today or tomorrow to run it all locally. I would appreciate it if someone else could take a look as well (perhaps @mgmccarthy or @stevejgordon ?) for another set of eyes.

mgmccarthy:
@MisterJames, I won't have any time tonight, but I will have time tomorrow morning to pull it down and take a look.

stevejgordon:
@MisterJames - I'm back from my trip to Wales. I'll try to pull this down during my lunch break and give it a once-over.

AllReadyApp/Web-App/AllReady/Features/Events/UnregisterEventHandler.cs
```diff
@@ -1,27 +1,42 @@
...
 using AllReady.Models;
 using MediatR;
 
 namespace AllReady.Features.Event
 {
-    public class UnregisterEventHandler : AsyncRequestHandler<UnregisterEvent>
-    {
-        private readonly IAllReadyDataAccess dataAccess;
+    public class UnregisterEventHandler : AsyncRequestHandler<UnregisterEvent> {
+        private readonly AllReadyContext dataContext;
```
stevejgordon: Personal niggle with the naming of the private instance fields. I personally prefer these are prefixed with the underscore. I believe this is the standard convention.

AllReadyApp/Web-App/AllReady/Features/Events/UnregisterEventHandler.cs
```diff
@@ -1,27 +1,42 @@
...
+        public Task DeleteEventAndTaskSignupsAsync(int eventSignupId) {
+            var eventSignup = this.dataContext.EventSignup.SingleOrDefault(c => c.Id == eventSignupId);
+
+            if (eventSignup == null) {
+                return Task.FromResult(0);
+            }
+
+            this.dataContext.EventSignup.Remove(eventSignup);
+
+            this.dataContext.TaskSignups.RemoveRange(this.dataContext.TaskSignups
```
stevejgordon: The use of the "this" keyword feels a little redundant and again, personal preference that it's not used when not needed. Especially if the convention of _dataContext is used instead. It's pretty clear then that you're referring to a private field.

mgmccarthy:
@MisterJames, didn't get eyes on the PR this morning b/c of some migration problems this morning in conjunction with #894. Looks like @stevejgordon got to do a once over. Let me know if you still need another set of eyes.

stevejgordon:
@robrich Great PR and fantastic to see IAllReadyDataAccess biting the dust!

I've added a couple of comments which are more preference than any specific issue with functionality.

In summary:

- Private instance fields should be prefixed with an underscore.
- No need for the "this" keyword usage in a lot of the cases, especially if the field naming is changed.

I've loaded up the site using your code and run through a fair chunk of the functionality. Everything seems to work fine. I found a few bugs with making users site admins etc, but those are present on master too so not related to your changes. I'll triage those a bit more and put in some issues for them.

I've not been able to review all of the code by eye but have scanned over the changes. I do have a small concern around some of the updated Mediatr handlers. In some of them you are calling context.Update passing in the Db model item directly from the message. Whilst this seems okay with the code so far, if anyone uses the handler, sending an incomplete model it could have undesired results including data loss. For example...

UpdateUserHandler expects an ApplicationUser in the message. If someone sends a complete ApplicationUser (including it's related AssociatedSkills), we're okay. But if someone sends a model with just the Id and FirstName then the Update method will still mark all properties as modified. When calling save changes all other properties are then nulled out / reset to defaults. Importantly, related records might get cleared also. In this case we could loose the AssociatedSkills.

I've tended to first query the record using the Id in my handlers and then update the desired properties in the handler. This is still open to issues if the properties are not set on the incoming message, but it is a bit more explicit and shouldn't affect related objects. The downside is an extra DB call. There are ways to specifically mark entity states for the properties which might be a safer middle ground and allow us to avoid the extra call.

Again, this doesn't seem to be a problem as it stands but is something to consider as other contributors start to use these Mediatr handlers.

cc / @MisterJames @mgmccarthy

mgmccarthy:
@stevejgordon is right about losing data. 

In the AllReady code base, when doing an update via a UI, we always read from a GET controller action, which most of the time, is going to return a ViewModel. In some instances that ViewModel could contain a Model as one of its properties, or we could be working directly with the Model itself. 

If that's the case, the Model's properties need to be carried through the UI form (hidden fields, etc...) when the POST method occurs. If a developer does not do this, they're sending the Model back with potentially only some of the properties populated from the initial read. If there's an Id match on that model class in the db, you're going to lose data.

A good rule of thumb with the "AddOrUpdate" approach is:
Unless the add or update is a 1:1 with the table you’re updating in the database, you will lose data. 

but another "rule" to go along with this is to make sure you carrying the fully populated model class through the entire, GET/read, POST/send command life-cycle.

(low importance) If anyone's interested, I recently wrote a blog post about this: https://mgmccarthy.wordpress.com/2016/08/24/entity-framework-addorupdate-is-a-destructive-operation/

as far as the underscores, I'm a fan of no underscores, but since AllReady mainly uses underscores, I'd stick with one style, and that should be underscores in this project's case. Either way, if not fixed, I don't think it's the end of the world.

AllReadyApp/Web-App/AllReady/Features/Events/UnregisterEventHandler.cs
robrich: I left the prefixing of private variables when I left `m_VB`.  Interesting that we differ on preference here.  The solution seems to include many of each technique.  I tried to leave the code using the style it had rather than enforce one convention or the other.

AllReadyApp/Web-App/AllReady/Features/Events/UnregisterEventHandler.cs
robrich: That's cool our preferences differ here.  I prefer `this` as it highlights instantly this isn't a local variable.  I also like to avoid `var` as `int` and `string` are much more descriptive.  I tried to leave the original author's style unchanged where possible, though R# likes to 'help'.

robrich:
@stevejgordon: Interesting that our preferences differ here, and if we're in the land of talking about tabs-and-spaces or curly-brace-placement, I think I definitely hit the mark.

On EF's `AddOrUpdate`:  Agreed.  If using update and not passing in the whole thing, data is lost.  See also https://mgmccarthy.wordpress.com/2016/08/24/entity-framework-addorupdate-is-a-destructive-operation/  I think the test I was modifying is definitely overly simplistic, but it proves that the code in question flows data into the persistence store, and that's the purpose of the test.

stevejgordon:
@robrich We all have our preferences and styles so no worries there. Wasn't sure if the project had a preference for any standard we should be applying.

My concern was around the direct use of the DbContext.Update method in the handlers (eg. UpdateUserHandler) when accepting an EF model directly in the Mediatr command/query message. If someone calls the handler with a partial ApplicationUser object in the message we could loose data.

If we first get the ApplicationUser from the db by Id and then explicitly update properties at least those we don't touch should remain unaffected.

robrich:
Ah, good call.  The method seems rather generic, and shame on me, I haven't traced the calls to it yet.  Do you know which properties need updating through this mechanism?

mgmccarthy:
@stevejgordon @robrich 

I would look at whatever controller is sending the command that invokes `UpdateUserHandler`, then trace how the model coming into the POST action method is populated in the subsequent GET action method. If we're fetching the whole ApplicaitonUser model and making sure we persist that object through the PRG lifecycle, we should be good on the AddOrUpdate.

A command should have one one logical sender and receiver (which makes sense in a CQRS world)... but there is no way to enforce this. There is also no way to enforce that the contributors know the ramifications of AddOrUpdate...

so, IMHO, I'd err on the side of caution, and have code that does a read first, then sends a command. 

It would be very easy to miss an AddOrUpdate "error" during a code review (if there is a code review at all before the PR is merged)...

again, just my two cents. At my current job, we use AddOrUpdate extensively, but all developers understand the potential destructiveness of that operation.

stevejgordon:
@mgmccarthy Just to be clear. That handler is not using the AddOrUpdate extension. It's directly calling update on the context. The danger with the update method is that it marks all reachable entities and properties on the object as modified. Therefore, as you state, if you don't pass in a complete object, including the related navigational entities those may end up being cleared when we call SaveChanges.

I was thinking about it a bit more on my commute home and agree, generally a command/query should be pretty specific to a use case and therefore the use of Update might not be a problem. However we can't prevent someone who perhaps wants to add a way to just update a name (simplified example) from creating a controller and sending an ApplicationUser in the command message with just id, firstname and lastname set. In that case any other properties could be lost.

Perhaps then, it's the command/query message where we start to enforce rules. We could expose the specific fields that must be supplied and even validate the message to ensure they are all set (perhaps in constructor etc). That way we more closely bind the message to it's intended use and force people to send the expected data when creating the message to send. At least then we can assume that if the message add constraints during it's creation that we can be reasonably confident the handler for the message won't be misused?

There seem to be a few possible solutions and perhaps most simply it's validating PR's to ensure they correctly use any handlers. Perhaps we tighten up the message class a little and give it some control to ensure messages contain everything. Perhaps we just need proper xml documentation comments to advise what we expect in messages and the behavior of the related handler. Another thing we can and should do is to use the ChangeTracker to properly mark elements as not modified when we are not expecting them to change. That way we limit the scope of the changes to those properties we truly want to affect. By calling update we do explicitly state we expect all properties to be marked as modified.

I want to re-read this but Julie Lerman wrote a good summary of some of the change tracking behavior changes - https://msdn.microsoft.com/en-us/magazine/mt767693.aspx

I'm not sure though if we need this all decided here within this PR. Maybe we can merge this and then add issues to review what we can do to safety check the handlers. We know that currently these seem to be working as expected. It might be easier to review each one and tweak as required in future PRs.

@MisterJames is more experienced reviewing PRs so perhaps can jump in here also.

mgmccarthy:
@stevejgordon, thanks for clearing that up. Either way, the problem is the same, I agree.

We could look at putting enforcement rules on the message, but I wouldn't want to enforce rules via the constructor, but rather, just rely on attributes like we do for ViewModels/Modles, etc.

The problem with validating messages (whether via constructor and/or attributes) is when you write unit tests, you will have a LOT more work to do setting up messages when you might only need to set one or two properties on a given message (coming or going) in order to do minimum setup on your test.

"Another thing we can and should do is to use the ChangeTracker to properly mark elements as not modified when we are not expecting them to change. That way we limit the scope of the changes to those properties we truly want to affect. "... I'd have to read that Julie Lerman link to get a refresher, b/c I don't know enough about change tracking in EF 7 yet.

I agree, we don't need to decide for this PR, we should merge the PR and we can work the Update stuff afterwards.

IMO, if our code base consistently does a read first, then only updates properties that need to be updated for a given operation, the code alone should be enough inform programmers how update ops should be done. Again, the key here is code consistency, and that's something all the developers on the project should be consistently refining/refactoring/improving.

tonysurma:
I merged #1208 and was going to merge this per comment on that thread but is it ready to go @robrich (certainly builds and is mergable)

robrich:
@tonysurma: I think this PR is ready to go.  There's an outstanding question on using `.Update()`, but we agree that's best for a separate issue.

tonysurma:
Thx @robrich

robrich:
Woo hoo!

PR #1183, opened 2016-08-27
Fixed #1059

### LINKED ISSUE #1059: When viewing campaigns, all dates are followed by 'Central Standard Time'
I assume that it is selecting the time zone based on the logged in user. However, we're currently in Central Daylight Time. 

The user selects their time zone on the 'Manage your account' page:
usertimezone

When viewing a campaign, could we just display the time zone of the user without adding 'Standard' or 'Daylight'? Otherwise it can be confusing, especially if a campaign overlaps CDT and CST such as:
campaigntimezone

Version: 1.0.0.0 ddb2198

kgipson:
Updated time zone diplay text formatting.


MisterJames:
Nice, thanks @kgipson. :shipit:

PR #1185, opened 2016-08-27
Issue #1184 added missing fields

### LINKED ISSUE #1184: Organization registration fails
Just try and 'Register as an Organization administrator?' Not sure if this is a feature that has not been implemented yet, I could not find a previous issue for it. So its asking for first name, last name, and phone number but the inputs for them are missing. 

sketch

Daven-:
Just added missing input fields that where required by organization registration Issue fixes #1184


robrich:
:+1:

PR #1186, opened 2016-08-27
Issue 519 Fixed 

NotTheNull:
Added check to Campaign Event to make sure that the Start Date is not… set prior to the Campaign's Start Date.  Also added a default End Date of 30 days from the Start Date.

A couple of questions/concerns that I had on the matter:

1. Should an error occur if the admin selects an Activity Start Date that exceeds the Campaign End Date?  Similarly, should an error occur if the admin selects an Activity End Date that exceeds the Campaign End Date?

2. If the defaulted Activity End Date exceeds the Campaign End Date, should the Campaign End Date be used instead?


MisterJames:
@NotTheNull can you rebase this for us?

PR #1188, opened 2016-08-27
Constrain date selection for tasks to within event date

### LINKED ISSUE #1022: Constrain choice of End Date on Task creation UI to the end date of the Event
When going to create a new Task, the user has to choose a Start Date and an End Date:
image

The user is left to pick the start and end date on their own that looks like it defaults to DateTime.UtcNow or DateTimeOffset.UtcNow?

Either way, the user is allowed to pick any end date using the calendar control
image

only after being allowed to choose an end date that is after the end date of the campaign and clicking save does the user receive and error that the "End date cannot be later than the event end date 12/31/2015"
image

Couple things we can do:
1. put the Event's end date on the UI so the user knows what the date range is that's acceptable without having to rely on an error to tell them.
2. even better, put default dates in the calendar for start and end date that are both within the range of the Event's end date. This is more work, but will guide the user to make the right choice by not allowing them to choose an start of end date outside the allowable range.

zimplicity:
Fixes #1022

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Views/Shared/_Layout.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Views/Shared/_Layout.cshtml
@@ -149,6 +149,13 @@
                 toolbarPlacement: 'bottom'
             });
             $(".datetimepicker").each(function (index, element) {
+                if ($(element).attr("min")) {
+                    $(element).data("DateTimePicker").minDate($(element).attr("min"));
+                }
+                if ($(element).attr("max")) {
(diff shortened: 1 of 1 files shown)


MisterJames:
@zimplicity can you rebase this for us? It should fix the build error when you do a new commit.

zimplicity:
Done!

tonysurma:
@MisterJames now that is rebased are we good to go?

tonysurma:
pinging @MisterJames

Issue #1097, opened 2016-08-07
Mobile App - create a login page within the mobile app


digitaldrummerj:
Currently the mobile app uses an in-app-browser to launch the mobile web sites login page.  This does not provide the best user experience as the user is then able to navigate around the mobile web page and is not longer really uses the mobile application.

digitaldrummerj:
I am working on this

digitaldrummerj:
having CORS issues with the code that is currently commited when running the mobile app in a web browser vs on a device.  Need to implement the ionic proxies solution to fix that issue.  [http://blog.ionic.io/handling-cors-issues-in-ionic/](http://blog.ionic.io/handling-cors-issues-in-ionic/)

tonysurma:
still open @digitaldrummerj ?

digitaldrummerj:
Yes still open.  I haven't been able to get back to it.  Right now it only supports email/password and doesn't support the social logins.

BillWagner:
@digitaldrummerj If there's work in progress, can you share it? We're at an event in London, and some folks are looking to help on this.

jonparker:
I think we can assume that any work done on this will need to be redone after #2031 
I am happy to work on this ticket basing it off the work @SteveKennedy has done on #2031.

PR #1199, opened 2016-08-30
Fixed CORS And Options for Login 

digitaldrummerj:
when calling the MeController.Login method from Angular it was erroring
for PreFlight Options.  Had to move the CORS configuration to the start of the
service configuration to fix this issue

https://github.com/HTBox/allReady/issues/1097


tonysurma:
I am going to pull this in so as to unblock any of the mobile app work but of course @stevejgordon @MisterJames let me know if you have any review comments for follow up issues

PR #1204, opened 2016-08-30
Adding knockout mapping via bower.json + updating path

### LINKED ISSUE #1203: Unable to volunteer for task - Missing knockout-mapping.js
The volunteer modal is not displaying due to missing knockout-mapping.js file.

This appears to have been removed in PR #1195

stevejgordon:
Fixes #1203 - knockout-mapping was missing which appeared to break the volunteer button.

@robrich - I think this was removed in your PR #1195. Let me know if this was for a specific reason? I'm not very familiar with bower so not sure if my fix is the correct approach or if there's a better way.


robrich:
@stevejgordon: I noticed a few files got committed to libs, so I deleted and excluded them.  This is the perfect approach.  Fail on me for not re-testing the page.  Well done for catching it.

Issue #910, opened 2016-06-23
Create/Edit campaign - Copy contact errors if user is site admin


stevejgordon:
We should hide this option for a site admin

stevejgordon:
@tonysurma Looking at this very briefly, shouldn't this still work for site admins, as the functionality would be to copy from the event's owning organization contact I believe?

I haven't seen or used this feature before so I may be misunderstanding the intended functionality.

tonysurma:
site admin = admin of all of allready (so not an option)
org admin = admin of an org

so in short should work and be allowed for everyone except for site admins

stevejgordon:
@tonysurma What should actually happen when this button is used?

tonysurma:
when you are an org admin then the contact info for your org should be copied into the campaign (like a nice starting point) if you are a site admin this button should not appear as it doesn't make logical sense etc.

Daven-:
working on this

Daven-:
Leaving this for someone else, error still present when logged in as an organization.

dchristensen:
I was going to try and take a look at this one, but it looks like it is functioning properly.  For site admin the button is not visible.  For org admin the button is visible and is correctly filling the contact information.  I believe this can be closed.

PR #1207, opened 2016-08-31
(#910) hides the copy contact info button

Daven-:
Hides it when logged in as site admin, error still present when logged
in as an organization so the issue is not completely fixed. Will leave
for someone else.


tonysurma:
@stevejgordon or @MisterJames can you take a review?

PR #980, opened 2016-06-29
WIP: #919 remove volunteers from event

VishalMadhvani:

PR #1146, opened 2016-08-18
Remove GetEventSignups(string userId) from IAllReadyDataAccess

mgmccarthy:
Fixes #1145

PR #1208, opened 2016-08-31
Removing Event Volunteer Limit

### LINKED ISSUE #919: Events - Remove concept of volunteers req at event level
Only tasks should require input of the volunteer amounts required.

stevejgordon:
Fixes #919 

After discussion with @tonysurma this PR is limited to removing the UI and model properties for the concept of a volunteer limit on events. Events don't allow direct signup and all signups are at the task level. Each task will maintain it's own limit.

The event UI for enforcing volunteer limits remains and applies to all tasks belonging to that event. Therefore if this is enabled for the event, all tasks will enforce a signup limit as per their own volunteer limit property.

There is also the concept on the model for allowing a waiting list. However this appears incomplete and the UI is not implemented. I have left the property on the model which defaults to false until this is completed.

The event details view model and view no longer expose a list of volunteers for an event. We can later review adding this back by building the volunteer list from all of the task signups.

Most importantly we still have code left in place which updates an EventSignups table. I started work on removing this table but it started to touch a lot of areas in the code, including notifications. On reflection it requires some thought around how to best update the notifications to work based purely on the task signups and to ensure we don't loose functionality which currently is underpinned by this table.

I do feel that eventual removal of the EventSignups table is the right move since we should be able to work out event level aggregated list of volunteers from the associated tasks. It feels redundant to maintain this longer term.

I suggest this is merged and once we have #1182 also merged we can begin work on steps to remove the EventSignup table.

NOTE: @VishalMadhvani had completed a WIP #980 around this issue.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Campaigns/CampaignDetailQueryHandler.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Campaigns/CampaignDetailQueryHandler.cs
@@ -11,7 +11,7 @@ namespace AllReady.Areas.Admin.Features.Campaigns
 {
     public class CampaignDetailQueryHandler : IAsyncRequestHandler<CampaignDetailQuery, CampaignDetailViewModel>
     {
-        private AllReadyContext _context;
+        private readonly AllReadyContext _context;
 
         public CampaignDetailQueryHandler(AllReadyContext context)
         {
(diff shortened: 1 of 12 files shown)


mgmccarthy:
@stevejgordon @tonysurma @MisterJames 

I'd like to wait on implementing  #894 until the EventSignup table is gone. Removing this unneeded table should make implementing #894 much easier, and I won't end up writing code that is going to eventually deleted.

As far as PR merge order, I have one outstanding PR that removes items from IAllReadyDataAccess, PR #1146. I'm not too sure if @robrich's PR's took this into account when he totally obliterated IAllReadyDataAccess, but if it did, then my PR is probably useless.

robrich:
@mgmccarthy: I don't think I accounted for #1146 when I did #1182.

mgmccarthy:
then we should be delete/ignore PR #1146 and just merge your IAllReadyDataAccess PR's @robrich

stevejgordon:
@mgmccarthy I think then steps are to get this and @robrich's #1182 merged. Then I can tackle taking out the EventsSignup table and sorting out the knock-on effects of that.

mgmccarthy:
@stevejgordon, let me know if you need any help with anything... in the meantime, there are other PR's I'm working on so no rush.

tonysurma:
merged this and will merge #1182 when confirmed

PR #1209, opened 2016-08-31
Delete existing image on Campaign and Event when uploading new image via Edit page

### LINKED ISSUE #338: There's currently no way for admins to remove an image from Campaign, Event

mgmccarthy:
(kind of) Fixes #338


tonysurma:
@MisterJames @stevejgordon can we get a review and then I will merge?

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Event/Edit.cshtml
```diff
@@ -49,6 +49,11 @@
         <hr />
         <div class="form-horizontal col-lg-6 pull-left">
             <div asp-validation-summary="All" class="text-danger"></div>
+            @if (isEdit)
+            {
+                <input type="hidden" asp-for="Id" />
+                <input type="hidden" asp-for="ImageUrl" />
+            }
             <input asp-for="Id" type="hidden" />
```
stevejgordon: You have the hidden id input both inside and outside the if block here, so if in edit mode it would be duplicated currently. Looks like we can remove the line outside of the if block since we should only have an Id to maintain when editing an event.

stevejgordon:
@tonysurma I've given it a bit of a look over this morning before work and it looks close to me. One comment above to update one of the views. One comment on logical order of the delete.

I think it might need a rebase though. It's showing no conflicts but seems to put back the event level volunteers required input that my prior PR removed.

I haven't personally set up the Azure bits to test uploads etc, but if @mgmccarthy has smoke tested that I don't see an issue. The delete image code looks similar to stuff I've had to write before to work with blob storage.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
```diff
@@ -117,11 +119,16 @@ public IActionResult Create()
                         ModelState.AddModelError("ImageUrl", "You must upload a valid image file for the logo (.jpg, .png, .gif)");
                         return View(campaign);
                     }
+
+                    if (existingImageUrl != null)
+                    {
+                        await _imageService.DeleteImageAsync(existingImageUrl);
+                    }
                 }
```
stevejgordon: We should probably enclose this delete code block within the if (fileUpload.IsAcceptableImageContentType()) and only on successfully getting a new imageUrl back from the imageService? If for any reason that fails to upload or is not the right content type we don't want to delete the existing image. Currently I believe this will fall through and delete the original image regardless of whether a new one is correctly uploaded.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventAdminController.cs
```diff
@@ -185,6 +187,11 @@ private string GenerateItineraryDetailsTemplateUrl()
                         ModelState.AddModelError("ImageUrl", "You must upload a valid image file for the logo (.jpg, .png, .gif)");
                         return View(campaignEvent);
                     }
+
+                    if (existingImageUrl != null)
+                    {
+                        await _imageService.DeleteImageAsync(existingImageUrl);
+                    }
                 }
```
stevejgordon: Same comment as above regarding only deleting the existing image if we maybe had a success in the imageService during the new image upload (and get a new imageUrl back)

mgmccarthy:
@stevejgordon, thanks for the code review, you caught a couple things here that I missed, so thanks for taking your time to read through the commit. I'm working on changes now. I will also be rebasing once the I've updated everything locally so I don't blow away your event level volunteer deletion work you did

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
mgmccarthy: Steve, how does this look?
```
var existingImageUrl = campaign.ImageUrl;
if (fileUpload.IsAcceptableImageContentType())
{
     campaign.ImageUrl = await _imageService.UploadCampaignImageAsync(campaign.OrganizationId, campaign.Id, fileUpload);
    if (campaign.ImageUrl != null && existingImageUrl != null)
    {
        await _imageService.DeleteImageAsync(existingImageUrl);
    }
}
```

we're checking two things before invoking delete:
1. if there is an existing ImageUrl for the Campaign/Event, we have to capture that in order to know which image to delete after we've uploaded the new image.
2. the check you recommended in the code review, that the campaign.ImageUrl is not null, meaning that UploadCampaignImageAsync was successful.

stevejgordon: Pretty much, although I think we need to store the result of the imageUpload call to a local var and check that for null/empty. If it has a value we can then update the campaignEvent.ImageUrl and perform the delete. With the above, if the upload failed here we'd be setting the imageUrl in the db to null so effectively still removing the image from the event.

Something like this...

```
if (fileUpload.IsAcceptableImageContentType())
{
    var existingImageUrl = campaignEvent.ImageUrl;
    var newImageUrl = await _imageService.UploadEventImageAsync(campaign.OrganizationId, campaignEvent.Id, fileUpload);
    if (!string.IsNullOrEmpty(newImageUrl))
	{
		campaignEvent.ImageUrl = newImageUrl;
		if (existingImageUrl != null)
		{
			await _imageService.DeleteImageAsync(existingImageUrl);
		}
	}
}
```

mgmccarthy:
@tonysurma, please **do not** merge this code yet. I have one more fix to make based on Steve's code review... but I have to cut it short this morning. I'll hopefully have this wrapped up by tonight or tomorrow at the latest

mgmccarthy:
@stevejgordon, when you have a chance, I just made the last fixes and rebased. Let @tonysurma know when you think it's good to merge. Thanks!

stevejgordon:
@mgmccarthy looks good.

@tonysurma :shipit:

PR #1231, opened 2016-09-12
Fix #1230 Change result testing to current culture

### LINKED ISSUE #1230: UnitTest.Areas.Admin.ViewModels Percentage test failed
All UnitTest.Areas.Admin.ViewModels Percentage test failed on my system as it's use "." decimal separator instead of NumberFormat.NumberDecimalSeparator in assertment

result.ShouldBe("0.0");  failed in cultures with comma  decimal separators

achasik:
All UnitTest.Areas.Admin.ViewModels Percentage test failed on my system as it's use "." decimal separator instead of NumberFormat.NumberDecimalSeparator in assertment

result.ShouldBe("0.0"); failed in cultures with comma decimal separators

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/ViewModels/EventDetailViewModelTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/ViewModels/EventDetailViewModelTests.cs
@@ -1,4 +1,5 @@
-﻿using AllReady.Areas.Admin.ViewModels.Event;
+﻿using System.Globalization;
+using AllReady.Areas.Admin.ViewModels.Event;
 using Shouldly;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


mgmccarthy:
@tonysurma, is culture and globalization something we need to worry about yet, or is it on the technical backlog?

stevejgordon:
@mgmccarthy We need to ensure that at least the tests function correctly invariant of the culture of the machine running them. I've never had to think about it before!

tonysurma:
@mgmccarthy - @stevejgordon is right.  Not a concern today, will be someday so need to not make the later impossible and certainly need to make sure tests can run

tonysurma:
I am going to merge this and then we should see if we can find a way to run tests 'independent' of current culture so each test doesn't need to specify it explicitly (until we go multiculture/language and then we will deal with it explicitly then)

PR #1238, opened 2016-09-15
Seed all ready task with number of volunteers required in sample data generator

### LINKED ISSUE #1232: unable to volunteer for any tasks using default seeded data on the master branch...
I remember being able to log is as User@example.com1.com, go to a Campaign, pick an Event, and start to volunteer for Tasks...

but now every Task in every Event for every Campaign (all seeded data) does not allow a volunteer:
image

Is this a result of removing the ability to volunteer for an Event or something else?

mgmccarthy:
Fixes #1232


mgmccarthy:
thanks for the review @stevejgordon

Issue #1031, opened 2016-07-30
Building Itineraries in allReady


OhMcGoo:
This is my attempt to describe the Itinerary creation process including needed features. In a separate issue, I will detail Rally Event creation.

As requests for services are provided to allReady (or created within the application), organizational admins should have the ability to group these requests into itineraries for fulfillment.

**Requests**
A request is received from a third party application (like [getasmokealarm.org](https://getasmokealarm.org)) OR a request is manually entered into allReady. If manually entered, a form is accessed that records basic information: first name, last name, street address, city, state, Zip Code, phone (including a way of acknowledging that the number accept text messages) and e-mail address. The request is tagged as unfulfilled or new.

An itinerary builder page should dispaly **Details** (including who will be fulfilling the itinerary), a **Map** and a **Listing** of addresses (that make up the itinerary).

**Details**
Fields should allow an admin to:

- Name the itinerary (an itinerary number should be automatically assigned).
- Enter a date and start time
- Enter the number and type of volunteers required to fulfill the request 
- Assign volunteers to the itinerary

**Map**

- Requests are displayed on a map with icons that denote their status.
- The map should have basic navigation controls: zoom in, zoom out and pan.
- Other status icons - besides new - should include: scheduled, cancelled, missed and completed.
- An admin should have the ability to turn icons off and on by status as well as turn all icons “on” or “off.”
- An admin should be able to select icons by clicking (and shift-clicking) or by lassoing.

**Listing**

- Icons can be dragged and dropped into a pane next to the map for listing requests. When dropped, each requestor’s details should be listed, the address being most prominent. A pop-up menu with status should display next to the address.
- In addition, the distance and travel time between each request should be calculated and displayed.
- Outside the itinerary pane should be the following:
1. Fields to enter a starting address
2. Fields to enter a finish address (and a way of denoting if the start and finish are the same)
3. A button that when clicked will optimize routing and reorder the addresses accordingly
4. A display of the total mileage and travel time.

A Save itinerary button should be displayed prominently. When an itinerary is saved, volunteer opportunities should be created.

I hope I have provided sufficient detail to permit creation of individual work items from this large issue. I welcome all questions and will do my best to provide clarifications.

Thanks,

Jim

PR #1250, opened 2016-09-20
Adding manual creation/edit of requests

### LINKED ISSUE #1227: Add Request Page
Per feedback from @OhMcGoo in #1031 we need a way to manually add a request into the allReady system.

Also relates to this features backlog item #900

### LINKED ISSUE #1251: Edit Request Page
Following on from #1227 we will need the ability to modify details of a request manually.

Relates to #1031

stevejgordon:
Fixes #1227 and fixes #1251 

Added initial code to allow the manual creation of a request in the allReady application.

This is accessed from the requests list page for an event by clicking the create request button.

PR updated to include the ability to edit requests.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/RequestControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/RequestControllerTests.cs
@@ -138,7 +138,7 @@ namespace AllReady.UnitTest.Areas.Admin.Controllers
         }
 
         [Fact]
-        public void Edit_HasHttpPostAttribute()
+        public void EditPost_HasHttpPostAttribute()
         {
             var sut = new RequestController(null);
             var attribute = sut.GetAttributesOn(x => x.Edit(It.IsAny<EditRequestViewModel>())).OfType<HttpPostAttribute>().SingleOrDefault();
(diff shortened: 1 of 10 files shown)


mgmccarthy:
@stevejgordon, just eyeballed this PR. I have a question about Requests... basically, what are they? #1031 doesn't really go into what a Request is.

My guess is that it's a way for people to request assistance in some way from AllReady (the example in #1031 is a request for a smoke alaram(s))....

Can anyone make a request to AllReady for something, or do they at least need to be a registered user? (I'm assuming they need to be registered)

stevejgordon:
@mgmccarthy Requests are essentially a unit of work against an event (specifically an Itinerary event). In the GetASmokeAlarm example, each install would be a request. In that scenario we expect most of these to be populated via API calls from the GetASmokeAlarm.org website (that is the user facing portal where members of the public can request a free smoke alarm install). There is no requirement at this time for a public user to be able to make a request directly in allReady.

The implementation prior to this PR only allowed API generated requests coming from other systems and no way for an org or allReady admin to manage the manually. In #1031 the requirement was confirmed that an org or allReady admin should be able to create and update requests. This is likely to be used less often from my current understanding but supports cases where something does need to be added manually.

Itinerary events can have 0 or more requests linked to them. Org admins can then create itineraries and assign requests to them. This allows the generation of sets of work that a team of volunteers can undertake. Requests can then be marked as completed (I don't think we have this yet as need to define if volunteers can do this and via what method).

Hopefully that makes some sense?

mgmccarthy:
@stevejgordon, makes perfect sense. Thanks for explaining it. Not being able to make standups is killing me, but I don't see that changing anytime soon as long as they're mid-day on a Saturday

Issue #1234, opened 2016-09-15
.NET Core 1.0.1, MVC 1.0.1 and EF 1.0.1


stevejgordon:
Patch releases of the above have been made my Microsoft.

https://blogs.msdn.microsoft.com/webdev/2016/09/13/asp-net-core-sept-2016-patch/

Suggest we upgrade the project to use these versions. I have successfully done this on some projects where I work. There is an update to install 1.0.1 on dev machines that contributors may need to do. I've done it on mine but not sure if it's strictly required or not.

@tonysurma Suggest we plan this in, however not sure if we can validate that Azure support this prior to moving ahead? In theory since it's just Nuget packages that should be fine but I'm a little confused in the blog posts about the required to install an updated SDK.

Either, way, I'm happy to do this and put in a PR so someone without an updated SDK could try it?

stevejgordon:
Fix merged in from #1252

PR #1252, opened 2016-09-20
Upgrading to new 1.0.1 versions of microsoft components

stevejgordon:
Addresses #1234 if we agree we are ready for this. It is probably good practice since this includes a fix to a security advisory (https://github.com/aspnet/Announcements/issues/203)

Since I have 1.0.1 installed on my dev PC for other projects I have had issues starting allReady. I upgraded the dependencies to this new version to solve my local issue. Since I have done this I thought I may as well put in a PR whilst we discuss this change.

@MisterJames - Welcome your input on whether moving formally to 1.0.1 is okay for the project.

I may have to base my current work on this version in any case since I cannot run the code unless it targets 1.0.1. I'll probably work with the 1.0.1 versions and rebase before the PRs. I've commented on https://github.com/dotnet/core/issues/267 to see if this is a known issue.


tonysurma:
@MisterJames can you agree/disagree quickly? I wanted to get us on this ASAP so we are all building on the most recent build and don't start requiring people's machines to be NOT upgraded.  Thanks @stevejgordon

Issue #1262, opened 2016-09-23
Mediatr Handler Naming (Discussion)


stevejgordon:
I think this is mostly my fault as I think I started the trend! I wanted to open a discussion around naming handlers as we start to think about documenting some of the allReady coding conventions.

Currently most handler classes which implement the IAsyncRequestHandler are named with the suffix Async e.g. EditCampaignCommandHandlerAsync

While async methods should be named in this way (based on some guidance I've seen) I'm not sure it makes much sense to explicitly name the handler class this way In my example EditCampaignCommandHandler would be sufficient. 

I'm wondering what other contributors think about revising these going forward?

cc/ @mgmccarthy @MisterJames @dpaquette

mgmccarthy:
@stevejgordon I agree. Although "convention" says to name anything that is aysnc with the word "Async" in the name of the class, I don't think it makes much sense. It's easy to see that a class is async based on the API of the class, and in the case of MediatR, there are explicit interfaces that are implemented that already have the word "Async" in them.

I'd like to remove all "Async" works from class names... I think it would make everything easier to read. If someone tries to code against an async class and does not use async/await in their consuming code, the compiler will let them know.

MisterJames:
There are all kinds of classes in .NET that have async methods and do not adorn the name with async. I am 100% behind not having it in the name of the class.

PR #1270, opened 2016-09-25
Remove use of async naming in Mediatr handlers and messages

### LINKED ISSUE #1267: Remove redundant use of Async on handler class names
Following on from discussion in #1262 we should remove the Async suffix from class names in handlers to clean up the code noise.

stevejgordon:
Fixes #1267 

After agreement in discussion #1262 that we should remove the use of async in the naming of our handler and message classes I have updated all occurrences in the code.

@tonysurma This will be a nightmare to rebase as it touches a lot of files (202). If you're happy with it and able to merge in at the earliest that would be great! 😄


mgmccarthy:
@tonysurma, I agree with @stevejgordon, please merge this first!

PR #1272, opened 2016-09-26
Refactor delete methods on event admin controller

### LINKED ISSUE #1271: Refactor Delete Methods on EventAdminControlller
and some refactoring on EventAdminController. All refactorings are aimed at the combination of the following:

- aligning each GET/POST action method to have its own message/handler the retrieves just the information it needs (aka, stop sharing ViewModels across the same controller or multiple controllers)
- eliminating db calls when we don't need to make them

mgmccarthy:
Fixes #1271


mgmccarthy:
@tonysurma @stevejgordon rebased and ready for merge

PR #1277, opened 2016-09-27
Remove ApplicationUser from ShowEventQuery

### LINKED ISSUE #1276: Remove ApplicationUser from ShowEventQuery
ShowEventQuery currently passes the entire ApplicationUser on the message only to grab the user's id in the ShowEventQuery handler.

Refactor ShowEventQuery to carry UserId instead of ApplicationUser, which should be retrieved by EventController before the message is sent.

Refactor ShowEventQueryHandler to use the UserId on the message to do its work

mgmccarthy:
Fixes #1276

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/EventController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/EventController.cs
@@ -1,6 +1,7 @@
 using System.Threading.Tasks;
 using AllReady.Features.Events;
 using AllReady.Models;
+using AllReady.Security;
 using MediatR;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Identity;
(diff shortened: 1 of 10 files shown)


AllReadyApp/Web-App/AllReady/Controllers/EventController.cs
```diff
@@ -46,7 +51,7 @@ public IActionResult Index()
         [AllowAnonymous]
         public async Task<IActionResult> ShowEvent(int id)
         {
-            var viewModel = await _mediator.SendAsync(new ShowEventQueryAsync { EventId = id, User = User });
+            var viewModel = await _mediator.SendAsync(new ShowEventQuery { EventId = id, UserId = await User.GetUserId(_userManager) });
```
stevejgordon: I'm not sure you need your GetUserId extension. The userManager has a GetUserId taking a ClaimsPrinciple which I think you could use e.g. _userManager.GetUserId(User);

AllReadyApp/Web-App/AllReady/Security/ClaimsExtensions.cs
```diff
@@ -1,13 +1,20 @@
...
 using System.Security.Claims;
+using System.Threading.Tasks;
 using AllReady.Models;
 using Microsoft.AspNetCore.Identity;
 
 namespace AllReady.Security
 {
     public static class ClaimsExtensions
     {
+        public static async Task<string> GetUserId(this ClaimsPrincipal claimsPrincipal, UserManager<ApplicationUser> userManager)
```
stevejgordon: As per earlier comment, not sure if we need this. Instead use the UserManager.HetUserId method?

mgmccarthy: @stevejgordon, you're right, we don't need it. I added it as a way for contributors to quickly be able to get the current user's user id w/out knowledge of user manager.... this might, or might not be a good thing.

For now, I'll remove the extension method and just reach out to UserManager directly in the code. 

Thanks for the feedback

PR #1278, opened 2016-09-27
Refactor assigntaskcmdhandler

### LINKED ISSUE #1256: Refactor AssignTaskCommandHandlerAsync
- The queries needs to be changed to async to make async all the way. 
- The handler to assign task also has the responsibility to send notifications to the Phone Number Confirmed and email confirmed recipients. I propose to move the notification responsibility to its own handler. The AssignTaskCommandHandler will be just reponsible for assigning task to the volunteers and publish an event that new volunteers have signed up. This would make the classes and corresponding tests smaller and simple. 

@tonysurma @mgmccarthy let me know what you guys think. I can work on this.

sanilpaul:
Fixes #1256


tonysurma:
@mgmccarthy and @stevejgordon or @MisterJames can you review?

Also we did a big merge of renames from async so looks like this will need a rebase as well @sanilpaul - thanks in advance

AllReadyApp/Web-App/AllReady/Features/Notifications/NotifyAssignedVolunteersForTheTask.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public NotifyAssignedVolunteersForTheTask(AllReadyContext context, IMediator mediator)
+        {
+            _context = context;
+            _mediator = mediator;
+        }
+
+        public async Task Handle(TaskAssignedToVolunteersNotification notification)
+        {
+            var users = _context.Users.Where(x => notification.NewlyAssignedVolunteers.Contains(x.Id)).ToList();
```
stevejgordon: Can you use the ToListAsync method here to make this db call Async?

sanilpaul: @stevejgordon Good catch. will do. I don't know the business reasoning on unassigning users.

stevejgordon: @sanilpaul No worries. As you've stuck with the original behaviour I'd say that once you're able to tweak this line this is good to go. We can then review where and how this is used if required.

sanilpaul: @stevejgordon Thanks. I have pushed that change

stevejgordon:
@sanilpaul Thanks for the quick fix.

@tonysurma I pulled this in before work this morning and it looks good so I say from a code perspective we can :shipit:

**_However_**.., I took a look into where this gets used and as far as I could see it no longer does since we removed the concept of volunteering for events. The only place sending an AssignTaskCommand is the admin TaskController in the Assign action. This previously allowed admins to select event volunteers and assign them to tasks (if I recall correctly).

We now only allow volunteers to signup to tasks directly so there is no concept of an admin assigning someone, since there is no pool of volunteers to assign from. With the current flow I can't see how an admin would ever assign someone to a task, unless we produce a list of all volunteers (against the all tasks for an event) and allow admins to assign/reassign people from that?

The code is not hurting anyone and if we might later need some way for admins to assign people to tasks we could perhaps leave this in. If not though, perhaps we actually strip this out and remove the TaskController Assign method also? Otherwise we're maintaining code that is not doing anything.

mgmccarthy:
@stevejgordon if the code is not used, we should remove it. We always have source control where we can go back and get it if we need it. Nice catch.

tonysurma:
Going to merge now and we can create an issue on the backlog for the code that can be removed

PR #1279, opened 2016-10-01
adding sanitized sample data

MisterJames:
Just a sample file that can be used with the import facilities at http://localhost:48408/Admin/Import when testing.

Note that you must be logged in as a user that you assign an API key to (they need to be an API user to import the data).


dpaquette:
Tested the file and this works after configuring the Geocoding service as per PR #1280

PR #1327, opened 2016-10-05
Initial fix to allow event details to load for unauthenticated users

### LINKED ISSUE #1320: Found issue as "An unhandled exception occurred while processing the request."
When click on link "Smoke Detector Installation and Testing-South Loop" in campaign "Working Smoke Detectors Save Lives" .More details "NullReferenceException: Object reference not set to an instance of an object."

stevejgordon:
Fixes #1320

NOTE: May need more work to improve it further but this is a fix to allow testing work to progress. I'll open a new issue to review this code in more detail.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/EventController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/EventController.cs
@@ -51,7 +51,7 @@ namespace AllReady.Controllers
         public async Task<IActionResult> ShowEvent(int id)
         {
             var user = await _userManager.GetUserAsync(User);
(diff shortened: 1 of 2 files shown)


mgmccarthy:
@stevejgordon, what would probably be helpful is a list of pages (controllers and their action methods as well) that an unauthenticated user can look at. Maybe a requirement of some sort so contributors know when to use nullable values.

mgmccarthy:
@stevejgordon, actually, now that I'm looking at some of these controllers, it would be easy to make an educated guess for which pages are allowed to be viewed by an unauthenticated user based on the [Authorize] attribute at the controller/action method level, as well as any [AllowAnonymous] attributes.

PR #1390, opened 2016-10-21
Fixes #1385

### LINKED ISSUE #1385: Move the Create Request button to the Event/Details page
cc @stevejgordon 

it's not very clear to the user how to create a new request, as the button is hidden a bit. Right now to create a requst, you need to navigate to the event details page:
image

where you can see the all the requests associated with that event. You then need to click on one of the event status squares to bring you to the event requests page:
image

this is where the create request button lives now.

I propose moving the Create Request button to the event details page, so it's one less click to create an event, and the button is a little easier to find for users. Since we're presenting Request information on the event details page in the form of the blocks that represent status, having the Create Request button there makes sense.

maguir:
Fixes #1385

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Event/Details.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Event/Details.cshtml
@@ -168,8 +168,14 @@
             <div class="row">
                 <div class="col-md-12">
                     <h3>Requests</h3>
-                    &nbsp;
                 </div>
+            </div>
+
+            <div class="row">
+                <div class="col-md-12">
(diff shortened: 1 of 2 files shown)


MisterJames:
Thanks @maguir! :shipit:

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Event/Details.cshtml
```diff
@@ -168,11 +168,17 @@
             <div class="row">
                 <div class="col-md-12">
                     <h3>Requests</h3>
-                    &nbsp;
                 </div>
             </div>
 
             <div class="row">
+                <div class="col-md-12">
+                    <button type="button" asp-action="Create" asp-controller="Request" asp-area="Admin" asp-route-eventId="@Model.Id" class="btn btn-default">Create Request</button>
```
dpaquette: The link tag helper doesn't target the button element. This needs to be an <a> tag. I have a fix incoming

PR #1394, opened 2016-10-22
Move models and web jobs to .NET Core projects

### LINKED ISSUE #1392: Convert Models and WebJob to .NET Core
Leaving behind the history of csproj so we can move to project.json so we can eventually move to csproj.

MisterJames:
Looks like the build script is working now in AppVeyor, so we should be good to close this one out.

 - created new DLL and WebJob projects
 - updated project references/cleaned up dependencies
 - fixed namespace references
 - removed `wrap` concepts for old csproj wrapper

Fixes #1392.


MisterJames:
@tonysurma one thing to check when you get a chance is the web jobs logs from the scm console, just to make sure they are running. We may need a cmd file in there to run the exe.

PR #1399, opened 2016-10-26
Adding request source to Request model

### LINKED ISSUE #1315: Store Request Source
Since we now have request that can be manually created, imported from CSV and also pushed into our API it would be useful to store the source of the creation of the request.

We might need conditional logic based on the source. For example, for any manually created requests I presume we might not need to send status updates back to getasmokealarm?

stevejgordon:
Fixes #1315 

Includes:
- model changes
- db migration
- updated commands to set appropriate source value (csv, manual, api)


tonysurma:
Thanks @stevejgordon I am going to merge in the big PR from @mgmccarthy and then pull this in if it doesn't have conflicts

stevejgordon:
@tonysurma No problem. I'll get this rebased at lunchtime.

stevejgordon:
@tonysurma Rebased

mgmccarthy:
@stevejgordon any questions about how the rebase effected the Request API/scheduling code?

stevejgordon:
@mgmccarthy I didn't look too deeply but did take your RequestApiController and updated the AddRequestCommand message to take the source value for the API. I need to dig in a little to understand all of the changes you've made :smile: 

One side note there was the requirement to manually create the Hangfire empty db as it blew up initially. Is it now a requirement to have that or could we add some logic to still function without the db when in dev?

mgmccarthy:
@stevejgordon, first, thanks for updating the AddRequestCommand to make sure the source value of API.

I talked to @MisterJames a couple weeks ago about possibly pulling all Request creation into a single command/handler, but for now, he says it's fine keeping it the way we have now which is:
- UI
- Csv import
- API

Moving onto the db issue, I was just thinking about that. The developer pulling this down and trying to run it should not have to worry about the local db being created, we should somehow create it for them.

Since the Hangfire db is not under the controller of Entity Framework, I am going to try to add something to Startup that will allow me to query for the existence of the the Hangfire db and if it doesn't exist, create it on the developers local machine...

either way, Hangfire needs a way to store all queued jobs (Enqueued, Scheduled, etc...) into some type of persistence... by default, it uses sql server.

There are other persistence stores we can use, which are available via NuGet package add-ons. For example, one package would allow us to use MSMQ for that. In speaking with @dpaquette shortly on the Hangfire implementation, he recommended sticking with the "default" queue/storage for now.

Also, if you get a chance, I just created Issue #1405 which is what needs to happen next for confirming/canceling a request we've received with the API that has been added to a Itinerary. I'm looking for feedback b/c I'm not too savvy with Twilio.

AllReadyApp/Web-App/AllReady/Features/Requests/AddRequestCommandHandler.cs
```diff
@@ -42,7 +42,8 @@ public AddRequestCommandHandler(AllReadyContext context, IGeocoder geocoder)
 
             var address = _geocoder.Geocode(message.RequestViewModel.Address, message.RequestViewModel.City, message.RequestViewModel.State, message.RequestViewModel.Zip, string.Empty).FirstOrDefault();
             request.Latitude = message.RequestViewModel.Latitude == 0 ? address?.Coordinates.Latitude ?? 0 : message.RequestViewModel.Latitude;
-            request.Longitude = message.RequestViewModel.Latitude == 0 ? address?.Coordinates.Longitude ?? 0 : message.RequestViewModel.Latitude;
```
mgmccarthy: @stevejgordon thanks for catching this!

tonysurma:
@stevejgordon and @mgmccarthy are we good with this PR?

mgmccarthy:
@tonysurma, actually, this line still has a problem:
`request.Longitude = message.RequestViewModel.Latitude == 0 ? address?.Coordinates.Longitude ?? 0 : message.RequestViewModel.Latitude`

it needs to be:
` request.Longitude = message.RequestViewModel.Longitude == 0 ? address?.Coordinates.Longitude ?? 0 : message.RequestViewModel.Longitude`

AllReadyApp/Web-App/AllReady/Features/Requests/AddRequestCommandHandler.cs
stevejgordon: @mgmccarthy It's picked it up as a change but I don't think I actually did change it. I can apply the fix in this PR if you want?

mgmccarthy: @stevejgordon, yes please, the whole second line should be dealing with Longitude, not Latitude. Thanks

stevejgordon:
@tonysurma This should be good to go if others are happy. I've fixed the Longitude issue that Mike spotted above.

PR #1400, opened 2016-10-26
First PR to start work in Issue #1084

mgmccarthy:
First, badly named branch name, sorry about that, I started by pulling the code from PR #1111 down to my machine and working from there. that is why you'll see the first couple commits from Sean Epping in there. You can ignore those commits b/c I've pretty much changed everything about his initial design.

Addresses these line items from Issue #1084:
- Upon adding a request to an itinerary send a "scheduled for [date] notification with "Y" response for setting request to confirmed, "N" for setting request to unassigned
- Daily Web Job for 7 days before scheduled date "scheduled for next week on [date]"
- Daily Web Job for day before scheduled date "scheduled for tomorrow on [date]"
- Daily Web Job for day of for requests not confirmed message "sorry you couldn't make it, we will reschedule" and mark unassigned

Items not addressed from Issue #1084: (and reasons why)
- Requestor can specify communication preferences (text, voice) for communications: @MisterJames I did not pick this up b/c I'm not too sure what we would expect from the Requestor in terms of how they would tell us. Knowing that we're building this for the Red Cross/getasmokealarm, is there an example of a request that they would send to this endpoint?
- Promote text and fall back to voice. All messages are going out as sms right now. I have not added email or void
- If request is confirmed add "'N' if you can no longer make it" and if they respond N mark unassigned: Until I figure out how to put the RequestId on the outgoing message (in this case, the sms) that the requestor receives with "Y"/"N" response that then is routed towards the newly creasted SmsRequetsController, I won't be able to correlate back to the request that the requestor is responsding to. We can't use PhoneNumber, b/c there is nothing that would stop a Requestor from having two requests that were added to separate intineraries. In this case, if we used PhoneNumber to look up the requets, then we'll have two reqeuests returned
- If request is not confirmed add "'Y' if you can make it, 'N' if you can't" and set status as above: same answer as above
- On Installer marking previous request in itinerary delivered day of send "See you in a few" message: I put in a new with @tonysurma about clarifying when this message needs to go out.

So, next techical steps are:
1. figure out how the requestor will provide us with communication preferences (if they can do this)
2. Finish implementing SmsRequetsController. In order to do this, we need to find a way to allow the RequestId to ride the outgoing message to the Requestor, then ride the reply of "Y" or "N" back SmsRequsetsController. I've already written the Hangfire jobs that are to be invoked by the response type (Y/N)... just need someone to pick up the actual mechanics of the RequestId (I'm not too sure if it's possible to do what I'm saying above)
3. Promote text and fall back to voice: I'm not too sure what the conditions would be to "fall back to voice". If we make a decision on this, I'm sure some Twilio code/research would need to be involved to see how easy it would be to make the outgoing message become a phone call that would read the message to the requestor over the phone.


tonysurma:
Thanks @mgmccarthy I will merge this in and point configuration to the right db for firebase later today!

darrylsk:
I notice that the AllReadyApp.sln solution file is out of date (its file listing contains several old .xproj entries), and requires an upgrade for VS 2017.  Are you planning to keep this solution file?  If so, is it ok to go ahead and change and PR it or would you rather handle this one from your end?

PR #1401, opened 2016-10-26
Remove unused method on DateTimeOffsetConverter and consolidated code

mgmccarthy:
Simple cleanup for DateTimeOffsetConverter


dpaquette:
This looks good @mgmccarthy  

Thanks for cleaning this up further. I also home to renaming some of this and making it happen more automatically as part of a refactoring but I won't have time to tackle that until Sunday.

PR #1402, opened 2016-10-26
Making the test asserts locale independent

nemesv:
When the tests are run with the current culture set to hu-HU then the following two tests fail:
- VolunteerFulfilmentPercentage_ReturnsZero_WhenVoluneersRequiredIsZero
- VolunteerFulfilmentPercentage_ReturnsCorrectPercentage

They are failing because of the asserts are hard coded the decimal point.

Fixing the asserts to use the same pattern as the other tests in the  `EventDetailViewModelTests`.


tonysurma:
Thanks @nemesv for your contribution!  @mgmccarthy and @MisterJames can you take a review peek?

mgmccarthy:
@tonysurma, taking a look now

mgmccarthy:
@nemesv, thanks for the contribution!

@tonysurma :shipit:

PR #1408, opened 2016-10-27
Check for Hangfire database on startup and create it if it does not exist

### LINKED ISSUE #1407: Check for local Hangfire db and if doesn't exist, create it.
This is based on a comment from Issue #1399. Currently, if you do not have a Hangfire db created Server=(localdb)\\MSSQLLocalDB, the application throws an exception when you start it.

We want a developer-friendly F5 experience, so I added code in Startup.cs that checks for the existence of a Hangfire db, and if it does not exist, to create one.

I had to use ADO's SqlConnection and SqlCommand, which felt really dirty and brought back bad memories of being a .NET developer 10 years ago, but since Entity Framework does not know about the Hangfire db (and it shouldn't), I needed some way to check and create if missing

mgmccarthy:
PR #1409 now fixes 1407


tonysurma:
@MisterJames @dpaquette @stevejgordon or anyone else can you verify this works for you before we merge?

mgmccarthy:
@tonysurma, the result would still be an exception in case the connection cannot be made or something else happens on the dev's local machine... so, we get an exception now, and if this fails... we get an exception.

Either way, the db will still need to be created locally.

If anyone has a better idea that does not involve SqlConnection and SqlCommand, then I'm definitely open to that discussion.

AllReadyApp/Web-App/AllReady/Startup.cs
```diff
@@ -371,6 +370,23 @@ private IContainer CreateIoCContainer(IServiceCollection services)
                 await sampleData.CreateAdminUser();
             }
         }
+
+        private static void CheckForHangfireDatabaseAndCreateIfItDoesNotExist()
+        {
+            using (var sqlConnection = new SqlConnection("Server=(localdb)\\MSSQLLocalDB;Integrated Security=true;MultipleActiveResultsets=true;"))
```
dpaquette: This should be using the connection string from config. Not every developer will be using the same connection string

mgmccarthy: @dpaquette I wanted to use that, but the current db connections strings in config specific the hangfire db in the connection string, I don't want that, I need the connection string to not direct me to hangfire db b/c it does not exists yet. the query work is against master

dpaquette: In that case, I don't think this approach will work. We can't assume that the connection string specified here will work for all developers. For example, Linux of Mac devs will be connecting to a remote database,

Can you remind me why we can't create the Hangfire database tables in the regular AllReady database?

mgmccarthy: B/c hangfire looks by default for a database named "Hangfire" and builds the schema on that db. 

I'll look to see if that is something configurable where we could park the hangfire tables under the AllReady db. 

It's easy to forget that .NET now means more than just windows devs.

mgmccarthy: @dpaquette, for users connecting to a remote db, is that b/c they're running allReady in their Azure accounts?

dpaquette: @mgmccarthy they are running allReady locally but SQL Server doesn't exist on their local machine so they have the database hosted remotely. My understanding is that this would typically be an Azure database

dpaquette: @mgmccarthy it looks like you can give Hangfire a connection string with a database name: http://docs.hangfire.io/en/latest/configuration/using-sql-server.html 

If I'm reading this right, it also creates the Hangfire tables in a schema named Hangfire within the specified database. If that's the case I think we should put the hangfire stuff in the same database as allReady.

dpaquette: Let's jump on Skype to review this.

mgmccarthy:
@tonysurma cc @dpaquette so after a short convo with Dave, he got the Hangfire tables successfully cretaed under the AllReady db (has "Hangfire." schema for each hangfire table)...

so, we don't have to worry about the separate db any longer, and Startup will park those hangfire tables under the AllReady db. The change also gets rid of the Hangfire connection string, so @tonysurma, sorry about all the work you did with the new db and the connection string for hangfire. all of that can go away now.

If there are no objections, I'll be updating this PR to move to the code that Dave just hashed out with me.

tonysurma:
no worries @mgmccarthy I like that we have to configurable connection strings so you can do whatever you want.

So from this my suggestion would be to end at this:
a) have two connection strings in config one for allready app and one for hangfire
b) default config is the same allreadydb so in dev locally the F5 experience works that EF creates db if not there and then hangfire uses it too
c) then either by individual choice or config in azure we can set them to the same, different, remote, local whatever - this allows both Mac/Non-Windows Dev as well as me to split them in production - which I don't mind at all.

Thanks!

tonysurma:
thanks @mgmccarthy for all the work on the hangfire stuff and charging through new territory.

dpaquette:
@tonysurma I like that approach as well

mgmccarthy:
@tonysurma, you can delete/don't merge this PR. I did the hangfire db work in a new branch, and just submitted a PR for that, PR #1409.

Sorry about that, I should have just done the work in this branch, but I started a new one.

tonysurma:
nothing to be sorry about @mgmccarthy

PR #1420, opened 2016-11-01
Ensured that all columns in the form properly wrap

### LINKED ISSUE #1353: Create / Save button should be at the bottom of the Create / Edit Event page
https://goo.gl/mi5zFr

As I was scrolling down the page to create an event, I got to the bottom of the page and panicked a bit that there was no 'Save' button on the page. After a few moments of panic, I scrolled up to the top of the page and found the save button squished on the top edge of the page. 

I would like to have these buttons at the bottom of the page so I can save when I get to the end of the page.

YannickMeeus:
I have noticed that the responsiveness of this particular page leaves something to be desired. But the fix I've applied will at least ensure that the buttons stay below the form elements, even when resizing.

Fixes #1353


tonysurma:
@YannickMeeus is this a fix for #1353 ?  Thanks!

YannickMeeus:
@tonysurma It is indeed :)

PR #1421, opened 2016-11-02
Add filtering to organization list retreival during Campaign creation

### LINKED ISSUE #1347: Org Admins should not see other organizations when creating Campaigns
When logged in as an organization admin
While creating a new campaign
I have the ability to select any organization to assign the campaign to

I would expect that as a limited user to my own organization, that I would only have visibility into my organization. 

It seems that I am unable to actually create the campaign in an organization that I am not the admin of. However, instead of a useful error, I am redirected to the login page, even though I am already logged in.

I would suggest removing the Organization field completely from the screen, and detecting the organization from my user so the user does not have to think about this at all.

tonylambert:
Add filtering to the list of organizations that is returned to list box during the "Create Campaign" workflow thereby limiting which organizations are displayed.

Caveats:

- I cannot run the UI tests on my machine. They get four tests in and then the chromedriver.exe crashes on me. Therefore, I did not create a UI test to verify that this works. I did, however, test this manually.
- I am not very excited about this implementation, but I wanted to start somewhere and get some feedback from the team.
- It was not clear to me whether an OrgAdmin was only every an OrgAdmin for a *single* organization or potentially many. Based on the sample OrgAdmin's **Organization** claim, it appeared as if it is only one Organization, but I didn't want to make that assumption without asking.
- I can clean up the commits, but wanted to get some feedback on the overall approach first.

Closes #1347

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Services/SelectListServiceShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Services/SelectListServiceShould.cs
@@ -4,50 +4,120 @@ using System.Linq;
 using AllReady.Models;
 using AllReady.Services;
 using Microsoft.EntityFrameworkCore;
-using Microsoft.EntityFrameworkCore.Infrastructure;
 using Moq;
 using Xunit;
+using System.Security.Claims;
 
 namespace AllReady.UnitTest.Services
 {
     public class SelectListServiceShould
     {
+        private static int _organizationId1 = 1;
+        private static string _organizationName1 = "Organization Name 1";
+
         [Fact]
-        public void GetOrganizationsTest()
+        public void GetOrganizationsForAdminUserReturnsAllOrganizationsTest()
         {
-            const int organizationId1 = 1;
-            const string OrganizationName1 = "Organization Name 1";
+            var mockSet = CreateOrganizationsInMockDataStore();
 
-            var data = new List<Organization>
-            {
-                new Organization
-                {
-                    Id = organizationId1,
-                    Name = OrganizationName1,
-                },
-                 new Organization
-                {
-                    Id = 1,
-                    Name = "Organization Name 2",
-                }
-            }.AsQueryable();
+            var mockContext = new Mock<AllReadyContext>();
+            mockContext.Setup(c => c.Organizations).Returns(mockSet.Object);
 
+            var principal = new ClaimsPrincipal(new ClaimsIdentity(new[] { new Claim(AllReady.Security.ClaimTypes.UserType, "SiteAdmin") }));
 
-            var mockSet = new Mock<DbSet<Organization>>();
-            mockSet.As<IQueryable<Organization>>().Setup(m => m.Provider).Returns(data.Provider);
-            mockSet.As<IQueryable<Organization>>().Setup(m => m.Expression).Returns(data.Expression);
-            mockSet.As<IQueryable<Organization>>().Setup(m => m.ElementType).Returns(data.ElementType);
-            mockSet.As<IQueryable<Organization>>().Setup(m => m.GetEnumerator()).Returns(data.GetEnumerator());
+            var service = new SelectListService(mockContext.Object);
+            var organizations = service.GetOrganizations( principal ).ToList();
+
+            Assert.Equal(2, organizations.Count);
+            Assert.Equal(_organizationId1.ToString(), organizations.First().Value);
+            Assert.Equal(_organizationName1, organizations.First().Text);
+        }
+
+        [Fact]
+        public void GetOrganizationsForOrgAdminUserReturnsOnlyAuthorizedOrganizationTest()
+        {
+            var mockSet = CreateOrganizationsInMockDataStore();
+
+            var mockContext = new Mock<AllReadyContext>();
+            mockContext.Setup(c => c.Organizations).Returns(mockSet.Object);
+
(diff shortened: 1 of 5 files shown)


mgmccarthy:
@tonylambert, I'm currently taking a look at the code and will get back to you with a review.

@tonysurma, can you answer the third bullet point question in the PR comment above?

mgmccarthy:
@tonylambert, first and foremost, thanks so much for the PR. We count on volunteers to keep this project running, and you time spent on the PR is much appreciated...

After reviewing the PR, I have a couple ideas/proposed changes for you.

To start with, I'd like to get rid of the current string comparisons for user type via what we find in the user's claims. So something like this:
```
public interface ISelectListService
{
    IEnumerable<SelectListItem> GetOrganizations(ClaimsPrincipal user);
    ...
}
```

which would be implemented like this:
```
public IEnumerable<SelectListItem> GetOrganizations(ClaimsPrincipal user)
{
    var listOfOrganizations = new List<SelectListItem>();

    if (user.IsUserType(UserType.SiteAdmin))
    {
        listOfOrganizations = GetOrganizationsForSiteAdmin();
    }
    else if (user.IsUserType(UserType.OrgAdmin))
    {
        listOfOrganizations = GetOrganizationForOrgAdmin(user.Claims);
    }

    return listOfOrganizations;
}
```

There are a lot of existing extension method around ApplicationUser, ClaimsPrincipal, etc... the code that is `user.IsUserType` is an extension method.

Next, stylisitically, it seems that you've put a space between every `(` and `)` character. To align with how the rest of the codebase looks, can you please get rid of those spaces?

I pinged @tonysurma about your third bullet point question above"
> "It was not clear to me whether an OrgAdmin was only every an OrgAdmin for a single organization..."

so hopefully he'll be able to provide an answer.

Let me know if you have any question.

Thanks!

tonylambert:
Thank you so much @mgmccarthy! So much happier about getting rid of the magic strings. I'll make the suggested changes straight away.

mgmccarthy:
@tonylambert, I haven't had a chance to look at all the changes yet, but one thing I did leave out of the original PR feedback was:
- rename `SelectListServiceShould` to `SelectListServiceTests` (make sure the name of the .cs file also changes along with the class name)
- any test methods should have the word `Test` remove from the end of them

I'll probably be able to give this a once over tomorrow morning. Thanks!

tonylambert:
@mgmccarthy - Should be okay. I condensed the commits down into one commit because ultimately, it was a pretty small change once all the sausage making was out of the way.

If we do not need to mess with the multiple organizations per OrgAdmin, then I believe this is in a good place unless you have any other concerns I need to address.

tonylambert:
Regarding renaming the file from *Should.cs to *Test.cs -> Is that a convention that the solution is moving to? There are many test files that are named *Should and many that are named *Tests.

Regardless, fixed in c3c76ce

tonylambert:
I should add that I verified that you cannot create a Campaign for an organization that you are not the OrgAdmin of. What I mean by that is the "POST" that occurs when you attempt to create the Campaign guards against it so we only needed to filter the listbox and not add additional checks elsewhere in the code for this.

The pertinent code in CampaignController:

```cs
        // POST: Campaign/Edit/5
        [HttpPost]
        [ValidateAntiForgeryToken]
        public async Task<IActionResult> Edit(CampaignSummaryViewModel campaign, IFormFile fileUpload)
        {
            if (campaign == null)
            {
                return BadRequest();
            }

            if (!User.IsOrganizationAdmin(campaign.OrganizationId))
            {
                return Unauthorized();
            }
            ...
```

tonysurma:
@tonylambert as @mgmccarthy said thank you so much for your contribution!  In regards to your question, yes an org admin is an org admin for only one org.  That is the requirement.  Thanks again!

mgmccarthy:
@tonylambert, we do have guidance for how tests should be named on our AllReady docs page:
http://htbox.github.io/

BUT, the guidance there probably should be updated. Right now it says that any Controller unit test class should be named ending in "Test" and the rest should be named "Should"...

what I should probably update it to is any test classes that test a class with more than one method should end in "Tests" and any class that only need to test one method should end in "Should"...

so for example, Handler unit tests end in Should b/c they only have one testable method per class, "Handle".

mgmccarthy:
@tonylambert, please make these changes:

- delete the SanitizeUser method in SelectListService and do not check for null on the ClaimsPrincipal passed into GetOrganizations. We don't need this check.

For your unit tests, we have a base class called `InMemoryContextTest` which you can inherit from, which gives you access to an  EF Context. It looks like you used Moq to mock out all calls to _context in SelectListService, and you don't need to do any of that work.

To see how to work with the Context exposed by InMemoryContextTest, you can look at any of the test classes that inherit from it and use that as a guide for how to implement your unit tests.

If you have any questions on InMemoryContextTest, please reach out to me.

Thanks!

tonylambert:
Thank you for the link to the docs. I hadn't run across them yet.

Regarding using InMemoryContext, I'll switch over to it in the next couple days.

tonylambert:
@mgmccarthy - New commits added. I also changed the other test in the file that didn't have anything to do with this PR because it was using Moq over InMemoryContextTest.

Let me know if anything else is needed.

mgmccarthy:
@tonylambert, thanks for all the work on this. I'll try to review it tonight or first thing tomorrow.

mgmccarthy:
@tonylambert, this looks good. Thanks for working with us to make the fixes. @tonysurma :shipit:

PR #1416, opened 2016-10-30
Add TimeZoneID to Event

dpaquette:
Adding TimeZoneID to Event class and cleaning up `DateTimeOffset` adjustment code.
- [x] Add TimeZoneID to Event class
- [x] Default Event's TimeZoneID to the TimeZoneID from the Campaign
- [x] Update Event and Task Edit actions to use the event's TimeZoneID instead of Campaign's TimeZoneID
- [x] Automatically adjust DateTimeOffset values based on TimeZoneID
- [ ] Unit tests
- [ ] Documentation on usage

PR #1442, opened 2016-11-04
WIP: Use specific input types on form elements

tonylambert:
If we are looking for something more significant with polyfills, masking keys, etc then this might not be the best one for me to tackle. This is a minefield in terms of browser support. It may be safest and most future proof to just use the simple tweak of `<input type="tel|email|etc">` to start. My recommendation is to KISS for now and do something more _intense_ if we get additional feedback.

More detail can be found here: https://www.smashingmagazine.com/2015/05/form-inputs-browser-support-issue/

It seemed safest to not try to add specific patterns to the input, but if you guys have strong feelings (or have suggestions like leveraging polyfills like http://chemerisuk.github.io/better-dateinput-polyfill/ then let me know).

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Edit.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Edit.cshtml
@@ -99,7 +99,7 @@
                 <p><cite>Optional link to an outside page</cite></p>
             </div>
             <div class="col-md-10">
-                <input asp-for="ExternalUrl" class="form-control" />
+                <input type="url" asp-for="ExternalUrl" class="form-control" />
                 <span asp-validation-for="ExternalUrl" class="text-danger"></span>
             </div>
         </div>
@@ -120,14 +120,14 @@
         <div class="form-group">
             <label asp-for="StartDate" class="col-md-2 control-label"></label>
             <div class="col-md-10">
-                <input asp-for="StartDate" class="form-control datepicker" />
+                <input type="date" asp-for="StartDate" class="form-control datepicker" />
                 <span asp-validation-for="StartDate" class="text-danger"></span>
             </div>
         </div>
         <div class="form-group">
             <label asp-for="EndDate" class="col-md-2 control-label"></label>
             <div class="col-md-10">
-                <input asp-for="EndDate" class="form-control datepicker" />
+                <input type="date" asp-for="EndDate" class="form-control datepicker" />
                 <span asp-validation-for="EndDate" class="text-danger"></span>
             </div>
         </div>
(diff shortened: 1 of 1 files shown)


tonysurma:
Big fans of KISS on a project like ours.  That is the right approach.  Will merge after a quick review.  Thanks

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Edit.cshtml
```diff
@@ -120,14 +120,14 @@
         <div class="form-group">
             <label asp-for="StartDate" class="col-md-2 control-label"></label>
             <div class="col-md-10">
-                <input asp-for="StartDate" class="form-control datepicker" />
+                <input type="date" asp-for="StartDate" class="form-control datepicker" />
```
dpaquette: Using type="date" conflicts with the bootstrap date picker that we use in AllReady. Please remove these as it causes strange behavior with overlapping dropdowns.

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Edit.cshtml
```diff
@@ -99,7 +99,7 @@
                 <p><cite>Optional link to an outside page</cite></p>
             </div>
             <div class="col-md-10">
-                <input asp-for="ExternalUrl" class="form-control" />
+                <input type="url" asp-for="ExternalUrl" class="form-control" />
```
dpaquette: I would prefer specifying the type using a data annotation on the ExternalUrl property of the view model. That way the input tag helper takes care of setting the type appropriately. 

In the ViewModel:

```
[Url]
property string ExternalUrl { get; set; }
```

More on the input tag helper here: http://www.davepaquette.com/archive/2015/05/13/mvc6-input-tag-helper-deep-dive.aspx

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Edit.cshtml
```diff
@@ -187,14 +187,14 @@
...
-                <input asp-for="PrimaryContactPhoneNumber" class="form-control" />
+                <input type="tel" asp-for="PrimaryContactPhoneNumber" class="form-control" />
                 <span asp-validation-for="PrimaryContactPhoneNumber" class="text-danger"></span>
             </div>
         </div>
         <div class="form-group">
             <label asp-for="PrimaryContactEmail" class="col-md-2 control-label"></label>
             <div class="col-md-10">
-                <input asp-for="PrimaryContactEmail" class="form-control" />
+                <input type="email" asp-for="PrimaryContactEmail" class="form-control" />
```
dpaquette: PrimaryContactEmail is already tagged with the [EmailAddress] data annotation which should output `type="email"`. Specifying the type again here is redundant.

tonylambert:
Thanks @dpaquette! I'll switch them to the helper tags this evening.

dpaquette:
Thanks @tonylambert 

I should have noted above. One of the reasons to prefer using data annotations is that we also get sever side validation that way. 

Also, it would be better if you didn't make any changes to the Date / DateTime properties because I am making changes to those as part of PR #1416

tonysurma:
thanks again to @tonylambert for jumping in and @dpaquette for the review

tonylambert:
@tonysurma, @dpaquette - I apologize. I didn't do my due diligence when starting this PR. I didn't _view source_ on the resulting code of the page. All of the input _types_ were already specified because they are already attributed with the helper tags in the model.

Therefore, at best, the changes in this PR would have effectively been a NOP and at worst they would have caused conflicts down the road.

In short, I'll close this PR out and I think the issue should be close too. I tested the site on my mobile device proxied through Fiddler to my machine and it properly displayed the the keyboard  you would like to see: number keyboard for phone, @ sign visible for the email field, date picker for dates, etc.

tonysurma:
No worries of course @tonylambert 

given nature of the open work it is not impossible for issues to be out of date etc.  Thanks for verifying via fiddler/mobile device!

Issue #1446, opened 2016-11-05
mobile-app: config.xml hook script does not work


digitaldrummerj:
In the mobile app's config.xml file the hook script scripts/bowerCopy.js does not work and causing ionic builds to fail.  The goal of this script is to reduce the size of the bower components in the android and ios builds.  It was part of the project before it was converted over to a true ionic project.  Need to investigate if this script is still needed.

SteveKennedy:
Some investigation:   

@digitaldrummerj added a .bowerrc (as part of conversion to true Ionic project) which forced bower to download all bower packages, with some having entire source repositories, directly in the www/lib folder.  This is a safe way to ensure that all bower package contents get restored to the project during bower's default restore routine.  However, it also does not try to exclude any files or folders included in those packages that are not needed by the project. 

Before the addition of the .bowerrc file, bower packages would be added to the default bower_components folder, and then the bowerCopy.js would only copy certain files and folders into the project's www/lib folder, and not the entire package's contents. This lead, as @digitaldrummerj stated, to  a smaller build size for ios and android.

After evaluation, we could one of the following:

1. Remove the bowerCopy.js and it's reference in the config.xml.  This would be the easiest/quickest solution, and have no ill effects, except it would leave unused package files in the www/lib folder, causing a larger build/binary size.  

2. Remove the custom .bowerrc file, ensuring bower components get restored to the default bower_components directory. We could then un-comment the bowerCopy.js reference in the config.xml, and reap the benefits of the file/folder savings. If we chose this route, we'd save something around 1.2mb in the lib folder (made up of 350 files, and 17 sub-folders).  There also might be a small tweak we'd have to make the  NONSTANDARD_DIST_PATHS array in the bowerCopy.js for the Ionic package - there's no more release folder in that bower package.

3. Come up with a new route of eliminating unused files directly from the www/lib.  This might be something like making a list of files/folders to explicitly delete out of the www/lib folder.  I wouldn't do this because it seems odd to download the entire package contents, just to delete some of them.

4.  ???

PR #1451, opened 2016-11-05
commented out the hook script until it is working

digitaldrummerj:
related to https://github.com/HTBox/allReady/issues/1446

the only change was to comment out the hook script line.  the rest of the changes are from visual studio deciding to reformat the file with new whitespace :(


digitaldrummerj:
note that this does not fix the underlying issue of the hook script not working but it at least makes the build work again to deploy to android or ios.

PR #1469, opened 2016-11-10
Update reference to tinymce cdn

### LINKED ISSUE #1455: bug: tinymce errors on homepage in dev
On the home page at [http://allready-d.azurewebsites.net/](http://allready-d.azurewebsites.net/) I saw the following errors:

image

tonylambert:
After exploring tinymce's site and doing some local testing, it appears that the cdn reference to tinymce has changed.

Closes #1455


tonylambert:
I should add that there isn't a way to explicitly ask for 4.2.6 from the new way to reference tinymce off of their CDN. It looks like you can ask for minor versions 4.x, but not 4.x.y.

This was the only thing I was concerned by

PR #1475, opened 2016-11-10
Issue1456

### LINKED ISSUE #1456: bug: login page giving this.optional is not a function for custom email validation rule
The custom email validation rule is throwing errors for me in the dev tools console when the page loads.  I do have the username and password saved in the browser and auto filled.  

image

pranap:
Removed custom validation for email because the spaces seem to be
trimmed before validation on the server side. If you still need to trim
on client side, change originalRule(value, element) to
originalRule.call(this, value, element); in _ValidationScriptsPartial.cshtml

fixes #1456


tonysurma:
@dpaquette @stevejgordon if you get a chance can you pass a second set of eyes on this?  thanks again @pranap for the contribution and for helping out at events for HTBox!

pranap:
I'll get a chance to work on it later this afternoon.

dpaquette:
I'm not sure why this trim logic was added in the first place. Someone explicitly put it there but I have no idea why. I think this should be ok to merge

PR #1488, opened 2016-11-13
Moved text on 404 screen above image.

### LINKED ISSUE #1484: Move/Fix overlapping text on image on 404 page
Per issue #1360 and PR #1483 there was a partial fix and now "there is text that overlays the image. This issue may have already existed. The black text on the black & white image creates a readability issue. My suggestion would be to move the text above the background image."

dchristensen:
Moved the text on the 404 screen above the image.
To address display issues caused by the text interacting with the existing CSS I placed the image in its own img tag which is sized using the bootstrap responsive image and cleaned up the unused CSS.

Closes #1484

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Views/Error/404.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Views/Error/404.cshtml
@@ -1,45 +1,15 @@
-﻿@section styles 
-{    <style>
-        .errbackimage         
-        { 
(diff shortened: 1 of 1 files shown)


dchristensen:
Not sure why CI failed.  Tests are all green locally and I certainly didn't change anything in the Validators.

tonysurma:
I will try to rerun and see if it was just a blip

tonysurma:
And it's good

Issue #1411, opened 2016-10-28
Validation is not working in the "Message All Volunteers" modal


nemesv:
When you try to create an invalid message, without a subject and message, "nothing happens": 
- no error message
- the modal stays open
- the "Send Message" gets disabled

The following JS error is displayed in the console: 

```
VM209:1 Uncaught SyntaxError: Unexpected token B in JSON at position 0(…)
error @ taskDetails.js:40
```

![image](https://cloud.githubusercontent.com/assets/251330/19819945/40d214ce-9d57-11e6-90da-b7da1571fbf9.png)

Repro Steps.
1. Log into Org admin That has an Event created.
2. Add volunteers to event task.
3. Navigate to Task Details /admin/task/details.
4. Click Message All Button.
5. Leave all inputs empty
6. Click Send Message Button

adam3039:
Doing this!

dcbrewster:
adam3039. Are you still working on this?

tonysurma:
@dcbrewster it's been long enough that I would consider this not work in progress by @adam3039 and feel free to start working on it.  Thanks.

stevejgordon:
@dcbrewster - Did you start working on this in the end?

dcbrewster:
Yes. It is done. Just need the other pull request (#1676) to be implemented before I commit these changes\

Dave

stevejgordon:
@dcbrewster - Great! Thanks for the swift reply too.

naeemsarfraz:
I've just submitted PR #1723 which adds a required attribute to the input elements which should stop an empty form being posted.

Looking into providing feedback to the user about what's gone wrong.

BillWagner:
Fixed in PR #1723

PR #1489, opened 2016-11-13
Validation for Task Message Volunteers (#1411)

adam3039:
I noticed some very weird behavior when returning the original BadRequest(ModelState) to an AJAX for submission. Sometimes it would return 1 of the validation errors, but most times it only returns "Bad Request" as the responseTest. The Modelstate errors do not appear to be in any part of the response object the AJAX error callback receives.  This was causing JSON parsing issues. I also noticed that this problem exists anywhere we have
`
return BadRequest(ModelState);`

which occurs in the EventAdminController when sending a message to all the volunteers within an event, and in the IntineraryController, when creating a new Itinerary. All three areas have this validation bug, but I have only fixed this one because I am unsure if we want to proceed with this solution.

I resolved this by returning a JsonResult with a success flag and the model state errors. In the AJAX success method, I am either returning a success message if the flag is true, or iterating through the errors and displaying them within the validation message box if false. If the error event is hit within the AJAX, it returns a generic error message, as this means it was something server related and not a validation error. With this solution there is no JSON parsing to worry about.


adam3039:
This fix obviously breaks the unit test which confirms a badrequest is returned if a modelstate error exists.

tonysurma:
So should we remove the unit test then? Cc @mgmccarthy

adam3039:
@tonysurma @mgmccarthy  I could also modify the unit test to check the JSON response to make sure the success flag is set to false if an error exists in the modelstate

mgmccarthy:
@tonysurma cc @adam3039, anything "javascript" is a bit outside my wheelhouse. I can take a peek at the code, but I'm not too sure if I should be the one making decisions on it or reviewing it.

tonysurma:
@dpaquette or @canadianjames, thoughts?

tonylambert:
@adam3039, @tonysurma - I realize you didn't ask me to weigh in, but I figured I'd share my thoughts anyway if that's okay.

A couple opinions:

- I am not a huge fan of returning automated errors from code behind. The errors that a user sees should be a view specific concern for usability, ease of localization in the future (may not be an issue for this project), etc. Therefore, I do not like pushing the model state's validation errors directly out to the user...they are a bit too "programmer-y" in my opinion.
- It is a bit alarming that the modelstate data isn't being populated when it shows up in the browser. I wonder if this is a bug with ASP.Net 5, if Microsoft decided to break with how the previous versions of ASP.Net Web API worked (intended and different than the old behavior), or something else is going odd in the allReady solution to override/break the default behavior.

As far as an overall solution, I think your solution style is okay if we feel that these are "application level" errors and not "HTTP level" errors. Personally, I like trying to leverage HTTP as much as possible and BadRequest feels okay because the request was bad...it didn't contain the appropriate inputs and therefore the request is a bad one.

This makes me think that we should actually do client side validation before the ajax call. We would still need to update the code to handle BadRequest and Unauthorized responses. However, this isn't likely going to be sufficient in all cases because we would likely duplicate a lot of logic on the client side that the server is trying to do automatically through attributes on the models.

However we decide to solve this particular issue (and the other places you found in the code), we should definitely update the unit tests to behave however we want these to.

adam3039:
Thanks for weighing in @tonylambert . I entirely agree that I think it may be a priority to investigate BadRequest(ModelState) further, as I do think it is the cleanest solution (at least for server side validation). I do think something is happening within the solution that is overriding/breaking the default behavior, as you said. 

I provided this as a solution because I've seen model validation handled within the success callback of an AJAX response before, while the error callback is reserved for more server-centric errors.

I agree that the cleanest solution is likely investigating what is causing the BadRequest(ModelState) to just return a "Bad Request" string, as well as providing client side validation for the required fields. Since the ViewModel already has the necessary validation attributes, I believe it would just be a matter of adding unobstrusive validation to the project, but someone could certainly correct me on this.

tonylambert:
@adam3039 - I haven't tried this, but maybe creating a brand new toy ASP.Net API project and see if you can get 

`return BadRequest(ModelState)`

to return the way we expect. It at least tells us if the framework has changed (or has a bug which is probably unlikely) or if we have broken something.

dpaquette:
I think that  `return BadRequest(ModelState)` would be the preferred approach. It looks like others have had problems with this and I'm not sure what the resolution was. I did not have time to read through the entire issue.
https://github.com/aspnet/Mvc/issues/4607

nemesv:
@tonylambert  The ```return Badrequest(ModelState)```;  is working fine, this is not a bug in ASP.NET. 

The problem is in the web.config because the web server is cutting out the response body and replacing it with the text "Bad Response", because of the ```existingResponse="Replace"``` setting:

https://github.com/HTBox/allReady/blob/master/AllReadyApp/Web-App/AllReady/web.config#L7-L10:

```
    <httpErrors errorMode="Custom" existingResponse="Replace">
      <remove statusCode="404" />
      <error statusCode="404" prefixLanguageFilePath="" path="/Error/Error404" responseMode="Redirect" />
    </httpErrors>
```

So I think the solution should be that we change this setting to ```existingresponse=PassThrough``` this would preserve the properly serialized ModelState in the response.

tonylambert:
@adam3039, @nemesv - That sounds like it would solve the problem, but is that opening up any information disclosure issues that we would want to avoid? I honestly don't know enough about other scenarios that could provide other details. If there are unhandled exceptions, then those would come back as 500 errors. It would seem that this wouldn't be a problem, but I haven't put a lot of thought into it.

tonysurma:
@adam3039 based on the comments here from @dpaquette and others is there anything more you need?  Thanks again

PR #1492, opened 2016-11-16
Request creation/edit looks up lat/long for address

### LINKED ISSUE #1398: Manual Request Creation - Latitude/Longitude Lookup
When creating a request via the admin UI and saving it via EditRequestCommandHandler we should call the Bing API to determine the latitude and longitude for storage in the DB.

We need these details when representing the requests on the map.

tonylambert:
Work in this PR

- [x] Leverage the IGeocoder object to fetch the lat/long for the request when it is saved.
- [x] Get Google GeoCoder keys in order to actually test this locally as by default its using the NullObject geocoder
- [x] Create unit tests around this handler
- [x] Explore where a "shared" location for the code that is duplicated between ImportRequestCommandHandler and EditRequestCommandHandler could live/belongs since I don't like the current duplication and creating a base class *GeoCoderRequestCommandHandler doesn't feel right to me (such that the base class could house the shared method. **Decision after exploration - Not worth the effort.**
- [x] Found a bug. If you edit a **Request** after creation, then command handler will explode and cause an HTTP 500 status code. Will fix this bug along with this PR. _I didn't log an issue for this, but can if there is a need to track it (the bug was present before this PR)._
- [x] Only update the lat/long of the **Request** when we detect an address changed when editing a **Request**.

Closes #1398

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
@@ -1,18 +1,36 @@
 ﻿using AllReady.Areas.Admin.Features.Requests;
 using AllReady.Areas.Admin.ViewModels.Request;
+using AllReady.Models;
 using AllReady.Providers;
 using MediatR;
 using Moq;
 using System;
 using System.Threading.Tasks;
 using Xunit;
+using System.Linq;
 
 namespace AllReady.UnitTest.Areas.Admin.Features.Requests
 {
     public class EditRequestCommandHandlerShould : InMemoryContextTest
     {
+        private Request _existingRequest;
         protected override void LoadTestData()
         {
+            _existingRequest = new Request
+            {
+                Address = "1234 Nowhereville",
+                City = "",
+                Name = "",
+                DateAdded = DateTime.MinValue,
+                EventId = 1,
+                Phone = "555-555-5555",
+                Email = "something@example.com",
+                State = "WA",
+                Zip = "55555"
+            };
+
+            Context.Requests.Add(_existingRequest);
+            Context.SaveChanges();
         }
 
         [Fact]
@@ -24,6 +42,22 @@ namespace AllReady.UnitTest.Areas.Admin.Features.Requests
             var requestId = await handler.Handle(new EditRequestCommand { RequestModel = new EditRequestViewModel {  } });
 
             Assert.NotEqual(Guid.Empty, requestId);
+        }
+
+        [Fact]
+        public async Task UpdateRequestsThatAlreadyExisted()
+        {
+            var mockMediator = new Mock<IMediator>();
+            string expectedName = "replaced name";
+
+            var handler = new EditRequestCommandHandler(Context, new NullObjectGeocoder());
+            var requestId = await handler.Handle(new EditRequestCommand
+            {
+                RequestModel = new EditRequestViewModel { Id = _existingRequest.RequestId, Name = expectedName }
+            });
+
+            var request = Context.Requests.First(r => r.RequestId == _existingRequest.RequestId);
+            Assert.Equal(expectedName, request.Name );
         }
     }
 }
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
@@ -40,8 +40,6 @@ namespace AllReady.UnitTest.Areas.Admin.Features.Requests
         [Fact]
         public async Task ReturnNewRequestIdOnSuccessfulCreation()
         {
-            var mockMediator = new Mock<IMediator>();
-
             var handler = new EditRequestCommandHandler(Context, new NullObjectGeocoder());
             var requestId = await handler.Handle(new EditRequestCommand { RequestModel = new EditRequestViewModel {  } });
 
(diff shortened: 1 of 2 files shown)


mgmccarthy:
@tonylambert, what Issue is this PR for?

mgmccarthy:
@tonylambert, what Issue is this PR for?

mgmccarthy:
@tonylambert, could you be more specific about the duplicated code  you need think needs to be eliminated?

tonylambert:
@mgmccarthy - Sure thing. In commit f3e4d24 there are lines focused on calling out to the geocoder:

```
            //If lat/long not provided, use geocoding API to get them
            if (request.Latitude == 0 && request.Longitude == 0)
            {
                //Assume the first returned address is correct
                var address = _geocoder.Geocode(request.Address, request.City, request.State, request.Zip, string.Empty)
                    .FirstOrDefault();
                request.Latitude = address?.Coordinates.Latitude ?? 0;
                request.Longitude = address?.Coordinates.Longitude ?? 0;
            }
```

This code was directly copy and pasted from ImportRequestsCommandHandler - https://github.com/HTBox/allReady/blob/96eca0e048cd4d3c8d1631844b8547053b0054dd/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Requests/ImportRequestsCommandHandler.cs#L32

mgmccarthy:
@tonylambert, that's not code duplication, that's plain old implementation. 

`IGeoCoder` was created to re-use code, and to properly abstract geo-coding functionality out of classes that need them, so that's pretty much the code re-use we're going to get there.

I'd remove the:

> Explore where a "shared" location for the code that is duplicated between ImportRequestCommandHandler and EditRequestCommandHandler could live/belongs since I don't like the current duplication and creating a base class *GeoCoderRequestCommandHandler doesn't feel right to me (such that the base class could house the shared method.

line item from the PR. Thanks.

Also, I'll see if I can take a look at this over lunch.

tonylambert:
@mgmccarthy - You may be right regarding no need to tweak the implementation. Anytime I copy and paste something I want to make sure there isn't a better way. As of right now the code doesn't support country codes, for instance. If we start to then we would need to change at least these two places (I'm oversimplifying, of course), and that gives me pause.

It is possible that a "solution" to the duplication issue I brought up above could be less maintainable or worse than keeping (what I consider) duplication.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
```diff
@@ -25,5 +43,21 @@ protected override void LoadTestData()
 
             Assert.NotEqual(Guid.Empty, requestId);
         }
+
+        [Fact]
+        public async Task UpdateRequestsThatAlreadyExisted()
+        {
+            var mockMediator = new Mock<IMediator>();
```
mgmccarthy: delete this line as the mockMediator is not used anywhere

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
```diff
@@ -25,5 +43,21 @@ protected override void LoadTestData()
...
+            string expectedName = "replaced name";
+
+            var handler = new EditRequestCommandHandler(Context, new NullObjectGeocoder());
+            var requestId = await handler.Handle(new EditRequestCommand
```
mgmccarthy: delete `var requestId` as it's not being used by the test

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
```diff
@@ -63,7 +65,7 @@ protected override void LoadTestData()
         }
 
         [Fact]
-        public async Task AttemptToGeocodeAddressToGetLatitudeAndLongitude()
+        public async Task AlwaysGeocodeAddressWhenUpdatingExistingRequest()
         {
             var mockMediator = new Mock<IMediator>();
```
mgmccarthy: delete this line as it's not being used

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
```diff
@@ -74,7 +76,7 @@ protected override void LoadTestData()
             var handler = new EditRequestCommandHandler(Context, mockGeocoder.Object);
             var requestId = await handler.Handle(new EditRequestCommand
```
mgmccarthy: delete `var requestId` as it's not being used

mgmccarthy:
@tonylambert, ignore my previous comment on not knowing where we can update the Request. The UI does both add/update. My bad. It's getting too late for me to be doing code reveiews ;)

tonylambert:
@mgmccarthy - Removed the unused variables in 46c7095. I miss my resharper :).

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
tonylambert: Address in 46c7095

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
tonylambert: Addressed in 46c7095

mgmccarthy:
@tonylambert, oh no! You lost ReSharper? I don't know how I'd work without it.

tonylambert:
@mgmccarthy - I have it at work (not that I am coding much there these days), but not at home. VS community edition and no Resharper  at home. At least VsVim works in VS Community :).

tonylambert:
@mgmccarthy - Following up on the duplication item, I think there are a couple ways that this could go that would actually be an improvement. As an example, ImportRequestsCommandHandler could *use* EditRequestCommandHandler in a composition model and only have the details dealing with geocoding in the EditRequestCommandHandler class. This would necessitate some more tweaks with the model classes (which may be a good idea anyway, but that's a topic for another time) as well as a mechanism to tell the EditRequestCommandHandler to defer calling SaveChanges on the context.

**In short, I am in agreement that working that minor duplication isn't worth it at this time.**

tonylambert:
There may be some changes needed, of course, but I am moving this out of *WIP* status now. I updated the PR title/comment to reflect its state.

mgmccarthy:
@tonylambert, this looks good. @tonysurma :shipit:

Issue #1102, opened 2016-08-07
geocoded requests: order the list of selected items as optimized route (given start address)


tonysurma:
sub issue of #1080

stevejgordon:
@tonysurma I had a quick look at this the other day and can't see an obvious way to do this with the Bing API. We can add waypoints to create a route, but it does so in the order you specify.

I'd looked at Bing since the maps we're using are from Bing. However we do use GeoCoding from Google for the request imports.

The Google API seems to allow for optimized routes of waypoints (exactly what we need) so I'd propose we use that to generate the routes. However I wanted to check before we do that since we might need to purchase a key. There is a limit on the free one of up to 8 waypoints.

cc/ @MisterJames

tonysurma:
Thx, let's do the first 8 to see how it works without a key and we can look at cost later.  8 may actually be enough for the majority of use cases

stevejgordon:
@tonysurma This is what I'm planning to do. Any feedback, please let me know...

**Optimise Route Feature**

On then Itinerary details page add a hyperlink to "optimise route". Click this will show a popup with fields required to calculate the optimal route (if more than 8 requests this should be greyed out and unavailable).

Clicking the popup opens a UI containing:
- Start Address 
- End Address (tick box to use same as start). If ticked then we hide the end address textbox
- Option to include unconfirmed requests? - Or what should we do with them.

On submit we'll query the Google directions API for optimal route from Start to End via the waypoints.
We'll map the response to an object so we can then arrange the requests into the optimised order.

**Display Driving Route Feature**

If possible we'll just include the route on the current map. If not easy to integrate with the map that is there then perhaps this will also be a hyperlink which opens the bing maps page with the waypoints populated. Or it could open a page on allReady which customises the display of this info.

tonysurma:
@stevejgordon looks good, @OhMcGoo any thoughts?

As for unconfirmed, I would say that any requests added to the itinerary get included.  If they included unconfirmed use them, if not don't so to speak.  This way you don't filter/add more logic, just go with what is in itinerary and follow user's choices.  Does that work?

stevejgordon:
@tonysurma Yep, that's fine. We can always enhance later if we need to. I will try to get to this ASAP.

stevejgordon:
@tonysurma This is coming along and I hope nearly there. I will have a bit of refactoring and testing to include before I submit.

[...]

PR #1495, opened 2016-11-16
Adding ability to edit itinerary

### LINKED ISSUE #1331: Add page to edit an Itinerary
This page or popup should allow an admin to update the date and title of an itinerary.

stevejgordon:
Fixes #1331 

This is an initial quick pass to enable editing an itinerary. The main purpose is to allow a start/end address to be set for an itinerary to the be used in my work for issue #1102 

I'll continue with that issue once this is merged.

Note: I've not included unit tests for the edit itinerary handler yet. I'll open a follow up issue for that.
The validation also needs tweaking to validate the addresses as well. Again, I'll follow that up in a new issue.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/ItineraryAdminControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/ItineraryAdminControllerTests.cs
@@ -20,6 +20,7 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using Moq;
+using Shouldly;
 using Xunit;
 
 namespace AllReady.UnitTest.Areas.Admin.Controllers
@@ -94,7 +95,7 @@ namespace AllReady.UnitTest.Areas.Admin.Controllers
 
             var sut = new ItineraryController(mediator.Object, null);
             sut.MakeUserNotAnOrgAdmin();
-           
+
             Assert.IsType<UnauthorizedResult>(await sut.Details(It.IsAny<int>()));
         }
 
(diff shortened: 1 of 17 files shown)


mgmccarthy:
@stevejgordon, are we allowing the editing of an Itinerary's date?

stevejgordon:
@mgmccarthy Currently, yes. This was mostly put in to manage the addresses though.

The editing of the date may have some impact on the notifications I guess. I can remove it from this PR. It does feel reasonable that someone may need to adjust an itinerary date but we'd need to decide on handling the change.

Also, there's not specific validation on this yet. It's a bit quick and dirty to enable the work on the route optimisation.

mgmccarthy:
@stevegordon. Don't remove it until I have a chance to look at the scheduling/hangfire stuff. There might be a way to handle it in the job code.

mgmccarthy:
@stevejgordon, to me, should we have the ability to edit an Itinerary's date is more of a business question than a technical concern, but by allowing the editing of an Itinerary's date at any point in the lifetime of the Itinerary will force us to write more code which will create complexity.

From a technical point of view, I have two concerns that add complication to the system:

1. Hangfire jobs "in-flight": since we schedule Hangfire jobs based on the Itinerary's date, changing the date could potentially lead to sms messages that are either not sent, or sent on the wrong date. 
2. Itineraries that have had requestor's confirm their request means they confirmed their request based on the date of the Itinerary that the Request was added to. If we were to change the Itinerary date after a requestor(s) confirmed their request with us, now we need to go through a look-up process, and re-send more confirmation messages to requestors notifying them that the date of the Itinerary has changed and they need to confirm the new date with us.

Because we'd have to add code to address these two issues above, which can create more complexity, here is a list of what we could do in order of easiest to hardest.

1. once an Itinerary has been changed, do not allow it's date to be edited
2. once an Itinerary has one Request assigned to it (regardless of the source of the Request), do not allow the date to be edited
3. allow the date to be edited at any point once an Itinerary has been created (I don't think we should choose this option for the reasons I outlined above).

@tonysurma @MisterJames @dpaquette what are your thoughts?

tonysurma:
Let's remove editing date to make simplicity a winner.  We can open an issue on feature backlog for date editing and see if it ever gets prioritized and then we're can deal with the complexity noted above.

When you remove that @stevejgordon we can merge this

mgmccarthy:
@tonysurma, had a quick chat with @stevejgordon this morning about this and it sounds like we're all on the same page. Thanks.

mgmccarthy:
@stevejgordon, let me know if I can help in any way

stevejgordon:
@tonysurma All done and ready to go

PR #1524, opened 2016-11-18
Target .NET 4.6 to fix thread context switch losing CurrentCulture an…

### LINKED ISSUE #1223: Edit Event Start and End times are incorrect
When clicking to edit an event the edit form date time is displaying the incorrect time. It appears the time is being converted to UTC rather than retaining the original local time.

This might be easier to fix after merge of #1154

**Time in event display page (correct):**
image

**Time in event edit page (incorrect):**
image

### LINKED ISSUE #1224: Edit Task Start and End times are incorrect
As per #1223 but for tasks

dpaquette:
…d CurrentUICulture

As per https://github.com/aspnet/Mvc/issues/4692

Resolves issues #1224 and #1223

Fixes #1224 
Fixes #1223

### DIFF
--- a/AllReadyApp/AllReady.NotificationsWebJob/project.json
+++ b/AllReadyApp/AllReady.NotificationsWebJob/project.json
@@ -15,7 +15,7 @@
   },
 
   "frameworks": {
-    "net451": {
+    "net46": {
       "dependencies": {}
     }
   }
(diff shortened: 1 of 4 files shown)


tonysurma:
@dpaquette does this remove the ability to run on mac non-windows then?  Not a show stopper just want to verify

dpaquette:
No, this just moves us to a newer version of full framework. We still plan to move to target .NET Core instead of full framework but we haven't removed the dependency on System.Drawing yet. 

The CurrentCulture problem was specific to 451. It won't be a problem once we move to target Core

tonysurma:
I accidently deleted my question so I am adding it back in so @dpaquette answer has context 😸 

"Does this remove the ability to work on non-windows platforms like Mac?"

Issue #1521, opened 2016-11-17
Change phone number to mobile phone number when gathered on user registration screens


schuback:
User page needs to change "phone number" to "mobile phone number" and put in the line about cost for SMS and using this service.

tonysurma:
@schuback @OhMcGoo can you provide the text for the cost of SMS we should add?

bmaluijb:
I'll take this one

BillWagner:
Fixed in PR #1529

PR #1529, opened 2016-11-19
#1521 Changed Phone number into Mobile phone number in a number of pl…

bmaluijb:
Changed Phone number into Mobile phone number in a number of places and added a static text saying (Standard text message rates apply)


BillWagner:
This looks great @bmaluijb :shipit: now.

PR #1540, opened 2016-11-19
Fixes #1356

### LINKED ISSUE #1356: STARWest_TestLab No Cancel button #1
Description: There is no cancel buttons provided for users

This error will cause users to have to push the back button, thus not continuing in a forward flow to complete a process.

Steps to Recreate: 
Access the web site.
Login as the admin user.
select the admin user name hyper link "Hello starwest.testlab@gmail.com"

Error: There is no cancel button for the user to continue moving forward in the business process.

Suggested Resolution: Provide a cancel button to the user.

screenshot 2016-10-06 at 11 59 35 am

stefan-van-de-griendt-centric:
Added cancel button and handling of it
image

Fixes #1356

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using AllReady.Features.Login;
@@ -53,6 +54,10 @@ namespace AllReady.Controllers
         [ValidateAntiForgeryToken]
         public async Task<IActionResult> Index(IndexViewModel model)
         {
+            // When any other button than the save button is clicked, consider it as a cancel
+            if(!String.Equals(model.Button, "Save"))
+                return RedirectToAction(nameof(Microsoft.EntityFrameworkCore.Metadata.Internal.Index));
+
             var shouldRefreshSignin = false;
 
             var user = await GetCurrentUser();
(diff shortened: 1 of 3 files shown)


AllReadyApp/Web-App/AllReady/Controllers/ManageController.cs
```diff
@@ -104,7 +109,7 @@ public class ManageController : Controller
 
             await UpdateUserProfileCompleteness(user);
 
-            return RedirectToAction(nameof(Index));
+            return RedirectToAction(nameof(Microsoft.EntityFrameworkCore.Metadata.Internal.Index));
```
dpaquette: I'm not sure this namespace should be here

AllReadyApp/Web-App/AllReady/Views/Manage/Index.cshtml
```diff
@@ -202,7 +202,8 @@
             </div>
             <div class="form-group">
                 <div class="col-md-offset-2 col-md-10">
-                    <button type="submit" class="btn btn-default" data-bind="enable: skills.isValid">Save</button>
+                    <button type="submit" name="button" value="Save" class="btn btn-default" data-bind="enable: skills.isValid">Save</button>
+                    <button type="submit" name="button" value="Cancel" class="btn">Cancel</button>
```
dpaquette: If we make this a link back to the index page (instead of a button), we won't need to add any special logic to the controller.

dpaquette:
Update to anchor tag looks good, thanks @stefan-van-de-griendt-centric 

Can we also get those redirects changed back to `return RedirectToAction(nameof(Index));` They shouldn't have the Entity Framework namespace in them

dpaquette:
Looks great! :shipit:

stefan-van-de-griendt-centric:
As requested I've updated the file to now directly use `nameof(Index)`, as it yields the exact same result.
See [nameof reference](https://msdn.microsoft.com/en-us/library/dn986596.aspx) for more details.

PR #1542, opened 2016-11-19
WIP: #1438 fix for EventsByPostalCodeQueryHandler

### LINKED ISSUE #1438: Implement skipped tests in EventsByPostalCodeQueryHandlerShould
@tonysurma please label with jump-in

peerhenry:
Fixes #1438 for EventsByPostalCodeQueryHandler


mgmccarthy:
@tonysurma do you need a code review on this?

tonysurma:
@mgmccarthy yes please

tonysurma:
@mgmccarthy did you get a chance to review?

mgmccarthy:
@tonysurma,  no I did not. I'll try to get to it his weekend.

mgmccarthy:
@tonysurma, we should have never written a unit test for this. The handler does nothing more than wrap the return of a stored procedure, so testing that stored procedure should be an integration test.

@peerhenry, I see what you did to get around the short-comings of not being able to mock .FromSql of the DataContext, but we do not want to encourage passing in dependencies that are only needed for a unit test to a class's constructor. We also want to avoid overloaded constructors that would do the same. My concern here is that `IFromSqlWrapper` will be mis-used by other developers for testing handlers that are not just wrapping a stored procedure.

I'm sorry about this because apparently I created the Issue in the first place. What historically happened here is that we DID have a unit test for this back when we used an interface to abstract AllReadyDataContext... when we moved away from that abstraction and started to inject AllReadyContext directly, that is when this unit test was Skipped when in fact it should have been deleted. It has stayed like this in the code base until now.

@peerhenry, can you please delete the unit test and return `EventsByPostalCodeQueryHandler` to it's original state where it just took AllReadyDataContext?

Also, if there are other unit tests for other handles that wrap stored procedure calls, they should also be delete. @peerhenry, if you want to do this, that would help us out.

There are other jump-in labels out there that can be picked up that I promise are legitimate!

Again, I'm sorry about this mix-up. I should have been more thorough before I created and issue for this looking into the actual implementation of the handler.

tonysurma:
thanks again @peerhenry for the contribution and they 'fun' and dynamic nature of a evolving codebase and @mgmccarthy document.  We appreciate your patience.  As for the update/fix that @mgmccarthy mentioned to unwind some of it, if you want to do that great but if you don't totally understand and we will document that up as an issue for another to jump in on.

Thanks again

peerhenry:
Thanks for the feedback guys. I think you're doing a totally awesome job on
the HTBox, and it's great that you take the time to be the guards at the
gate so to say : )

I'll remove the test and the wrapper, that's really no problem for me.
Before I do that however, allow me to take a shot at convincing you that
this approach does have its value. The test verifies the following:

mockFromSqlWrapper.Verify(w => w.FromSql(It.IsAny<DbSet<Event>>(),
"EXEC GetClosestEventsByPostalCode '{0}', {1}, {2}",
message.PostalCode, 50, message.Distance), Times.Once);

This ensures that the correct stored procedure is called, and with the
expected arguments. As such it improves the safety of the code base.
Without it, all you really know is that the handler is being called with no
guarantee of it executing what it's supposed to execute. FromSqlWrapper
aims to fix that.

Hope that convinces you. If not, no worries. Either way I'll see if I can
pick up another jump-in item soon.

Keep up the good work! Regards, Peer.

mgmccarthy:
cc @MisterJames @dpaquette 

@peerhenry, you do have a point about there still being some value in testing some of the other items you mentioned above.

But, before we discuss that, let's step back. Why are we using stored procedures in `EventsByPostalCodeQueryHandler` and `EventsByGeographyQueryHandler`?

I'm assuming it's because of this type in `GetClostestEventsByPostalCode`?:
`DECLARE @g geography`

the `geography` type?

poking around, it looks like spatial data types are on the roadmap for EF 7 as well as ASP.NET Core:
https://github.com/aspnet/EntityFramework/issues/1100#issuecomment-262329088
https://github.com/aspnet/EntityFramework/wiki/Roadmap
https://github.com/dotnet/corefx/issues/12034

For now, let's stick to the plan of getting rid of the unit tests we have now (the PR code will still exist even if we don't merge it) and see what the deal if with the EF7/.NET Core spatial data types... it might be release 1.2 of EF 7 now it looks like.

dpaquette:
@peerhenry It is true that adding the `IFromSqlWrapper` allows you to verify that the correct SQL String and parameters are being passed in. Unfortunately, there is still no easy way to test the `FromSqlWrapper` implementation since it also suffers from not being able to test that the extension method is called.  Since this is a bit of an edge case in our code base (We don't do a lot of raw SQL or execute many stored procedures), I would tend to agree with @mgmccarthy that we shouldn't be unit testing this. The added complexity of introducing the wrapper is not worth the benefit in this case.

As @mgmccarthy mentioned, this would be better suited as an integration test which would allow us to actually execute the stored procedure against an instance of the database and verify the results. I'm not sure that we have done that yet since our tests all use an In Memory database instance.  It should be possible to create a localdb instance that and execute the tests against that instance.  That approach, in my opinion, would provide more value overall.

IF we were in a situation where we executed lots of raw SQL, I would consider some variation of `IFromSqlWrapper` to try to isolate and allow for unit testing of raw SQL execution. I think for now though, we want to discourage the use of raw SQL.

dpaquette:
@peerhenry if you are interesting in setting up an integration test, let me know and I can try to help. If not, we can track it as a separate issue and get it sorted out later

dpaquette:
Thinking about this today inspired a blog post on the topic of integration testing EF Core + SQL Server
http://www.davepaquette.com/archive/2016/11/27/integration-testing-with-entity-framework-core-and-sql-server.aspx

MisterJames:
@peerhenry Hope all is well and you have seen the comments above. @dpaquette is willing to help get an integration test going if you like, and that is something that can likely be done off of this thread (let's get this one cleaned up/merged first).

Issue #1431, opened 2016-11-02
Implement skipped tests in AddTeamMemberCommandHandlerShould


mgmccarthy:
@tonysurma please label with jump-in

PR #1544, opened 2016-11-19
Implement skipped tests in AddTeamMemberCommandHandlerShould #1431

HamidMosalla:
Made the test `AddTeamMemberCommandHandlerPublishesItineraryVolunteerListUpdatedWhenMatchedOnTaskSignupId` pass, I've added a new `TaskSignup ` to the context b/c ef was complaining that I'm trying to update an entity that didn't exist, and I had to detach it in the test setup because when the handler tries to attach it, I wouldn't get an error related to change tracker...

Closes: #1431


mgmccarthy:
@tonysurma do you need a code review on this?

HamidMosalla:
@mgmccarthy No, the test passes now.

PR #1592, opened 2016-12-01
1349 campaign impact goals

### LINKED ISSUE #1349: Edit Campaign > Add Impact Goals  - Plural descriptor is confusing
screen shot 2016-10-06 at 9 33 55 am

The text description mentions "Goal**s**". However an individual campaign (it appears) may only have 1 goal attached.

I spent some time trying to figure out how to add another goal. My assumption is that an actual org admin would spend similar time and frustration as the app currently functions.

A very simple fix would be to remove the 's' from 'goals'. It may also be useful to add capabilities for multiple impact goals, but I don't know if that is desired behavior.

colhountech:
fixes #1349

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Itineraries/OptimizeRouteCommandHandlerShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Itineraries/OptimizeRouteCommandHandlerShould.cs
@@ -1,0 +1,259 @@
+﻿using AllReady.Areas.Admin.Features.Itineraries;
+using AllReady.Models;
+using AllReady.Services.Routing;
+using Moq;
+using System;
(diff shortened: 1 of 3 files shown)


tonysurma:
I merged this in too quickly and wanted to say thanks @colhountech for contributing and thanks for taking part in our project!  Every contribution counts and please let us know how we can help you contribute more in the future.  Btw, it looks like a commit snuck in from another PR into yours.  I am guessing it had to do with merging changes form a PR since you had checked out the system.

There may be something in this [guide and its linked resources](https://github.com/HTBox/allReady/blob/master/docs/git/gitprocess.md#additional-resources) that is helpful to support the workflow with git we are using and @dpaquette and @stevejgordon would be happy to help with any questions.

Thanks again

colhountech:
Thanks @tonysurma for the guidance. I'll go through your workflow see if I can get aligned better.

PR #1593, opened 2016-12-02
Rename ProviderId to ProviderRequestId on Request model

mgmccarthy:
a simple renaming to make this property intent more explicit.

ProviderId sounds like a unique identifier for the provider of the data, not a unique id per message from that provider.

ProviderRequestId does a better job at conveying the meaning.


tonysurma:
Great point on the name

PR #1594, opened 2016-12-02
Enhance quick add team member UX (#839)

### LINKED ISSUE #839: Enhance quick add team member UX
Follow up from PR #837 

At the moment the structure of the AddTeamMember action and the form which calls it from Itinerary Details does not provide UI feedback in the case of a failure.

We should implement the form post via ajax and handle better display of any errors.

We should enhance the AddTeamMemberCommandHandlerAsync to return details of any failures so that an appropriate error response can be sent by the action.

dchristensen:
Fixes #839


dchristensen:
I'll take a look at the failing tests and update the PR.

dchristensen:
This is ready to go now.

tonysurma:
thank you @dchristensen 

@MisterJames can you give this a quick review again and then we will merge.  Thanks all

Issue #1597, opened 2016-12-02
Organization Request List - Assign One or More Requests to an Event (Step 3)


stevejgordon:
Breakout requirement from the master discussion #1591

STEP 3 - After #1596 is merged

From the "unlinked" Request lister each request should have a checkbox against it. One or more requests can be checked. A button should be available to "link request(s)" which will open a model. Inside the modal a campaign and event (or maybe just event for v1) can be selected. Clicking continue, will update all selected requests with the relevant Event Id

Note: Wording might need to change here. @tonysurma Any preference of what we call a request that has an org but no event? We already use the term unassigned to designate requests under an event which are not in an itinerary. Might be confusing to use it twice?

- UI work to include checkboxes, button and modal
- Controller action to handle the POST event on submiting the event assignment
- Mediatr message and handler which is called from the action to update the records in the DB
- Redirect back to the lister page

tonysurma:
@stevejgordon what is unassigned used for already?

stevejgordon:
@tonysurma We use assigned/unassigned as language to identity whether a request is assigned (or not) to an itinerary.

I think we said we use linked/unlinked for whether a request is linked (or not) to an event.

I think we also suggested a request could be optionally associated (or not) to a campaign as well.

tonysurma:
agree with @stevejgordon 

all requests belong to an org so no need for a word there (aka one cannot exist without org assignment)

requests are then linked/unlinked to events

requests are then assigned/unassigned to an itinerary.

no words needed for campaign because the link is direct to event.

Thanks!

stevejgordon:
@tonysurma just to clarify on:

> all requests belong to an org so no need for a word there (aka one cannot exist without org assignment)

If we manually create a request via the UI it is linked to an event. This means we can navigate to the original Org via event > campaign > org, but in the db we don't assign an Org Id value (it's null).

On the API we currently hard code the Org ID to 1 for v1 and Red Cross. This is because we don't know the event to link them to.

On the CSV you can now select the event to add them to. Again as with manual requests these don't have their Org Id set in the DB but we can navigate through the layers to find the Org.

We used the Org Id only for API as a means to associate the request to the Org which we can work out from the API token. So you're correct that all requests belong to an Org in the system. The means to navigate to the Org do differ. We could solve this by ensuring we always set the Org ID from the API and Manual UI process also? At least then we have consistency. If you agree, I can open a couple of small issues to cover this.

tonysurma:
[...]

Issue #1596, opened 2016-12-02
Add Org Level Page To List Requests (Step 2)


stevejgordon:
Breakout requirement from the master discussion #1591

STEP 2 - After #1595 is merged

We need an admin page that can be accessed from the organization admin homepage/dashboard via a link. This page will list any requests which have an org id matching that organization and where there is no event id currently set.

- Add a view and action for the request lister
- Add a Mediatr query message and handler to get the required view model for the page
- Add a link from the Org homepage which links to this new page. Ideally include the count of unlinked requests so an admin knows there are requests to link.

chris-sargood:
WIP

stevejgordon:
@tonysurma @MisterJames  - We have a quick query on this. @chris-sargood has made good progress with an initial page to list the requests which currently he's added via a link on the Org details page (admin area).

However, currently that page is only accessed by SiteAdmins (in fact all Org admin pages are only open to the site admins). Therefore we need to work out the best controller to locate this action on, and where to present the link to an Org Admin so that they can access their unlinked requests.

One option would be to make modifications to the Org admin area so that an Org admin can access their Org Details page (view only). This is not too complex but would require changes to the flow from the lister onwards including the checks to ensure org admins only see their org's data. 

Another would be to put the unlinked request lister on it's own controller and route, and find somewhere sensible to locate a URL for the Org Admin for now. For v1 it might be best to go with this as the quickest option and just add a link from the menu for org admins which just takes them to unlinked requests. However this might be confusing as it's not all requests. Any linked requests would be managed via the Itinerary level admin pages.

Thoughts?

cc @mgmccarthy

mgmccarthy:
@stevejgordon, I vote for this option:

>  to put the unlinked request lister on it's own controller and route, and find somewhere sensible to locate a URL for the Org Admin for now.

we should be aiming for smaller controllers in general, and I think this would be fine for something to keep around post v1. Thanks for bringing the question up.

tonysurma:
agree with @mgmccarthy own controller and lets make sure org admins can access.  thanks

stevejgordon:
@tonysurma @mgmccarthy - Sounds good to me.

@chris-sargood Can you make a new controller with the relevant action to handle the route /Admin/UnlinkedRequests. This can be the same code (I think) as you showed me yesterday. We just need to ensure requests are filtered only to those for the Org id of the user (if an org admin). I can't remember your code 100%, you might already have that!).

[...]

Issue #1595, opened 2016-12-02
Add OrganizationId to the EF Request Model (Step 1)


stevejgordon:
Breakout requirement from the master discussion #1591

STEP 1

For v1 release we have agreed that on creation/import each request will be assigned an organization id to link it to an allReady Organization (i.e. Red Cross).

We need the model updating and a migration created to add this property and a foreign key back to the Organization.

stevejgordon:
We'll hopefully tackle this during our code-a-thon on Monday 5th

chris-sargood:
Work in progress

Issue #1591, opened 2016-12-01
Idea/Discussion: Do not associate a Request to an Event when creating Requests from API/CSV sources


mgmccarthy:
Breaking the comments about this from #1590 and #1587 into this separate issue.

So far, the ideas that have been floated:
from @MisterJames in [1590](https://github.com/HTBox/allReady/issues/1590#issuecomment-264197849):
>  I'm not sure that I'm opposed to orphaned requests, but there are two ways that come to mind immediately to solve this.
 > 1. If we decide that orphaned requests are okay, we simply display a warning/alert at the top of every event page (for admins) and prompt them to "assign" the requests to the event. This could be done via a list with a filter set at the top/selection. This would be more work, but might allow finer-grained control.
> 2. If we prefer direct assignment, we could add an event drop down as part of the CSV import process. This would be trivial but would mean that all requests would have the same event assignment (which is likely okay).

with @stevejgordon giving a +1 to option 2

from @tonysurma in [1587](https://github.com/HTBox/allReady/issues/1587#issuecomment-264191816)
> @mgmccarthy interesting idea and then have the admin also have an 'associate' request just like the create request on the event. We could at least dump the requests under the organization at ingestion then so that Tony Org can't see requests that came in to Steve Org for example.

> Should we continue discussion here and change name (again :smile_cat: ) or move to another issue?

from @stevejgordon in [1587](https://github.com/HTBox/allReady/issues/1587#issuecomment-264198419)

> tonysurma Sounds sensible to have somewhere we can view "orphaned" requests. We could have a few issues and possibly address this at the code-a-thon. Something like

> Add Org level page (or maybe campaign level if we can at least determine campaign via API) to list requests without an event id set
Clicking on a request displays a page/modal allowing an campaign and event to be set
If that sounds reasonable I can break out some issues along those lines

> For CSV as it's a manual process to import I presume on the import screen we can/or should be able to choose the event we're importing under? With the assumption that data is validated by the admin prior to importing?

MY two cents:

- for @MisterJames option 2. This could be a good solution for CSV import, but will not work for API import
- for @tonysurma in his comment:  "We could at least dump the requests under the organization at ingestion then so that Tony Org can't see requests that came in to Steve Org for example".  The problem here is when we import a Request from CSV/API, we lack the context of Organizations/Campaigns/Events, etc... all we have to go on is the raw data in the Request. The most useful information to us is the geographic information we receive.
[...]

PR #1598, opened 2016-12-02
RequestApiController and Invocation of GASA's endpoint for status update

mgmccarthy:
High level details:

- `RequestApiController` is the endpoint that receives the API request from GASA
- after field validation, and making sure we're not about to import a dupe, we enqueue the  `IProcessApiRequests` Hangfire job, which immediately returns control to the caller so we can return a 202 to GASA, and kick of a pipeline process on our end.
- `ProcessApiRequests` inserts the Request into the db and publishes `ApiRequestProcessedNotification`
- `ApiRequestProcessedNotificationHandler` will be the code responsible for deciding whether or not we can service this request. I'm not too sure is want to defer writing the Request to our database until we know we can service the request, but until we can figure out the mechanism/rules for making that decision, all Requests are written to the db. `ApiRequestProcessedNotificationHandler` then enqueue's the 
`ISendRequestStatusToGetASmokeAlarm` Hangfire job. By enqueue'ing the job here, we allow immediate release of the thread, and scope invocation of GASA's endpoint to a single hangifre job which will handle failures and retries
- `SendRequestStatusToGetASmokeAlarm` uses HttpClient to invoke GASA's endpoint to communicate the acceptance of the request back to their system. **NOTE: right now, this is wired up to send requests back to their demo server. If you're going to test importing a Request via the `RequestApiController`, please remember that you will be invoking an RPC on the back end.**

Still a couple things up in the air:

- Issue #1591 directly impacts the ability of what to send to GASA's endpoint when we receive a request (true we service the request, false we do not service the request) now that we'll be mapping incoming Requests to their owning OrgId's and building a UI from there that will allow users to associate a request to an Event (see Issues #1595, #1596 and #1597 for details). 
- I'm waiting to hear back from GASA about potentially sending them a false initially for every Request we import from them, but when a Request is associated with an Event, to then send a true. (https://github.com/redcross/smoke-alarm-portal/issues/196#issuecomment-264457770)
- RequestApiController is still not using this route `[Route("api/request")]` until Issue #1586 is figured out. When that is working, we can get the correct route on this endpoint controller


mgmccarthy:
@tonysurma, argh... non-deterministic unit test failure! FYI, this is in the latest master branch, not an artifact of my PR.

`TaskStatusChangeHandlerShould.VolunteerCannotCompleteTaskFromAssignedStatus` runs fine on my local machine but b/c of how the test is written, it fails on a slower build server.

I'll see if I have time to get a fix in for it...

mgmccarthy:
@MisterJames @stevejgordon cc @tonysurma is this going to be looked at or merged, or is the code in this PR incorrect as a result of the conversations around assigning an org to an incoming API Request?

tonysurma:
@mgmccarthy - what is throwing me off (and maybe others) is the 'WIP' in the title.  I had assumed it wasn't ready to go yet given that...

mgmccarthy:
@tonysurma, it is WIP b/c of two things:

1. pending the decision made on Saturday after the standup if we're going to be associating an API Request to an Org, which looks like from one of the PR's merged from yesterday's codeathon, is good to go.
2. pending a discussion with @cecilia-donnelly regarding [my last comment](https://github.com/redcross/smoke-alarm-portal/issues/196#issuecomment-264457770) on [Issue 196](https://github.com/redcross/smoke-alarm-portal/issues/196) about what we can report back to GASA when we assign an API Request to an Organization. One possibility is: we return a false, then when we associate a Request to an Event, we invoke their endpoint again and return true. Problem is, I don't know if their code can handle this change in workflow. For now, the code in this PR invokes their API with a status of false when we import an API Request, and does not assign it to an Org nor associated it with an Event. I can easily change the code in a separate PR to assign an Org to the Request now that we have the OrgId/Org props on the Request model. Also, we're currently invoking their demo server, so nothing is for "real" yet as far as how this is wired up.

Given the two things above, I think it's still safe to merge this, but pending a discussion with GASA, we don't have the "report status back to GASA" 100% done and I still need to add some code on this end assigning an Org to the Request.

tonysurma:
And you may have already done this but lets log two issues for those two items and then we can consider the 'feature' to be WIP but your PR is not WIP and is good to go for this round.  I think that is an okay project pattern to use do you agree?  cc @stevejgordon related to chat on saturday

mgmccarthy:
@tonysurma, let me break those two items out into separate Issues, I'll link the other issues/items to them. Thanks for the help!

tonysurma:
I am going to merge as is and we will track via additional issues as per above

PR #1615, opened 2016-12-08
Added filtering to request list on itinerary details.

### LINKED ISSUE #1100: geocoded requests: filter by status and other items already on screen
sub issue of #1080

dchristensen:
Fixes #1100

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Itinerary/Details.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Itinerary/Details.cshtml
@@ -1,4 +1,6 @@
-﻿@model AllReady.Areas.Admin.ViewModels.Itinerary.ItineraryDetailsViewModel
+﻿@using AllReady.Services
+@model AllReady.Areas.Admin.ViewModels.Itinerary.ItineraryDetailsViewModel
+@inject AllReady.Services.ISelectListService SelectListService
 
 @{
     ViewData["Title"] = Model.Name;
@@ -94,7 +96,7 @@
 
 <div class="row">
     <div class="col-md-12">
-        <h3>
+        <h3 id="requests">
             Requests
         </h3>
     </div>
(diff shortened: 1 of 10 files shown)


tonysurma:
@stevejgordon can you give this a quick review, @dchristensen thanks

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Itinerary/Details.cshtml
```diff
@@ -125,6 +127,20 @@
         </div>
     </div>
 }
+<form method="post" action="@Url.Action("Details", "Itinerary", new { area="Admin" })#requests" class="form-horizontal">
```
stevejgordon: There seem to be a couple of problems with this form. Firstly it's not sending an anti forgery token so we get blocked by MVC (bad request) on the POST. The best option to fix this is to use the form taghelpers which will actually render an anti-forgery token for us automatically.

```
<form asp-action="Details" asp-controller="Itinerary" asp-area="Admin">
    // You inputs etc here
</form>
```

Secondly, I'm not clear why the #requests string is in there currently?

dchristensen: @stevejgordon The `#request` is being appended to the URL (and the tag helpers are not being used) so that the posted URL ends up as `/Admin/Itinerary/Details/5#requests`, for example.   I'm adding the target hash because when the filter button is clicked and the page reloads you are left back at the top of the page.  This felt like a bad experience for the user, especially if there are several team members assigned to the itinerary pushing the requests further down the page.  This addition makes sure the the requests are shown at the top of the screen when the page loads. 

I can just add in the older style `@Html.AntiForgeryToken()` inside the form unless you have a suggestion for an alternative solution.

dpaquette: I have not tried adding a hash to a form action using the tag helper approach. If it doesn't work then adding `@Html.AntiForgeryToken()` should be sufficient here.

tonysurma:
@stevejgordon can you take a re-review? looks like @dchristensen updated post your comments.  Thanks

PR #1621, opened 2016-12-10
Changed "var task" to "var theTask".

TomWalkerCodes:
Issue 1387 - Hackathon London .net Developers

PR #1624, opened 2016-12-10
Fix for "var task" to "var theTask" #1621 

TomWalkerCodes:


mgmccarthy:
this should really be `var @task`. I'm assuming this change is being made b/c `task` is a reserved word in .NET?

Check out how we use `var @event` in the code.

PR #1625, opened 2016-12-10
Fix for "var task" to "var theTask" #1621 

TomWalkerCodes:


mgmccarthy:
Thanks for the PR and the last minute change to 'var @task' @TomWalkerCodes.

PR #1633, opened 2016-12-13
Itinerary request lasso selection

### LINKED ISSUE #1101: geocoded requests: tools for visual selection (lasso, click individual)
sub issue of #1080

dchristensen:
Closes #1101

I found as I was working on this that there are two different versions of the Bing Maps API in use in the application.  I upgraded this map from v7 to v8 because I am using some of the functionality in v8, but the _EventScripts partial is still referencing the older v7 API.

It would also be fairly simple now to make the search feature on this page all client side since the list is rendered via knockout now.


tonysurma:
@stevejgordon can you take a look at this?

Issue #1245, opened 2016-09-19
Site Admin action icons are not obvious what they do


digitaldrummerj:
From a user experience standpoint, the icons that are on the site admin page are not obvious without having to rollover them.  Personally I prefer to have some kind of text along with the icon (if there even is an icon) so that I don't have to rollover them to figure out what they are.  

Also, if you are on a mobile device using the website, there is no way to rollover them.

Here are the icons that I am talking about.  You can view the page locally at [http://localhost:48408/Admin/Site](http://localhost:48408/Admin/Site)

![image](https://cloud.githubusercontent.com/assets/708423/18652457/7de195d6-7e87-11e6-9271-1445f736aaef.png)

smahesh5:
Cannot open the link above. Error: "Web page does not exist"

digitaldrummerj:
@smahesh5 that is the localhost url when you are running the site out of Visual Studio

YannickMeeus:
I've had a stab at replacing the icons with text, but I'm not convinced as to how it makes it better. It makes it clearer but it feels it detracts from not just the feel of the menu but also from the overall feel of the page. It becomes a lot more cluttered as a result:

![admin pane - text instead of icons](https://cloud.githubusercontent.com/assets/588607/19908457/57f26a56-a07b-11e6-85fc-8e4a4c39bdfe.png)

Any opinions? I can shove this branch up to my fork if people are interested in the code changes, but it'd be cool to just get some feedback on the visuals of the page.

digitaldrummerj:
@YannickMeeus I agree that it makes it way more obvious as to what each button does now but makes the page look a bit more cluttered.  I do like that at least they are obvious now where before you had no clue what the buttons did.  With a little bit of UX love I think you could make it work.  Might just have to rethink how the page works.

some potential ideas:

* hide the buttons by default with an action button that pops the buttons down into view
* make multiple rows of buttons
* move the buttons except edit and delete into an edit dialog.  Then clicking edit would open the edit dialog and have all of the additional buttons

tonysurma:
From a user perspective options 1 and three would be nice

YannickMeeus:
I'm looking at option three right now.

[...]

PR #1635, opened 2016-12-13
Changes for the Site Admin buttons

YannickMeeus:
Applies to #1245.

Better late than never! So the changes are as follows:

- Swapped the columns around, it felt better to have the email on the left-hand side and the buttons on the right-hand.
- Swapped all icons for the action titles themselves.
- Moved all buttons apart from two into a modal popup.
- Renamed the Edit button within the modal to "Manage Skills" as that's the only thing that linked page seems to allow you to do.
- Centered the table (3 columns on each side for padding and 6 columns in the grid allocated to the table)
The Page after the changes:

the page itself

The modal when the Edit button is clicked:
the modal when visible

Another issue will have to be raised to change the "Admin/Site/EditUser" page, as it's still branded as a generic "Edit" page.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Site/Index.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Site/Index.cshtml
@@ -7,6 +7,7 @@
 <h2>User Admin</h2>
 <div class="container-fluid">
     <div class="row">
+        <div class="col-md-3"></div>
         <div class="col-md-6">
             <table class="table table-responsive table-condensed">
                 <tr>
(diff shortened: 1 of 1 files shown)


YannickMeeus:
I just noticed, by sheer fluke, that there are some other modals in the system that are styled differently from what I've done. Is there a styling guideline or an example modal I should base my styling on?

tonysurma:
Thanks @YannickMeeus !  Love seeing this UX update

@dpaquette I don't think we have a style guide for this but do you have any basis by which the previous ones have been designed or patterns we should follow?

dpaquette:
@tonysurma I don't think we were using any particular style guides or patterns. It just kind of evolved the way it did as more and more people worked on it. 

I do like this design as it removes any ambiguity in terms of what the actions. Sometimes words are better than icons.  Moving them into a modal makes sense too since it avoids the problem of having too many action buttons in the grid. A dropdown button might have worked for this too.

Overall I think this is a great change. I will keep an eye out for some good UX guidelines that we can reference but I don't have anything off the top of my head

tonysurma:
Thanks @dpaquette

Issue #1405, opened 2016-10-27
Build out SmsResponseController


mgmccarthy:
pinging @MisterJames @tonysurma @stevejgordon for feeback on if the assumptions I'm making below are correct.

This Issue is expanding upon the following line items from Issue #1084: 
- If request is confirmed add "'N' if you can no longer make it" and if they respond N mark unassigned
- If request is not confirmed add "'Y' if you can make it, 'N' if you can't" and set status as above

allReady sends sms confirmation messages to requestors whose Request has been Assigned to an Itinerary. The following code sends these messages:

```
RequestsAssignedToIntineraryHandler
SendRequestConfirmationMessagesAWeekBeforeAnItineraryDate
SendRequestConfirmationMessagesADayBeforeAnItineraryDate
SendRequestConfirmationMessagesTheDayOfAnItineraryDate
```

The requestor can reply "Y" or "N" to confirm the request. This Issue includes the following work:
- identify how to attach the RequestId from the Request model to the outgoing sms messages. The RequestId has to survive the round trip of being sent out (by the classes listed above) and then trip back to SmsRequestController.
- in SmsRequestController, we look up the Request by RequestId. We **cannot** correlate the incoming request  from the Requestor by PhoneNumber as the Requestor could have more than multiple requests that have been added to different Itineraries. If we looked up the requestor by phone number, we would end up with two requests, so we would not know which request to mark as Confirmed.
- there is code already written in SmsRequestController that needs to be uncommented based on the Requestor's reply of "Y"/"N"
- most likely, this will involve Twilio to a greater extent than how we currently use it (which is just to send sms messages). You will need to set up a Twilio account and either use your mobile phone or open a Google Voice account to test the receiving of text messages as well as the reply of "Y"/"N" to those text messages.

tonysurma:
Those assumptions look correct to me

mgmccarthy:
whoever picks this up, please ping me first to answer any questions you may have

mgmccarthy:
@tonysurma this might need to be updated to P1. If we're sending out sms confirmation messages to requests we receive via the getasmokealarm portal, the requestor's will be unable to reply to our text message to confirm or cancel the request.

EmilyLuijbregts:
@tonysurma -shall i update the label to P1? or is it OK to leave as P2?

stevejgordon:
[...]

PR #1664, opened 2016-12-20
WIP - Adding notifications table

### LINKED ISSUE #1647: Adding table to store details of sent notifications
Breakout from #1405

We need to store details of the notifications we have sent from allReady to requesters. This will enable SMS response handling as well as serving as a useful history type table that we can expose where needed in the UI.

stevejgordon:
This will fix #1647 but needs a little review and further work after I clarify some points with @mgmccarthy.

I have added a notification table to store details of all notifications we send from the application. I have also added a notificationhandler design to handle messages whenever we send a notification. I have tried to make those messages quite easy to reuse and build out using some static methods.

I have linked this into the SendInitialRequestConfirmationsHandler where we fire off initial messages to requesters. This works locally as I expect.

First question for @mgmccarthy and @MisterJames is for any comments on the code so far. The static methods for building up the message and as a helper to convert to a EF model record are a bit different to what we've done elsewhere but I hope, allow us to ensure rules a bit and hide unnecessary internals.

Other questions for @mgmccarthy are around the other three notification points...
You have the classes such as SendRequestConfirmationMessagesADayBeforeAnItineraryDate which are triggered via hangfire. These are currently void methods and are not awaiting for example SendSmsAsync which I'm not sure if gives us any issues?

1. Can we use Mediatr within them? I'd like to leverage my AddNotificationsRecord which is a Mediatr IAsyncNotificationHandler to update the database. Can I just take a dependency on Mediatr in the constructor as normal? Not sure if there's any hangfire specific rules?
2. Assuming we can use Medaitr, can we make change them from regular void methods to async Tasks so that I can await the PublishAsync I'd need to use? Again, not sure if hangfire is okay with that?

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Notifications/SendInitialRequestConfirmationsHandler.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Notifications/SendInitialRequestConfirmationsHandler.cs
@@ -4,6 +4,7 @@ using AllReady.Models;
 using AllReady.Services;
 using MediatR;
 using Microsoft.EntityFrameworkCore;
+using AllReady.Features.Notifications;
 
 namespace AllReady.Areas.Admin.Features.Notifications
 {
@@ -22,12 +23,23 @@ namespace AllReady.Areas.Admin.Features.Notifications
 
         public async Task Handle(RequestsAssignedToItinerary notification)
         {
-            var requestorPhoneNumbers = await context.Requests.Where(x => notification.RequestIds.Contains(x.RequestId)).Select(x => x.Phone).ToListAsync();
+            var requestData = await context.Requests
+                .Where(x => notification.RequestIds.Contains(x.RequestId))
+                .Select(x => new {
+                    phone = x.Phone,
+                    requestId = x.RequestId
+                }).ToListAsync();
+
             var itinerary = await context.Itineraries.SingleAsync(x => x.Id == notification.ItineraryId);
 
             //TODO mgmccarthy: need to convert itinerary.Date to local time of the request's intinerary's campaign's timezone
-            await smsSender.SendSmsAsync(requestorPhoneNumbers, 
+            await smsSender.SendSmsAsync(requestData.Select(x => x.phone).ToList(), 
                 $@"Your request has been scheduled by allReady for {itinerary.Date.Date}. Please respond with ""Y"" to confirm this request or ""N"" to cancel this request.");
+
+            foreach(var item in requestData)
+            {
+                await mediator.PublishAsync(NotificationSentMessage.SmsMessage(NotificationMessageType.InitialItineraryAssignment, item.phone, requestId: item.requestId, responseRequired: true));
+            }
 
             await mediator.PublishAsync(new InitialRequestConfirmationsSent { ItineraryId = itinerary.Id, RequestIds = notification.RequestIds });
         }
--- a/AllReadyApp/Web-App/AllReady/Features/Notifications/AddNotificationsRecord.cs
+++ b/AllReadyApp/Web-App/AllReady/Features/Notifications/AddNotificationsRecord.cs
@@ -1,0 +1,23 @@
+﻿using AllReady.Models;
+using MediatR;
+using System.Threading.Tasks;
+
+namespace AllReady.Features.Notifications
+{
+    public class AddNotificationsRecord : IAsyncNotificationHandler<NotificationSentMessage>
+    {
+        private readonly AllReadyContext _context;
+
+        public AddNotificationsRecord(AllReadyContext context)
+        {
+            _context = context;
+        }
+
+        public async Task Handle(NotificationSentMessage message)
+        {
+            _context.Notifications.Add(Notification.FromAddNotificationLogMessage(message));
+
+            await _context.SaveChangesAsync();
+        }
+    }
+}
--- a/AllReadyApp/Web-App/AllReady/Features/Notifications/NotificationSentMessage.cs
+++ b/AllReadyApp/Web-App/AllReady/Features/Notifications/NotificationSentMessage.cs
@@ -45,8 +45,11 @@ namespace AllReady.Features.Notifications
         /// <summary>
         /// Creates a new <see cref="NotificationSentMessage"/> for an Email Message
         /// </summary>
-        public static NotificationSentMessage EmailMessage(NotificationMessageType messageType, Guid? RequestId = null, string userId = null, bool responseRequired = false, string email = null)
+        public static NotificationSentMessage EmailMessage(NotificationMessageType messageType, string email, Guid? RequestId = null, string userId = null, bool responseRequired = false)
         {
+            if (string.IsNullOrEmpty(email))
+                throw new ArgumentNullException(nameof(email));
+
             return new NotificationSentMessage
             {
                 Method = NotificationMethod.Email,
--- a/AllReadyApp/Web-App/AllReady/Features/Notifications/NotificationSentMessage.cs
+++ b/AllReadyApp/Web-App/AllReady/Features/Notifications/NotificationSentMessage.cs
@@ -1,0 +1,61 @@
+﻿using AllReady.Models;
+using MediatR;
+using System;
+
+namespace AllReady.Features.Notifications
+{
+    /// <summary>
+    /// A message containing the properties for a notification which has been sent by the system
+    /// </summary>
+    public class NotificationSentMessage : IAsyncNotification
+    {
+        public NotificationMessageType MessageType { get; private set; }
+        public NotificationMethod Method { get; private set; }
+        public string Email { get; private set; }
+        public string Phone { get; private set; }
+        public bool ResponseRequired { get; private set; }
+        public DateTime SentDateTime { get; private set; }
+        public Guid? RequestId { get; private set; }
+        public string UserId { get; private set; }
+
+        private NotificationSentMessage()
+        {
+            SentDateTime = DateTime.UtcNow;
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="NotificationSentMessage"/> for an SMS Message
+        /// </summary>
+        public static NotificationSentMessage SmsMessage(NotificationMessageType messageType, string phoneNumber, Guid? requestId = null, string userId = null, bool responseRequired = false)
+        {
+            if (string.IsNullOrEmpty(phoneNumber))
+                throw new ArgumentNullException(nameof(phoneNumber));
+
+            return new NotificationSentMessage
+            {
+                Method = NotificationMethod.Sms,
+                MessageType = messageType,
+                ResponseRequired = responseRequired,
+                Phone = phoneNumber,
+                UserId = userId,
+                RequestId = requestId
+            };
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="NotificationSentMessage"/> for an Email Message
+        /// </summary>
+        public static NotificationSentMessage EmailMessage(NotificationMessageType messageType, Guid? RequestId = null, string userId = null, bool responseRequired = false, string email = null)
+        {
+            return new NotificationSentMessage
+            {
+                Method = NotificationMethod.Email,
+                MessageType = messageType,
+                ResponseRequired = responseRequired,
+                Email = email,
+                UserId = userId,
+                RequestId = RequestId
+            };
+        } 
+    }
+}
--- a/AllReadyApp/Web-App/AllReady/Migrations/20161220185618_AddNotifications.Designer.cs
+++ b/AllReadyApp/Web-App/AllReady/Migrations/20161220185618_AddNotifications.Designer.cs
@@ -1,0 +1,1037 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Migrations;
+using AllReady.Models;
+
+namespace AllReady.Migrations
+{
+    [DbContext(typeof(AllReadyContext))]
+    [Migration("20161220185618_AddNotifications")]
+    partial class AddNotifications
+    {
+        protected override void BuildTargetModel(ModelBuilder modelBuilder)
+        {
+            modelBuilder
+                .HasAnnotation("ProductVersion", "1.0.2")
+                .HasAnnotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn);
+
+            modelBuilder.Entity("AllReady.Models.AllReadyTask", b =>
+                {
+                    b.Property<int>("Id")
+                        .ValueGeneratedOnAdd();
+
+                    b.Property<string>("Description");
+
+                    b.Property<DateTimeOffset>("EndDateTime");
+
+                    b.Property<int>("EventId");
+
+                    b.Property<bool>("IsAllowWaitList");
+
+                    b.Property<bool>("IsLimitVolunteers");
+
+                    b.Property<string>("Name")
+                        .IsRequired();
+
+                    b.Property<int>("NumberOfVolunteersRequired");
+
+                    b.Property<int?>("OrganizationId");
+
+                    b.Property<DateTimeOffset>("StartDateTime");
+
+                    b.HasKey("Id");
+
+                    b.HasIndex("EventId");
+
+                    b.HasIndex("OrganizationId");
+
+                    b.ToTable("AllReadyTask");
+                });
+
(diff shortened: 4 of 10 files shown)


mgmccarthy:
@stevejgordon 

> Other questions for @mgmccarthy are around the other three notification points...
You have the classes such as SendRequestConfirmationMessagesADayBeforeAnItineraryDate which are triggered via hangfire. These are currently void methods and are not awaiting for example SendSmsAsync which I'm not sure if gives us any issues?

Since Hangfire does not support an async API and b/c ISmsSender only exposes an async interface (and this is driven by two layers "down", where we wrap `CloudQueue` with out `IQueueStorageService`), I don't really have much of a choice. When invoking an async API in a synchronous method, the call runs synchronously, but as you pointed out, could this be a problem? I'm not too sure.

I do know that `CloudQueue` does offer an `AddMessage` API that does not appear to be Task oriented...so in theory, we could add that to our API offering on `QueueStorageService` and allow ISmsSender to have a non-async API for any invocations in any Hangfire jobs.

@MisterJames @dpaquette, do you have any feedback on this?

As far as using MediatR, you can publish a notification from the jobs, but publish them using MediatR's synch API, and know by tying MediatR into the job, now TWO things have to complete successfully in order for the job to succeed:

1. the sms's need to be sent with no exceptions
2. the notification you're publishing needs to be handled successfully (aka, written to the notifications table) with no exceptions or the job will retry

#2 is is the important one:

if you hit #2 above as a result of the notification/handler that "logs" sent messages to the notifications table, you run the risk of the job retrying, and all the sms messages re-sending. The way you would prevent this is to code reads at the beginning of the job to look up previously sent sms messages in the notifications table and filter out any sms message form this query:
```
var requestorPhoneNumbers = context.Requests.Where(x => requestIds.Contains(x.RequestId) && x.Status == RequestStatus.PendingConfirmation).Select(x => x.Phone).ToList();
```

why is this? Because we're combining a non-transactional resource (azure storage/queues) with a transactional resource (the db) in the same logical transaction.

The other option is that instead of using MediatR to publish an event out of the job, you instead Enqueue a new Hangfire job that writes to the notifications table. That makes the "logging" aspect 100% asynchronous.

mgmccarthy:
@stevejgordon, I've just had a once over of the PR. 

My first question is, are we logging ALL published notifications, or just the ones that are going out as sms confirmation messages? If we're logging all, I'd rather not unless we have a business reason to do so. I know we have a business reason to log the outgoing sms confirmation messages so we can take our best guess at correlating the incoming sms response from the requestor to the correct request. If we can just keep it to that, I think that would be a better place to start.

Either way, I think there might be an easier/better way to handle the logging of the different notifications we need using something called "polymorphic messaging"... apparently, MediatR supports this concept.

I'm going to look into it later tonight.

mgmccarthy:
@stevejgordon, some more feedback...

The reason we're adding the Notification table to the system is to help us correlate the outgoing request confirmation sms messages when requestors reply to them to confirm/cancel a request. The minimal amount of data we need to do this is:

- Id (a PK field)
- RequestId
- MessageType/Name (initial request, seven day, one day, etc...)
- DateTimeSentUtc
- Phone

I could be missing something here, but I think that's it.

We probably don't need Email, b/c when we move to email, the user will most likely click on a link we send them in the email to bring them to a page that will allow them to confirm or cancel the request. Or the email has a "confirm" or "cancel" link that does the same thing.

That link can contain a unique id (Guid or whatnot) that uniquely identifies the email message and we have a table that links that unique id to the RequestId on our end so we know which request to update when we hear back from the requestor. Basically, email does what SMS won't let us do, and that's embed correlation data directly into the message

That being said, recording just those four pieces of data when we these four things happen:

 - InitialRequestConfirmationsSent
- SendRequestConfirmationMessagesSevenDaysBeforeAnItineraryDate
- SendRequestConfirmationMessagesOneDayBeforeAnItineraryDate
- SendRequestConfirmationMessagesTheDayOfAnItineraryDate

should be enough to get us started to know which Request to look up and then set to Confirmed.

The second part of this is the publishing of notifications you mentioned earlier. Right now, you're publishing `NotificationSentMessage` from `SendInitialRequestConfirmationsHandler`. That's heading in the right direction, but I'd rather use polymorphic messaging for this.

What this means is we have a base "message" class the other messages inherit from, and then a handler that handles that "base" message. When any of the inheriting messages are published, that base handler is invoked as well.

So, a quick example of notifications we'd want to publish:
- InitialRequestConfirmationsSent (we already have this one)
  (we'd have to add the ones below)
- SevenDayRequestConfirmationsSent
- OneDayRequestConfirmationsSent
- DayOfRequestConfirmationsSent

each of these messages inherit from a new message class called `RequestConfirmationsSent`
```
public class RequestConfirmationsSent : IAsyncNotification
{
    //we can either look up the phone numbers for these RequestIds in the RequestConfirmationsSentHandler or send the phone numbers and the requestids on the message together
    public List<Guid> RequestIds { get; set; }
}

public class RequestConfirmationsSentHandler : IAsyncNotificationHandler<RequestConfirmationsSent>
{
    public async Task Handle(RequestConfirmationsSent notification)
    {
        //we log the info here
    }
}
```

instead of `InitialRequestConfirmationsSent` looking like this:
`public class InitialRequestConfirmationsSent : IAsyncNotification`

it now looks like this:
`public class InitialRequestConfirmationsSent : RequestConfirmationsSent`

now when we publish `InitialRequestConfirmationsSent`, both `SendInitialRequestConfirmationsHandler` and `RequestConfirmationsSentHandler` are invoked in the MediatR pipeline, with the newly added `RequestConfirmationsSentHandler` doing our logging work for us.

what we've done is essentially separate the logging concern out of the classes that send out the sms confirmation messages.

As usual, this is longer than I wanted. Let me know if you want to chat about this. There is still the problem of publishing MediatR notifications from the Hangfire jobs, but I think that's something we can get around for now based on our discussion this afternoon.

stevejgordon:
@mgmccarthy Thanks as always for your valuable feedback. A lot of what you've suggested/raised makes sense.

> My first question is, are we logging ALL published notifications, or just the ones that are going out as SMS confirmation messages? If we're logging all, I'd rather not unless we have a business reason to do so. I know we have a business reason to log the outgoing SMS confirmation messages so we can take our best guess at correlating the incoming SMS response from the requester to the correct request. If we can just keep it to that, I think that would be a better place to start.

There may be an element of YAGNI in what I've done so far. The main thing as you state is the SMS messages in order for us to correlate the replies. However, I can imagine a situation where it would be useful to know who has been notified for request scheduling, regardless of the method. At this stage all that's really specific to that would be a couple of columns which we don't have to use and wire up to emails unless we need/want to. @tonysurma - Do you have views on wider logging of email notifications and whether it would be useful for org admins to view any communication history (email and SMS) for a request / user?

One field I have included is the reponseRequired. If we go down the route of only logging the confirmation type messaging then we don't need it, but if we may also send out general info messages, it might be useful to exclude them from any searches when handling SMS responses callbacks. Again, possibly not going to need it yet.

> The second part of this is the publishing of notifications you mentioned earlier. Right now, you're publishing NotificationSentMessage from SendInitialRequestConfirmationsHandler. That's heading in the right direction, but I'd rather use polymorphic messaging for this.
> 
> What this means is we have a base "message" class the other messages inherit from, and then a handler that handles that "base" message. When any of the inheriting messages are published, that base handler is invoked as well.

To the above quote and general points around where to do the work to record the log to the DB, I share some of the concerns about causing issues with the message sending (retires etc) from issues in the DB. The polymorphic messages sounds interesting, but, do we run the risk of logging to the DB that we've sent a message, but in fact if that message send handler failed for some reason, we find the message did not actually send! I'd rather only log send messages once we know the message has been at least placed on the queue. Perhaps we wrap our db calls in a try/catch to instead log an error but not break the SMS sending flow? We then run the risk of not logging everything we'd need to identify a reply?

I'm wondering if a better integration point for our logging might actually be within the WebJob that handles sending the SMS. One advantage is that we can be sure we're logging at the time of actually sending an SMS, rather than when we hope to have triggered it. Another bit advantage would be that we can consider recording the SID from Twilio and using their API to get updates around that message. This would allow to query if it's been received or not. At the moment we have to trust that the number we're given is correct, valid and a mobile/cell number. Using Twilio we could get a better idea of any messages which fail to get to the requester, possibly allowing some more manual followup. 

This idea is without some issues we have to tackle also. Firstly we'd need to pass down the message type and request ID into the SMS queue so that the WebJob can use the data to record the DB record. Secondly, we need to be able to access the DB from the WebJob. @mgmccarthy - I believe you were seeing some challenges there? We might be able to forgo the use of EF for this type of logging and instead fall to a general SQL command?

So, that's a long reply to some long feedback! Let's discuss and try to pick the best path forward that meets the requirements and gives us some flexibility.

To the point about identify messages that we've touched on (in issue #1405) ie. working out what reply is for what actual message. One idea I've had that might work is to require different response wording for each message (if we can think of enough). So first message might ask for Y or N response. Second message might ask for YES or NO response, third might be CONFIRM or CANCEL etc. That way, we can imply from the response content, the notification that the user is responding to? It's not foolproof but could help us a little. @tonysurma - Any views/concerns on this idea?

mgmccarthy:
@stevejgordon, for this:

> However, I can imagine a situation where it would be useful to know who has been notified for request scheduling, regardless of the method. At this stage all that's really specific to that would be a couple of columns which we don't have to use and wire up to emails unless we need/want to. 

check out this blog my Jimmy Bogard about building a MediatR pipeline for cross-cutting concerns:
https://lostechies.com/jimmybogard/2014/09/09/tackling-cross-cutting-concerns-with-a-mediator-pipeline/

if we want to add logging for EVERY outbound notification, this is an ITOps/cross-cutting convern that runs across all handlers that are doing this type of work. So the technical approahc to solving that issue is much different than the specific issue we're dealing with for correlating incoming sms confirmation responses.

> One field I have included is the reponseRequired. If we go down the route of only logging the confirmation type messaging then we don't need it, but if we may also send out general info messages, it might be useful to exclude them from any searches when handling SMS responses callbacks. Again, possibly not going to need it yet.

I understand what the field is for now. The thing is, if the table is specific just to solving the response by the requestor/correlation problem, we know all those messages require a response, so in theory, we could still get by w/out the field

>  The polymorphic messages sounds interesting, but, do we run the risk of logging to the DB that we've sent a message, but in fact if that message send handler failed for some reason, we find the message did not actually send!

we do, but this is tied to the implementation of publisning notifications out of meditR handlers. The one way we could get around this is to still have the base class handler handle RequestConfirmationsSent  as outlined above, but from this handler, instead of calling directly into the db to log the message info, we instead create a new hangfire job, enqueue it from that handler, and that way, we don't need worry about a logging failure causing a retry in the sending of the sms messages.

> I'm wondering if a better integration point for our logging might actually be within the WebJob that handles sending the SMS. 

this is a **very good idea**! the problem is in order to get access to AllReadyContext from the web job project, we know need to share the context across mutliple projects. @dpaquette is actually looking into the work that would need to be done to manage a single context across mutliple projects. Of course, we could have a new context, just for logging, taht the web job project would own, and we could do it that way.  the downside is now, how do we query that data from the web project?

> Another bit advantage would be that we can consider recording the SID from Twilio and using their API to get updates around that message.

**I like this idea even more.** B/c we're technically addressing the problem at the root cause, which is SMS messages do not allow us to "embed" data. I realize the work would be Twilio-specific, but this might be the best avenue to look at, b/c we truly can act on the unique SID, instead of "guessing" with most recent message sent out of the notifications table.

> One idea I've had that might work is to require different response wording for each message (if we can think of enough). So first message might ask for Y or N response. Second message might ask for YES or NO response, third might be CONFIRM or CANCEL etc. 

**I like this idea best**. Simplest to implement, b/c we can now uniquely identify the message.

One more idea is an endpoint per message type.... but I'm not too sure how that would work.

I might be online later this afternoon where we can chat, but I'm pretty busy today w/the family as we both have the day off and are spending it together.

Thanks for all the great thought and work put into solving this problem @stevejgordon

mgmccarthy:
WOAH, I did not mena to close This!

mgmccarthy:
sorry, I did not mean to close this....

stevejgordon:
> I like this idea even more. B/c we're technically addressing the problem at the root cause, which is SMS messages do not allow us to "embed" data. I realize the work would be Twilio-specific, but this might be the best avenue to look at, b/c we truly can act on the unique SID, instead of "guessing" with most recent message sent out of the notifications table.

Unfortunately that doesn't fully help ID the response. We can make the same assumption as before which a slight improvement in that if the message log shows a notification has not been received yet, we can assume that the reply is not for that message and try the previous one.

> I like this idea best. Simplest to implement, b/c we can now uniquely identify the message.
> 
> One more idea is an endpoint per message type.... but I'm not too sure how that would work.
Yeah this gives us some more control over identifying the response. We can have multiple Twilio callback endpoints, but only one per number we register with them. So we'd have to pay for 4 numbers for 4 distinct endpoints.

mgmccarthy:
@stevegordon, u are correct about the "like the idea best" comment. It does not address the unique message id w/out the lookup table

tonysurma:
> Do you have views on wider logging of email notifications and whether it would be useful for org admins to view any communication history (email and SMS) for a request / user?

We don't have a requirement for that at the moment (cc @OhMcGoo @schuback) but I could imagine it coming up quickly

stevejgordon:
@tonysurma Thanks. I'd propose I stick with the schema I have for now (saves redoing it in the future) but we will just not actually use the columns until it becomes a requirement. If we review and decide it's not needed for vNext then it's quite easy to migrate the unused columns away.

mgmccarthy:
@stevejgordon @tonysurma, I just want to say one more time that if we're looking to log all outbound communication sent from AllReady, then we should be looking into a different technical implementation than the implementation Steve is working on for the explicit problem of being able to correlate incoming SMS confirmation messages for Requests, regardless of where we write the logging data to.

We want to be careful about violating SRP here....

for "overall" outbound logging of communication message, I think building up a MediatR pipeline to handle cross cutting logging concern like Jimmy Bogard outlined in this blog:
https://lostechies.com/jimmybogard/2014/09/09/tackling-cross-cutting-concerns-with-a-mediator-pipeline/

would be a good thing. We can use marker interfaces, or come up with some type of convention to identify which messages constitute "outbound communication" here.

Also, the best approach for overall logging is to come up with something on the WebJob side. Since both email and sms are being sent from web jobs, it's possible we could use another database to log this information (which, again, pretenses the problem of not being able to have one context shared across multiple projects in allReady).

Steve, up to you if you want to try out the polymorphic messaging stuff. I understand if you do not, as we're so close to the launch and we need to have this problem solved.

We can always re-visit/refactor in the future.

Thanks!

stevejgordon:
@mgmccarthy - I'm working on this right now and have been thinking things through. I'm looking to try and go minimal and I'm thinking perhaps we can do without the table for the v1 release and just correlate the number we get the response from and the most recent request that has a matching number.

Can I confirm - A yes is a yes in our system isn't it? i.e. once confirmed either on the first message or the followups, we do not send further messages requiring a Y/N reply? And a no, would trigger the request to not be accepted and again stops the notification cycle. So know exactly which message is being responded to is crucial for now perhaps.

I'm looking at trying to at least get the handler prepared which takes the reply from Twilio and updates the status of a matching request. That's the key thing for now. Adding more to it and making it more robust later is I think very much needed, but might delay and over complicate it right at this moment. It feels like a bit more brainstorming to get it right would be good before we engineer too deep.

mgmccarthy:
@stevegordon

>  A yes is a yes in our system isn't it? i.e. once confirmed either on the first message or the followups, we do not send further messages requiring a Y/N reply?

Steve, you are correct. We don't care which message it is (initial, 7 day, 1 day), we just care that the requestir confirmed the request. You are also correct that once a SMS message has been confirmed, we do not send any other SMS confirmation messages that might still be in the "pipeline" of messages to send.

tonysurma:
@stevejgordon should I pull this in (when rebased?) wasn't sure on status?

mgmccarthy:
@tonysurma (cc @stevejgordon)

I think the decision here is NOT to merge this. Steve came up with a much simpler approach after we chatted a couple days ago...

this PR and the Issue with it can probably be moved to the backlog and picked back up after the v1 release.

@stevejgordon, does this sound correct?

stevejgordon:
@mgmccarthy Correct, I'll kill this PR for now.

@tonysurma - For v1 we're comparing against the phone number stored against the request. That was the quickest option to get us going and start testing the flows. The requirement to store sent SMS messages in a table would still be useful as a post v1 goal. 

a) it would provide history of the messages sent (might be something we want to expose in the UI for admins).
b) we can save some additional meta data from the Twilio send message response (if we use the API directly rather than the client wrapper).
c) we could include some logic to call the Twilio endpoint after a message is sent to try and get an updated status. Twilio allows you to query for the message status if we store the SID. So we can see if a message fails to deliver for example. This might be useful?

I propose we keep the related issue open, but updated and on the backlog for post v1.

We will also add an issue to include the API call to Twilio to verify the format of the phone number on request creation which is a more pressing requirement.

PR #1666, opened 2016-12-22
#1596 - Added new 'unlinked requests' option to the 'Admin' menu

### LINKED ISSUE #1596: Add Org Level Page To List Requests (Step 2)

chris-sargood:
I've added the following new option to the admin menu:
image

This option should only be visible to OrgAdmins and it should only show requests that are at 'unassigned' status for the users current Org ID.

One thing I noticed when I was testing my rebase onto master, was that when I entered a Request via the campaigns/event screen, it doesn't automatically assign an orgid. To test I had to manually assign one via SQL. I don't think that having requests assigned an orgid automatically was part of this issue but I just wanted to make a note of it here!

@stevejgordon It may be worth you having a quick look at this to ensure I got the right idea from your requirements.

Closes #1596


stevejgordon:
@chris-sargood Thanks very much. I'll try to pull this later today or tomorrow and give it a review. I think the org id not being set from the creation of a request is expected since it's explicitly tied to an event so can be linked back, but there have been a few threads on this so I need to review and confirm the final behaviour.

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/UnlinkedRequests/List.cshtml
```diff
@@ -0,0 +1,42 @@
+@using AllReady.Areas.Admin.ViewModels.Request
+@model AllReady.Areas.Admin.ViewModels.Request.UnlinkedRequestsViewModel
+
```
stevejgordon: Need to add a ViewBag.Title for the page which renders in the browser tabs please.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/UnlinkedRequestsController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            var orgId = User.GetOrganizationId();
+
+            var viewModel = new UnlinkedRequestsViewModel()
+            {
+                Requests =
+                    await
+                        _mediator.SendAsync(new RequestListItemsQuery()
+                        {
+                            Criteria = new RequestSearchCriteria() { OrganizationId = orgId, Status = RequestStatus.Unassigned }
```
stevejgordon: It's slightly confusing as we have a lot of statuses kicking around but in this case we want to return all requests for a particular Organization, where they do not currently have an EventId. This is how we would classify the unlinked requests. The RequestStatus is used for tracking status within the event management to identify assignment to itineraries. I'll include a bit more in the summary review.

chris-sargood:
@stevejgordon 
I've made the following changes:
1) The UnlinkedRequest controller now queries using eventId = null
2) The tab title is now "Unlinked Requests"
3) I've removed the hyperlinks out of the unlinked request list

tonysurma:
thanks again @chris-sargood 

@stevejgordon can you take another look?

chris-sargood:
@stevejgordon 
I've just squashed my commits to make things a bit neater. Can you pull this again before you review?

Issue #1575, opened 2016-11-29
Edit Itinerary - Enhance - Improve UX


stevejgordon:
If the use start address as end address is checked we should hide the end address fields on the form.

joro550:
Hey there, I was going to lok to do this, but it seems that this allready happens, someone at some point added some javascript at the bottom of the cshtml to allow this to happen, couple of points on this page though.

* Having the class "form-control" on the checkbox makes it really big on the screen but very difficult to click for me - not sure if this is just my browser though, I had to click the top edge of the box overwise it wouldn't see it as me clicking the box
* Information is mis-spelt on the page - small thing

I have a branch fixing these two smaller points, not sure if you want to fix the first issue or whether I should open a seperate card - I'll wait for feedback

stevejgordon:
Hi @joro550. Thanks for taking a look at this. You are correct the show/address was closed in #1631. That was my mistake for not linking back to this original master issue.

I agree that the checkbox is a bit large so if you have code to improve that and the spelling error, please do submit those as a PR. Thanks for contributing.

EmilyLuijbregts:
@joro550  - thanks for picking this one up! Just wanted to check in and see if you had any update? or if you needed any support?

joro550:
@EmilyLuijbregts There was a pull request that was merged a few days ago #1575

EmilyLuijbregts:
Great- thanks - i'll close this issue.

PR #1671, opened 2016-12-27
#1575 - Improving Edit itinerary UI

joro550:
Making some tweaks to the ui so the checkobox is more easily clickable (and much smaller), I also made a change so "address line 1" would apear on one line and not two.


tonysurma:
Thank you for the contribution, happy new year!

PR #1674, opened 2016-12-29
Initial code for handling Twilio responses

### LINKED ISSUE #1669: Add Twilio Notification Confirmation Endpoint
Breakout from #1405

We should add a public endpoint for the Twilio notification reply handling. For v1 we'll target the simplest solution which works to confirm/reject request scheduling by requesters.

Links to #1647 which may also be required to form the complete solution.

stevejgordon:
Fixes #1669 

The Twilio response handling has opened a lot of process and technical questions. In a bid to simplify for v1 this initial code should address exposing a public endpoint that Twilio can POST to in order to send us any SMS responses for our SMS notifications.

This code uses the from phone number provided by Twilio to attempt to find a matching request that includes the same number. In an attempt to simplify we have used the most recently added request should we match more than one request.

If we can find the request we update it based on the answer from the sender. We send them a final message also in either case.

Currently acceptance sends "Thank you for confirming you availability."
Declining a scheduled request sends "We have canceled your request and once it is rescheduled you will receive further communication."

@tonysurma - Not sure what actual messaging you want here?

One issue with this code currently is around Twilio's use of number formatting. Twilio will expect and return numbers in the E.164 format (https://support.twilio.com/hc/en-us/articles/223183008-Formatting-International-Phone-Numbers).

Currently we might not store our numbers in this format and so it'll be hard to match up responses to the original request. There are various potential solutions to this which I feel should be follow on issues.

1) We use the Twilio free lookup API to lookup numbers when requests are added (manual, csv and API). This will return an internationalised number based on a phone number and country code. We could use this lookup and update the request with the correctly formatted number to make future matching easier.
2) We store the sent SMS messages in a table (as planned in #1664) which should include the number twilio returns from it's API when sending the request. This number should be formatted to E.164 already so if we store this, we should be able to match up properly.

Ultimately I think we should do both to try and make this as robust as possible. Starting with part 1 feels reasonable. It will also somewhat help us validate numbers are valid and then handle those requests as appropriate (i.e. if the lookup fails we cannot use SMS as a communication method for this request and fall back to email). There is an additional pay per use option where we could validate if the number is landline / mobile etc as well. This would be useful but adds cost. $0.005 per request.

@tonysurma - Any thoughts on what is minimal needed for v1? At the moment I think we'll fail to match a lot of the responses unless we store the correct Twilio formatted number when adding the requests or store the number when we send SMS via the API.

Regardless of these questions, this handler should need minimal change so can perhaps be merged.

@mgmccarthy - I moved the controller out of the admin area since it's publicly exposed and renamed to SMSReponseController. Is that okay for you? Any other thoughts after our Skype today? I've not changed much except moving the DB logic into a Mediatr handler.

If this looks good I can try to add a few tests (or stub an issue to follow up with those).


mgmccarthy:
@stevejgordon thanks for all the work on this, I think you're approach is right on target for v1 release. Moving and renaming the controller is good to go.

@tonysurma, Steve and I had a chance to Skype and chat today about the implementation of the request confirmation functionality as it relates to the messaging medium (aka, sms vs. email). For a variety of reasons (some voiced in other issues and some that Steve has brought up here), using sms for the messaging introduces problems... 

- we're "guessing" at the most recently sent Request matched by incoming phone number, which could lead to us confirming or canceling the wrong request if the requestor has put it two requests, and each request would be added to a different event. b/c sms does not let us embed a unique id for the request that will go out with the confirmation message, then come back to us with the requestor's reponse, we have no way to fix this problem using sms. At least not yet.
- we need to now "scrub" incoming phone numbers using a Twilio lookup service b/c of number formatting. To get around this, I'd like to enforce the E.164 format on Request creation, regardless of the source (UI, CSV, API). That way, we don't have to validate later in the pipeline. We send the number out in the correct format, and the response back to us will be in the correct format (unless I"m missing something here @stevejgordon?)
- we probably should verify the number we've received is a mobile phone number, and not a land line, or the requestor will never receive the initial confirmation message or any of the follow up messages. To do this, it sounds like there is a fee involved, and I don't know who/what will be paying the bill for this service.

Using email for communication could make these problems go away, but there will be work involved in doing so. Between the research that Steve and I have done, if you're okay with the caveats we've outlined above, I'm good to go with Steve's approach as it's simple and gets the job done for v1.

As Steve and I also discussed today, maybe seeing the system used in the wild would help us make better changes to how we manage the requestor communications...

Thanks

stevejgordon:
Thanks @mgmccarthy for the feedback. I'm working to finish this off and add some tests so that this can be merged it @tonysurma agrees.

stevejgordon:
@tonysurma - I'm taking the WIP off this as it should be good to go for v1, pending your thoughts to my earlier questions and caveats!

mgmccarthy:
@stevejgordon, looks like you're still using the `IChangeRequestStatus` abstraction that used to be wrapped by `Hangfire.Enqueue()`.

The only reason that abstraction was created is b/c Hangfire needs a job abstraction in order to allow injection of dependencies into the job code...

Since `IChangeRequestStatus` is sync, not async, then we might as well use MediatR here. Can you send `ChangeRequestStatusCommand` for both the Confirmed and Unassigned status changes? That way we can get rid of `IChangeRequestStatus` and its implementer.

stevejgordon:
@mgmccarthy - Yep, makes sense - Will refactor that before we merge this

tonysurma:
thoughts here:

1. for content of messages - go with what you have and we can log issues to change it as needed later
2. for phone number storing - I agree we need to correlate so we should do the phone number lookup/formatting via twilio as you suggested
3. correlating for two requests - for now we are good and that is not a use case I would worry about at this time, may change and we can deal with it later (I have seen some patterns with SMS to avoid this)
4. verifying it is mobile - I *believe* this is something that twilio API can do as well.  If so that 'cost' is part of what we are using twilio for (along with sending and receiving of SMS) and that cost is discounted as a non-profit and something we assume to be an operating cost of the application so we are okay

Did I capture all of the needed points?  THanks @stevejgordon and @mgmccarthy for staying on this!

mgmccarthy:
@tonysurma @stevejgordon 

> 2. for phone number storing - I agree we need to correlate so we should do the phone number lookup/formatting via twilio as you suggested

again, I'd like to "front" this validation to Request creation. When we create a Request (regardless of source), we do this lookup and make sure the phone number is validated/formatted properly before we write it to the Request table.

That way, we don' have to worry about re-validating later... once the data is in our system, then we know it's good to use to send sms messages via Twilio.

Since the `Phone` property on the `Request` table will always need to be a mobile number based on how we're sending out confirmation messages initially, then following up with the requestor when we don't hear from them, I think fronting the validation/Twilio call would be a good idea.

mgmccarthy:
@stevejgordon 

> Since IChangeRequestStatus is sync, not async, then we might as well use MediatR here. Can you send ChangeRequestStatusCommand for both the Confirmed and Unassigned status changes? That way we can get rid of IChangeRequestStatus and its implementer.

setting a Request to Unassigned involves more than just a status update. We also need to delete the corresponding entry in ItineraryRequest table. 

`ChangeRequestStatusCommandHandler` should be updated to handle the removal of this row only for request's whose status is changing to Unassigned.

Thanks.

stevejgordon:
> again, I'd like to "front" this validation to Request creation. When we create a Request (regardless of source), we do this lookup and make sure the phone number is validated/formatted properly before we write it to the Request table.
100% agree. Doing it up front is the correct place. We might need to consider what happens to a request when the number does not parse / is landline etc. Do these go into a bucket for org admins to review, are they immediately automatically rejected etc.

stevejgordon:
@mgmccarthy - I've removed the unnecessary hangfire code as per your feedback above.

PR #1679, opened 2016-12-31
Ensuring the OrgId is set on every request

### LINKED ISSUE #1667: Ensure all requests have an Org Id set in the DB (CSV and Manual Creation)
For consistency we should update where appropriate to ensure that we always set the Org ID when adding requests via the CSV and Manual UI screens? Currently in those cases we set the Event directly.

As we expect requests to have an Org Id when added via the API it'll ensure that all requests then have a direct relationship to the organization.

stevejgordon:
Adding code so that manually created and CSV imported requests have the OrgId set directly. This adds consistency since all API added requests ill already have the OrgId set.

Also made the ImportRequestsCommandHandler async while working in that area

Fixes #1667

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/ImportControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/ImportControllerTests.cs
@@ -17,6 +17,7 @@ using Xunit;
 using System.Linq;
 using AllReady.Areas.Admin.Features.Import;
 using Microsoft.Extensions.Logging.Internal;
+using System.Threading.Tasks;
 
 namespace AllReady.UnitTest.Areas.Admin.Controllers
 {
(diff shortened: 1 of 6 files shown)


mgmccarthy:
@stevejgordon, is this something we want to do? 

Knowing that the OrganizationId "fix" was for v1 and came about as a workaround for our inability to be able to assign a Request to an Itinerary (b/c we can not figure out yet how to associated an incoming API Request to an Event). I think leaving the OrgId as null for Requests that are created via CSV or UI is okay, b/c this "solution" could be removed later on when we come up with a better one.

I don't know what the value it in assigning the OrgId to the CSV and UI Requests, b/c the OrgId is only used to funnel the API Requests into the "holding area/UI" for Event association. for CSV/UI Requests, taking on the OrgId is just extra data that is not needed, and more data to clean up later when we move away from this approach to processing incoming API Requests...

thoughts?

tonysurma:
The thought was to be consistent regardless of source so yes we want to do this

stevejgordon:
@mgmccarthy - As per @tonysurma this is just a consistency thing for now.

PR #1690, opened 2017-01-06
Ensure only future or current featured campaigns are shown

### LINKED ISSUE #1689: Closed Featured Campaign is returned on homepage
It's possible for the site to blow (as it did for me locally) if a campaign is set as featured but has passed it's end date. The campaign is returned but not found in the active campaigns list and the removal throws an error.

We should filter the search to include only featured campaigns which are active.

stevejgordon:
Fixes #1689


chinwobble:
should the canopy integration tests also be updated to check for this?

stevejgordon:
@chinwobble - Are you referring to the f# project? I've never loaded that before, so not sure how current it is I'm afraid. I believe it's UI testing which I don't think we'd need here as we just need to ensure our query only includes the correct data.

tonysurma:
@chinwobble thanks for looking in.  Those tests do need maintenance but that is seperate from this.

PR #1693, opened 2017-01-08
Twilio Lookup to Validate Phone Numbers

### LINKED ISSUE #1680: Validate Request Phone Numbers on Creation
Breakout from #1405 

To ensure that phone numbers are both valid and also stored in the correct E.164 format we should utilise the Twilio lookup API during request creation (API, CSV and Manual). To do this we will need to utilise the Twilio Lookup Tool at https://www.twilio.com/lookup.

We need to do this to ensure we can properly match SMS responses to the originating request. As the returned SMS response phone number will be the E.164 format, ensuring we store this with the request at creation will be a requirement for v1 to function properly.

On creation via any method, phone numbers should be posted to the lookup and where valid, saved in the proper format within the DB (instead of the original input). Unmatched numbers should be handled appropriately (TBD)

It might also be worth us including a test endpoint with the app to use in developer mode to simulate possible responses? Although that might be too much for v1 it's worth thinking about. For now I could create a faked service implementing the same interface which we can use when this is in development mode.

Questions for @tonysurma 

1. The lookup tool appears to be a separate service. The basic lookup is free but the type lookup (landline vs mobile) is chargeable. From your previous reply on the earlier issues I believe you think this may be included with the allReady/Htbox account or something the project can fund. Are you able to test if your account has the access? It'll require specific code if we intend to use the type information.
2. What would we want to do for CSV/API requests where the number is not valid (Twilio cannot process)? Should these be immediately rejected / not imported? Or should they be stored with some other status for review? Is possible a human may need to validate oddly formatted numbers that Twilio can't handle. For manually created requests we can probably include a UI validation message for phone numbers we can't process. At the moment we require a phone number for all requests. Is it possible that those without a valid phone number might need to be stored and handled in the future via email?
3. Similar to q2. If the number is a landline, how would we handle the new request creation?

I'd be happy to pick this one up next since it's close to what I've worked on so far. I had a very quick play with the API when looking into other issues.

cc / @mgmccarthy as once we clarify the above it might be worth putting heads together on this before I get coding.

stevejgordon:
Fixes #1680

This code adds a service wrapper around the Twilio Rest API client. It uses the Lookup service to validate phone number formats and types.

It also ensures we save the correct E.164 formatted number to ensure our SMS response code can find matching requests.

Includes a faked version of the service for dev use - doesn't require Twilio account.

Currently a WIP for initial feedback and discussion. Needs to be implemented for CSV and manual request creation + unit testing and general checking through. Also needs some code to convert Country strings to ISO format country codes.

Tested against the live Twilio service using a real number and it worked as expected.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/RequestApiControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/RequestApiControllerTests.cs
@@ -11,6 +11,7 @@ using MediatR;
 using Microsoft.AspNetCore.Mvc;
 using Moq;
 using Xunit;
+using AllReady.Features.Sms;
 
 namespace AllReady.UnitTest.Controllers
 {
@@ -41,6 +42,8 @@ namespace AllReady.UnitTest.Controllers
             const string providerRequestId = "ProviderRequestId";
 
             var mediator = new Mock<IMediator>();
+            mediator.Setup(x => x.SendAsync(It.IsAny<ValidatePhoneNumberRequest>())).ReturnsAsync(new ValidatePhoneNumberResult { IsValid = true, PhoneNumberE164 = "0000" });
+
             var sut = new RequestApiController(mediator.Object, Mock.Of<IBackgroundJobClient>());
             await sut.Post(new RequestApiViewModel { Status = "new", ProviderRequestId = providerRequestId });
 
@@ -65,7 +68,10 @@ namespace AllReady.UnitTest.Controllers
         public async Task PostEnqueuesProcessApiRequestsJobWithCorrectViewModel()
         {
             var viewModel = new RequestApiViewModel { Status = "new" };
+
             var mediator = new Mock<IMediator>();
+            mediator.Setup(x => x.SendAsync(It.IsAny<ValidatePhoneNumberRequest>())).ReturnsAsync(new ValidatePhoneNumberResult { IsValid = true, PhoneNumberE164 = "0000" });
+
             var backgroundJobClient = new Mock<IBackgroundJobClient>();
 
             var sut = new RequestApiController(mediator.Object, backgroundJobClient.Object);
@@ -79,7 +85,10 @@ namespace AllReady.UnitTest.Controllers
         [Fact]
         public async Task PostReturns202StatusCode()
         {
-            var sut = new RequestApiController(Mock.Of<IMediator>(), Mock.Of<IBackgroundJobClient>());
+            var mediator = new Mock<IMediator>();
+            mediator.Setup(x => x.SendAsync(It.IsAny<ValidatePhoneNumberRequest>())).ReturnsAsync(new ValidatePhoneNumberResult { IsValid = true, PhoneNumberE164 = "0000" });
+
+            var sut = new RequestApiController(mediator.Object, Mock.Of<IBackgroundJobClient>());
             var result = await sut.Post(new RequestApiViewModel { Status = "new" }) as StatusCodeResult;
 
             Assert.IsType<StatusCodeResult>(result);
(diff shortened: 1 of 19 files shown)


mgmccarthy:
@stevejgordon I'm going to pull this locally tonight or tomorrow and take a look. I'll get back to you. Thanks!

mgmccarthy:
@stevejgordon, the code looks solid, thanks for figuring all this Twilio stuff out. A couple questions:

1. do you need me to take care of updating the CSV request import and UI request creation with your Twilio functionality?
2. do you know if there is a request limit on this Twilio API (for example, Twilio only supports "x" amount of API invocations for a given IP address over a time period of "y") where we would need to worry about throttling the amount of api invocations for csv and api request imports?
3. Do you think that the Request model/entity should be updated to include more than one phone number field? For one Request, depending on the integration, we might be given both a landline and a mobile number for example. In the future, I know we want the request confirm process to be robust where we fallback to different communication types if one type "fails" or we might not have the data on the incoming request to try different types (for example, if we're not getting a landline, we can't fall back to a phone call via an IVR system)

wondering what your thoughts were.

Thanks!

stevejgordon:
Thanks @mgmccarthy for the review. On the basis of your thumbs up I'll look to get this rebased and a few tests added so that I can take this out of WIP.

Answers to your questions

1. That would be great. I'll leave that to you once this is finalised and merged in if that's okay. I can create an issue to cover that work.
2. I wasn't able to find any rate limits listed for the lookup API. The docs were a little slim, or I was looking in the wrong place.
3. It's possible we'll need a landline, but I guess until we have functionality around it / a new organization which needs it for integration it's a case of YAGNI. @tonysurma - Any requirements for this in the near future do you think?

tonysurma:
For two, you are good, we are charged per call so they are okay with all but crazy levels of calls. 

For three, stay as is and keep simple, new requirements can come later, if they do at all for that.

stevejgordon:
Thanks @tonysurma - I'll get this finalised for inclusion ASAP

stevejgordon:
Actually, I found this info:

> Twilio's Lookup API currently only handles a single request per phone number. If you have a batch of several phone numbers you would like to lookup, we recommend that you make multiple requests to Twilio's API. Please keep in mind that each Twilio account may only make 100 concurrent API requests per second.

So on CSV we need to be wary of it. Otherwise I think we're okay.

tonysurma:
@stevejgordon ah interesting, they have a slightly different plan for the lookup API then.  Agree only on CSV and then that is still pretty high so just need to watch for batches greater than 100 calls.  Thanks for researching.

stevejgordon:
@tonysurma no problem. The easiest option might be to wrap the code with Polly; a retry framework and then if we get any failed responses from the endpoint we can try again after a short delay. Covers us for short outages also.

mgmccarthy:
@tonysurma 

> or two, you are good, we are charged per call so they are okay with all but crazy levels of calls.

any example of "crazy level of calls" is something imports a CSV of request with tens of thousands of rows... we don't throttle nor offload asynchronously any of this work, it's API invocation for each request in a for loop.

stevejgordon:
@mgmccarthy I've refactored slightly for testing and added the tests. I think this is good to go for now. Can open a new issue to consider CSV import rate limiting.

@tonysurma If Mike's happy then this is ready to merge. Removed the WIP title.

mgmccarthy:
@stevejgordon @tonysurma I'll see if there is a way to "throttle" the processing of CSV imports for Requests when I plug the `ValidatePhoneNumberRequest` into the CSV code.

stevejgordon:
@mgmccarthy Was thinking the same thing!

tonysurma:
@stevejgordon @mgmccarthy - good to merge?

mgmccarthy:
@tonysurma yep

PR #1695, opened 2017-01-10
Moving to .NET Core (v 1.1)

### LINKED ISSUE #1637: Target .NET Core and Version 1.1.0
To enable Mac developers to work on this project it would be nice to move over to targeting the .NET core framework.

@tonysurma - I believe some work was fairly close for that?

I can see one potential issue where we rely on System.Drawing to generate a QR Code for the mobile app. I'm not sure if that functionality is used / needed at the moment? If we need this, then we are tied to the full framework or need to look at any .NET core libraries that may be available.

At the same time, perhaps we review moving to the 1.1.0 release train to get the latest features and improvements?

Throwing this out as a question so we can determine if this should be addresses pre v1. At least the Mac support would be nice. I have one Mac user coming to the code-a-thon in January at NDC so would be good to have support for it if possible.

stevejgordon:
Fixes #1637

This is initial code that will take us away from the full .net framework and allow the projects to be built cross-platform. I have switched us to .net core 1.1 and as a result removed some non compatible dependencies and refactored code where the API surface changed.

I've cleaned my commits, but kept them quite verbose since it'll help see exactly what changed here.

With these changes I'm able to build and run the site use dotnet.exe + run the unit tests etc.

If anyone has a Mac and can pull this locally, would be good to know if it works as expected.

@tonysurma - I presume the Azure website can run a .net Core app?

If at all possible it would be nice to get this finalised and merged in this week, pre NDC if we're happy to do that? There are some risks though so if we feel safer waiting until after we can do that also (although rebasing will become more and more nasty!) One of the devs joining me is a Mac user so would be helpful if we could run native, rather than in a Windows VM.

Things we loose:

- I have removed SMTP based code we were using purely to create a dev version that could send to local SMTP server during dev. We already have a fake SMTP client which write to the output so this isn't crucial. SMTP was removed in .NET core and while there are replacements, they are quite heavy for this single use.
- I have removed the XZing dependency that was generating a QR code for the phone app. This was reliant on System.Drawing which is not in .net core.

Not yet done:

- The service replacements for Twitter and Geocoding work locally for me. This code is initial WIP and I will be refining it further, adding some comments and also adding tests. 

Not done (might need help):

- I've never run the F# Scenario testing - I'm not sure if that will still work or not. Would welcome input on that if we need it running for v1.

@mgmccarthy and @MisterJames - Would welcome some local testing and a few sets of eyes on this since it's a fundamental change.


tonysurma:
> @tonysurma - I presume the Azure website can run a .net Core app?

Yes that is true/works so we are good there, I will probably do a test deployment to another website just to see what if anything needs updating but yeah it should be good

mgmccarthy:
@stevejgordon, I have eyes on this now.

mgmccarthy:
@stevejgordon, I didn't realize you could do this:
```
var request = new HttpRequestMessage(HttpMethod.Get, requestUrl.ToString());
request.Headers.TryAddWithoutValidation("Authorization", string.Concat("Bearer ", bearerToken));
response = await _httpClient.SendAsync(request);
```

you're passing in the http auth headers on the HttpRequestMessage, instead of setting the auth information on HttpClient itself. This might help me with a problem I've been having at work.

Have you tested the sign-on auth via Twitter at all, or did you just refactor the code?

mgmccarthy:
@stevejgordon, wow! This is a HUGE effort/amount of work. After eyeballing it and walking through your commits, everything looks great.

I have not actually stood it up and smoke tested it yet (at least the parts that have the old API removals and the new API additoins)....

did you want me to use the site a bit before I give the thumbs up on the merge?

stevejgordon:
@mgmccarthy - Yep tested it through and I got the name details back from Twitter. Nothing through for email, but it doesn't look like that is always exposed (unless you found otherwise). Even with a manual call to Twitter I couldn't get one back.

r.e. the headers, yep it's useful since you can keep a static HttpClient around without it being tied to a specific use. I tend to build up the request with the headers it needs and then use a general client. I think this could help with the SendRequestStatusToGetASmokeAlarm also?

stevejgordon:
@mgmccarthy - Thanks, it went a little deeper than I expected to get everything up and running. I've stepped through some flows but if you're able to pull it and validate a) it builds for you and b) you can load the site that would be good for peace of mind since it's a big change.

I'm going to try and test as many flows manually before finally finishing it up. But if the code looks good I can begin on some unit tests for my services.

mgmccarthy:
> I think this could help with the SendRequestStatusToGetASmokeAlarm also?

you are correct, it could. Let me take a crack at that refactoring. I'll see if I can get what I need from building up the HttpRequestMessage and passing it to the IHttpClient held as a singleton with StaticHttpClient implementor.

mgmccarthy:
> Nothing through for email, but it doesn't look like that is always exposed (unless you found otherwise). Even with a manual call to Twitter I couldn't get one back.

at this point, I've given up on Twitter and the user's email. Twitter makes it so impossible to get what should be a simple piece of data for a user from it's API, I'm willing to live with the email being non-retrievable when a user is logging in via Twitter (@tonysurma, I assume you agree?)

mgmccarthy:
@tonysurma, I want to have a chance to stand this up and excersise the changed part of the system before i say good to go for the code review. Give me a day or two and I'll let you know. I know that steve will also be smoke testing on his end as well

YannickMeeus:
@stevejgordon Has anyone tried standing up everything on a Mac yet, as otherwise I could try and give that a go one of these days?

stevejgordon:
@YannickMeeus - It's not been tried yet so yes that would be great if you're able to, thanks. Would be great to validate that it gives the desired result. If you need any advise to be able to pull down this code, let me know.

tonysurma:
@YannickMeeus that would be great, we have done it in parts and repeatedly run into issues but hopefully this PR with maybe some added bits) will clear that up.  I and others have done it in the past and we kept stumbling on the issues 'fixed' in this PR

mgmccarthy:
@stevejgordon, I just tried to stand the site up and can't. I ran dotnet restore, rebuilt. When I pick IE for my brower the window opens, then immediately closes. When I pick Chrome for my browser, the site spins up then stops and I get this from Chrome:
image

my guess is I"m doing something wrong. Do I need to download/install anything new on my system in order to run the code in this PR?

stevejgordon:
@mgmccarthy - I think that's the fairly un-guessable sign that you might be missing the latest 1.1 SDK. If you go to https://www.microsoft.com/net/download/core#/current and download the latest SDK I hope you'll be good to go.

stevejgordon:
I think it might throw something into the output window to that effect. But otherwise it's pretty unhelpful!

mgmccarthy:
@stevejgordon , got the SDK squared away. Just finished running through invocation of IGeoCodeService and everything worked great.

FYI, I used the same Google API key for:
```
"Mapping": {
    "GoogleDirectionsApiKey": "[directionsapikey]"
```

as I did for:
```
"Geocoding": {
    "EnableGoogleGeocodingService": "true",
    "GoogleGeocodingApiKey": "[geocodingapikey]"
```

so I'm not too sure if we need two separate ApiKey values for different Google services that both use the same API? Or maybe on the live server, there will be a different ApiKey per Google service?

Anyhow, figured I'd just bring it up.

Moving on to testing TwitterService

mgmccarthy:
also, you can blow away SmtpEmailSender. I wrote that and added it to the project before I truly understood how Azure worked.

mgmccarthy:
@stevejgordon, I have all my Twitter values set in my user secrets file that correspond to the Twitter values in config.json:

"Twitter": {
      "ConsumerKey": "[twitterconsumerkey]",
      "ConsumerSecret": "[twitterconsumersecret]",
      "OAuthToken": "[twitteroauthtoken]",
      "OAuthSecret": "[twitteroauthsecret]"
    },
and the code is not throwing any errors, but I'm getting a null back for name and email. I know we already discussed email, and we're giving up on it b/c Twitter makes it impossible, but in your testing, could you at least retrieve the name of the user?

stevejgordon:
@mgmccarthy - Thanks for testing it. I'm very pleased to hear that it loaded and ran! I spent some time early this morning updating with some of your points. I found a bug in the twitter code and fixed that up too. I'll push at lunch and see if you can check that again.

I had the same thought on the API key for Google so I'll consolidate that as part of this work.

stevejgordon:
@mgmccarthy These changes should resolve the twitter issue and tidy a few things up. I'll get some tests in around the geocode stuff next.

mgmccarthy:
@stevejgordon, the changes look good. Twitter is now pulling first/last name properly and the Google API key stuff looks to be consolidated.

@tonysurma :shipit:

tonysurma:
Before we merge in do we have updated/clear guidance for what one has to do on their dev machine?  Also I want to try to get it working in a web app on azure as a test first but more worried about the guidance part.

Thank you, this is excellent!

@dpaquette  @MisterJames can you run this and add thumbs up so we make sure it works for as many folks as possible before we move the system over.

Again thanks @stevejgordon and @mgmccarthy for this.  Very excited

stevejgordon:
@tonysurma - No problem. More too it than I thought but ultimately got there in the end.

Local Dev Setup:

- Visit https://www.microsoft.com/net/download/core#/current and download the latest SDK - Currently 1.1
- Install the SDK
- Open the AllReady Solution
- Clean the solution
- Perform a nuget package restore either through VS or if you prefer using dotnet restore
- Build and run the application

Possible issues:
In some cases things can get a bit messed up in VS and not cleaned. If you have any problems, try a fresh clean and the navigate to the project folder and delete the obj and bin folders manually. Then nuget restore and rebuild the solution.

stevejgordon:
@tonysurma - I have some tests I'd like to add, but those are not critical and the sooner we merge once approved by everyone, the better. So I'll either add them to this PR if not yet merged or start a new issue to cover them.

stevejgordon:
@YannickMeeus - Did you manage to try this on your Mac?

tonysurma:
@stevejgordon - wonder if there is a way we can add this to docs/wiki and link from readme maybe?

stevejgordon:
@tonysurma - Sure, I can take a look at doing that.

stevejgordon:
@tonysurma - I've added a small page to the wiki - https://github.com/HTBox/allReady/wiki/Developer-Setup - We can link there from readme once this goes live.

We also have a more detailed guide for a full setup that @digitaldrummerj put together - https://github.com/HTBox/allReady/blob/master/docs/prerequisite_install_guide/prerequisite_install_guide.md - We could update that also to ensure it mentions the latest SDK version and URL

tonysurma:
great @stevejgordon thanks.  You can edit the readme as part of this PR and then it will go live at the same time.

stevejgordon:
@tonysurma clever!

stevejgordon:
@tonysurma Guides updated.

dpaquette:
Sorry guys, I'm really late to the party here. What is our strategy going forward? Do we always want to be on the Current release our should we stick to LTS (Long Term Stable)? Something tells me that due to the nature of the application we should avoid the fast moving Current release and stick to LTS. Unless there is a compelling reason (i.e. new feature that we really need) I don't see why we would need to be on Current.

I'm curious what the rest of the team thinks about this?

YannickMeeus:
@stevejgordon I have. I'm afraid it doesn't stand up correctly due to the following line in the sampleDataGenerator.cs class, which brings about a larger compatibility issue:

TimeZoneId = "Central Standard Time",

The reason why this value bricks the app is because the time zone Database on Mac OSX and Linux is different from the one used on Windows platforms. Linux and Mac will be using the IANA tz db. There's a Jon Skeet podcast [floating about](http://nodogmapodcast.bryanhogan.net/60-jon-skeet-part-1-noda-time/) from a couple of months ago, which explains that particular issue in great detail, including how he's trying to tackle that in NodaTime.

Unfortunately, the corefx team are aware of this but have yet to commit to a fix, and are suggesting handling the windows CLDR <-> IANA tz mapping ourselves.

The future issue for handling mapping and updates to the tz dbs can be found [here](https://github.com/dotnet/corefx/issues/11897).

stevejgordon:
@dpaquette Fair question. I don't think there's a specific requirement for it. It's supposed to be quicker than 1.0.x and I thought that the Azure App Service logging feature could be useful as we do need to consider logging going forward. Otherwise we could stick 1.0.3 but targeting .net core. Even with that we've had to update with the security release versions a few times. If we think 1.0.x is better for supporting allReady I can make that change.

@YannickMeeus Oh [TODO: insert suitable cross-platform expletive here!]. That's a shame. We always said dates were hard. Agree, that it doesn't stop this merging in and getting close to the end goal, but it's a shame Mac development is still an issue. I listened to Jon on that podcast and it was very interesting. In fact we had an email thread with him over other datetime issues just after that was released.

I'll do a bit of reading as well.

mgmccarthy:
@dpaquette @stevejgordon my two cents is when we finally launch this thing, we move to LTS as it will makes things easier on maintenance going forward.

mgmccarthy:
@YannickMeeus

cc @dpaquette @stevejgordon 

the plan is to move to NodatTime, but until they have a stable release for .NET Core, we're pretty much stuck with what we have.

That being the case, since the `SampleDataGenerator` class is not required for the functionality of the site (all it does is seed data), then this is not a show stopper for Mac users. They could easily comment out/not run this code.

If that is done, will the site run on Mac?

Thanks

tonysurma:
> That being the case, since the SampleDataGenerator class is not required for the functionality of the site (all it does is seed data), then this is not a show stopper for Mac users. They could easily comment out/not run this code.

However having seeded data is helpful for testing/using the site as you develop.  Is this a case where we could have some way of knowing we are on non-windows and run a different line (like a compiler directive),  or get a 'default' systemtimezone and just use that as we don't need to compare or jump between mac/windows with this seed data, and/or have an alternate line of code for Mac users OR is it just a pure bug and nothing we can do than comment it out?

Thanks all for testing this

stevejgordon:
@mgmccarthy @YannickMeeus - Not had chance to look yet but I thought we used TimeZoneIds for creation of campaigns etc as well, selected from the dropdown. So might be a wider issue?

stevejgordon:
So we think we should move back to 1.0.3 but targetting .net core? I have some time this evening so can update this PR if that's what we prefer. So far LTS hasn't seemed any more stable than current releases. But I guess the key thing is that when they do change they are small bug fix only level fixes, not API changes etc.

dpaquette:
@YannickMeeus thanks for pointing out the timezoneid problem. On Windows, .NET gets the list of system timezones from the windows registry. I'm not sure how that works on Mac or Linux.  

As @mgmccarthy mentioned, we will be transitioning to NodaTime and it's more standard timezone database in the near future (Once NodaTime 2.0 is out of Alpa). I will create a separate ticket to track this.

YannickMeeus:
@mgmccarthy @stevejgordon I'll have a go without test data and see how far I get with testing before it bricks again. Earliest I can get some results on this will probably be tomorrow morning (GMT).

stevejgordon:
@tonysurma I've rebased this and also fixed up some failing unit tests after the rebase. I think those should have broken when I updated the context and Request model yesterday. Not sure why they didn't seem to. Anyway; all good now :-)

tonysurma:
@stevejgordon to close this off, we discussed in slack and are set to make this move.

@dpaquette and @MisterJames any thoughts to add?

stevejgordon:
@tonysurma Reverted back to LTS. I also ran this up on Windows on VS code and add those launch files into the project.

stevejgordon:
@YannickMeeus Been thinking on this and chatting to @dpaquette - We might be okay as the only places we're hard coding a value for the time zone id is in the seed data and in tests.

If you do manage to try this, skipping the seed data we might be able to validate if the rest functions. Presumably the calls into TimeZoneInfo will return values which work for you locally. We believe if it does load, if you go to create a campaign, the time zone lister there should also give us a list of IANA values?

If you can see if any of those match a Windows TimeZone name we could perhaps just change sample data to use a commonly named timezone.

PR #1701, opened 2017-01-11
Updating model and context to enforce 1 itinerary max per request.

### LINKED ISSUE #1700: Change EF mapping/model class(es) to make one Request tied to one Itinerary and any one point in time explicilt
tied to conversations in #1675

stevejgordon:
- Added a unique index on the request id for the ItineraryRequests table to prevent a request being linked there twice.
- Updated the Request to have a single Itinerary navigation rather than a collection.
- Updated the mapping for the context creation
- Added a migration - This tested locally for me without an issue. It should only fail in an unlikely case that a request is listed more than once in the ItineraryRequest table. This shouldn't be possible through the UI so is unlikely. I've checked the allReadyD database and there's no records in that table so this should deploy fine (famous last words!)

Fixes #1700

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Requests/RequestListItemsQueryHandler.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Requests/RequestListItemsQueryHandler.cs
@@ -44,7 +44,7 @@ namespace AllReady.Areas.Admin.Features.Requests
 
             if (message.Criteria.ItineraryId.HasValue)
             {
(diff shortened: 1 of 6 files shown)


mgmccarthy:
@stevejgordon, you must have eaten your Wheaties today b/c  you're knocking stuff out left and right! Let me create a couple intineraries, requets, etc... in the branch I have local now... then I'll pull this. That will be a good test of the migration (b/c as we all know, I am now super-aware of migration issues)

stevejgordon:
@mgmccarthy Haha, on a bit of a roll this week. It's the excitement for the code-a-thon next week energising me. That and not sleeping well; so I got an early start!

mgmccarthy:
@tonysurma :shipit:

PR #1703, opened 2017-01-12
Add an enable/disable setting for Twilio API calls

### LINKED ISSUE #1702: Add an enable/disable setting for Twilio API calls
Following on from #1680 we should add a setting to control whether Twilio API is enabled (or not). This will allow developers to use the fake service when they do not have Twilio credentials.

This can be similar to the setting used for the GeoCoding Google Maps API

mgmccarthy:
Fixes #1702

@stevejgordon, one note on the `FakePhoneNumberLookupService` class. There is implementation in this class that looks like it's still doing work as well as providing instruction on how to use the fake based on input to it.

IMO, we should have no implementation in the fake, it should just return IsValid is true and spit back whatever number is given to it.

If the user needs to test the conditionals surrounding the service, they should get a Twilio API key, change the config to use the real thing, and run the system that way.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Startup.cs
+++ b/AllReadyApp/Web-App/AllReady/Startup.cs
@@ -199,16 +199,16 @@ namespace AllReady
                 //services.AddTransient<IQueueStorageService, SmtpEmailSender>();
             }
 
-            if (!string.IsNullOrEmpty(Configuration["Authentication:Twilio:Sid"]) && !string.IsNullOrEmpty(Configuration["Authentication:Twilio:Token"]))
+            if (Configuration["Authentication:Twilio:EnableTwilio"] == "true")
             {
                 services.AddSingleton<IPhoneNumberLookupService, TwilioPhoneNumberLookupService>();
                 services.AddSingleton<ITwilioWrapper, TwilioWrapper>();
             }
             else
             {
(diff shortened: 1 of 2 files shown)


tonysurma:
Do I need a config setting in live site to make it enabled?

mgmccarthy:
@tonysurma set `Authentication:Twilio:EnableTwilio` to `true `

tonysurma:
@mgmccarthy @stevejgordon ok, thanks.  Going to merge.  Is there an enable needed for Google as well?  We should probably start a list of all the settings and update them as they are edited/added/whatever in code (my bad)

mgmccarthy:
@tonysurma `"EnableGoogleGeocodingService": "false"` would need to be set to `true` as well.

stevejgordon:
@mgmccarthy re the fake service. Yes I agree. However, at the time I didn't think the Twilio lookup API would work unless you paid for the carrier lookup rate (not a free lookup). I checked out the site and it seemed like I needed to pay $20 to add the credit. In the end though, it seems you do get some amount of carrier lookups included within a free trial. But I'm also not sure if that free trial lasts forever or if it's time limited. So in the end I left this code in so a dev could simulate the error cases without potentially needing a paid Twilio account. Does that seem reasonable?

PR #1714, opened 2017-01-16
Change VerifyUserToken to use same purpose as GenerateUserToken

### LINKED ISSUE #1586: Token Protection Middleware generated token is not recognized on post back to endpoint
When invoking an endpoint in our system (any controller's action method with the `[Route("api/request")]` attribute), the code invoked by the filter, `TokenProtectedResource`, is not recognizing a token that was issued from the site admin page:

image

The token by first enabling the API token functionality:
image

and then clicking the key icon to assign a token:
image

more information about how to generate this token is in PR #964.

To test this, the developer should
- create a new Controller and decorate that controller's action method with `[Route("api/request")]` and decorate the action method with `[HttpPost]` and [HttpPost]`
- log in, generate a token for one of the users (preferably, the user you will be using to test this)
- once the token has been generated, copy that information down
- using [Postman](https://www.getpostman.com/), put together an http POST to the newly created endpoint and put both the `ApiUser` and `ApiToken` key/values into the Http headers. The user should be the username for which you generated the token, and the ApiToken shoudl be the value that was shown to you when you generated a token for that user
- set a breakpoint in the `Invoke` method in `TokenProtectedResource`, then send the http post request to the endopint via Postman
- note when `VerifyUserTokenAsync` is called, it returns false for authorized. The exception it's throwing is `Microsoft.AspNetCore.Identity.UserManager: Warning: VerifyUserTokenAsync() failed with purpose: api-request-injest for user d2964736-5e52-473c-b3b5-e56ade40f227.` The userid will change based on which user you've added the token to

recumbent:
- There was a mismatch between the two tokens which meant the verification failed, changed the code in TokenProtectedResource to use TokenTypes.ApiKey and the token is then verified correctly.

Closes #1586

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Security/Middleware/TokenProtectedResource.cs
+++ b/AllReadyApp/Web-App/AllReady/Security/Middleware/TokenProtectedResource.cs
@@ -43,7 +43,7 @@ namespace AllReady.Security.Middleware
                 var token = headers.FirstOrDefault(h => h.Key == "ApiToken").Value;
 
                 var user = await manager.FindByNameAsync(apiUser);
-                var authorized = await manager.VerifyUserTokenAsync(user, "Default", "api-request-injest", token);
+                var authorized = await manager.VerifyUserTokenAsync(user, "Default", TokenTypes.ApiKey, token);
 
                 if (!authorized)
                 {


BillWagner:
Thanks @recumbent for working with us at NDC.

This passed my tests, so I'm going to :shipit:

PR #1715, opened 2017-01-16
Initial Polly wrapper for GoogleOptimizeRouteService

### LINKED ISSUE #1708: Add Polly library for HttpClient retries (OptimizeRoute)
We have discussed adding in the Polly retry library to allReady to support fluent retry logic. We will implement this in areas were we perform Http calls that could experience transient errors/delays.

We will need to review in each case, how many retries and under what delay is acceptable for the user in each given scenario. @mgmccarthy  and I have discussed and believe that the optimize route service can experience a small delay for retry as it's an expected slower operation.

I have initial code started for this.

stevejgordon:
Fixes #1708 

Adds a Polly retry around the call to the Google optimize route API endpoint. We will retry 2 times after the initial attempt on any exception. This could be improved once we have some logging to determine if there are exceptions / return codes that we feel do not warrant a retry.

- Added Polly dependency to the project
- Wrapped the Google API call code with a Retry Policy


BillWagner:
Thanks @stevejgordon This looks good.  :shipit:

PR #1720, opened 2017-01-16
Adding the beginings of some documentation for the application, keepi…

mjmilan:
…ng it simple as per BillWagner's request


BillWagner:
Thanks for the start @mjmilan Merging this now. :shipit:

PR #1721, opened 2017-01-16
Show manage key option for user appropriately

### LINKED ISSUE #1716: Visibility of "manage keys" option in Site Admin is not correct
In order to manage keys for a user that user has to have been given API access but in the UI for site admin when you click "edit" for a user that does not have API access you see the following:

image

If you click on the "Manage Keys" button you get a bad request - because the user has to have API access in order that their keys may be managed.

If you then choose to grant the user API access then the two bottom options disappear.

image

So you can't then manage the keys for a user that has API access.

I believe the intent is to show either one or the other with an else here:

https://github.com/HTBox/allReady/blob/14acb526a967daa29390993ca3e555bdc45feef9/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Site/Index.cshtml#L73-L77

recumbent:
- Only show manage keys when the user is allowed API access

Closes #1716


recumbent:
Want to reduce the changes to just the lines that are changed

PR #1685, opened 2017-01-05
Solution to Issue #1411

dcbrewster:
Fixes Issue #1411

PR #1723, opened 2017-01-16
Solution to Issue #1411 Attempt 2

naeemsarfraz:
Using pull request #1685 I've made a fix to address issue #1411 

A required attribute exists on the input elements to prevent submission of empty form

There is a bit of code which looks for errors from the API and displays appropriate error messages however it seems like no information about an invalid model is being sent back. Currently investigating the issue.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Task/Details.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Task/Details.cshtml
@@ -104,11 +104,11 @@
                     <input type="hidden" value="@Model.Id" name="taskId" id="message-taskId">
                     <div class="form-group">
                         <label for="subject" class="control-label">Subject</label>
-                        <input type="text" class="form-control wide" name="subject" id="messageVolunteersModal-subject" required>
+                        <input type="text" class="form-control wide" name="subject" id="messageVolunteersModal-subject">
                     </div>
                     <div class="form-group">
                         <label for="message-text" class="control-label">Message</label>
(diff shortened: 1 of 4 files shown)


nemesv:
@naeemsarfraz thanks for looking into this!

As far as I can tell currently the application does not use the built in HTML validation attributes but does the validation on the server side. With this change empty "Message all" forms cannot be submitted but the validation message will be inconsistent with the rest of the application.

With the `required` attribute this looks like this:

image

However with the correct server side validation it should look like this:

validation-server

I still think that the proper fix would be to change the `existingResponse` attribute in the web.config to `Auto` or `PassThrough` as I already mentioned here: https://github.com/HTBox/allReady/pull/1489#issuecomment-261490006

I still not sure about the security implication but as far as I can tell from the docs the `Auto` setting should not cause any problems.

I did some searching now and I've found these related questions on SO:

- http://stackoverflow.com/questions/15532667/asp-netazure-400-bad-request-doesnt-return-json-data/15532685#15532685
- http://stackoverflow.com/questions/31040671/what-does-existingresponse-passthrough-mean-in-iis/31041696#31041696

naeemsarfraz:
Thanks for your feedback @nemesv.

Any thoughts on @tonylambert's [concerns ](https://github.com/HTBox/allReady/pull/1489#issuecomment-261733281)around information disclosure?

BillWagner:
@stevejgordon This PR incorporates the work from PR #1685 + your suggestions.  Can you give it a quick review, and then I'll close that one and merge this one.

BillWagner:
Thanks for submitting this @dcbrewster and @naeemsarfraz 

We really appreciate it.  I'll merge this and close #1685

Issue #1506, opened 2016-11-17
Pop up window closing requires page refresh for update (connected to #1505)


schuback:
When in a campaign and signing up  to volunteer, after writing a comment and clearing the window box, it goes back to the volunteer list. It does not show that I have signed up. Even after page refresh.

tonysurma:
We need to review this workflow and have the page (post submitting to the server and closing the popup) either refresh or reditect to a page that would show the new volunteer 'result' or some other indication of success

stevejgordon:
@schuback - I'm struggling to replicate this either locally or on the test site.

I go to an event and volunteer for a task

![image](https://cloud.githubusercontent.com/assets/3669103/21619506/e25807d0-d1e7-11e6-9f5c-e137b46c238f.png)

after add comments I hit volunteer - this is the result

![image](https://cloud.githubusercontent.com/assets/3669103/21619480/c4746100-d1e7-11e6-9bbd-abf743aee200.png)

Are you able to advise is this is the same flow you followed when creating the issue? Or are the different replication steps I can try?

shahiddev:
I've just tried to repro this and get same result as @stevejgordon

mipre100:
I have been able to reproduce an issue here. When I volunteered, I got a number of System.InvalidOperationException exceptions, but I also get a slightly different screen from @stevejgordon 
<img width="713" alt="screen shot 2017-01-16 at 14 05 28" src="https://cloud.githubusercontent.com/assets/4063056/21985913/e52f1fe0-dbf4-11e6-92f3-8e2c03153c3f.png">
Example exception: [here](https://gist.github.com/mipre100/9206fd50356cfdb4b35cc8573110c684)

BillWagner:
Fixed in PR #1724

Issue #1505, opened 2016-11-17
Pop up window issue


schuback:
When in a campaign and signing up to volunteer and write a comment.  When I hit "volunteer" after writing my comments the window does not close. I have to go to the X. Would be good for it to close.

tonysurma:
clicking the action button for volunteering should close the popup

tonysurma:
Now that I see #1518 what url is this for @schuback and @OhMcGoo ?  Also how is this different from that issue? are you logged in as a different type of user?

schuback:
Here is a screen shot from a user logged in trying to sign up for a task in a campaign.  Includes the URL in the screen shot.  When I click the blue button in the window it does not close. 

<img width="1391" alt="screenshot 2016-11-18 09 06 45" src="https://cloud.githubusercontent.com/assets/830036/20432689/55d89d3a-ad6e-11e6-8dd9-800e2ae576a3.png">

Hamels:
I'll take this up during the Hackathon in Eindhoven.

Hamels:
I can't reproduce this issue in the dev environment. I can reproduce this on the  http://allready-d.azurewebsites.net/. It seems the dev and test environments are not in sync.

tonysurma:
@Hamels the site gets updated with every accepted pull request and I just verified it is up to date.  Perhaps this is due to the OS/browser combo that @schuback was using?

tonysurma:
@Hamels verified that the code on the site and in the repo are now in sync (there was a build issue in azure).  @schuback can you verify if you still have this issue live?  Thanks

tonysurma:
ping @schuback

tonylambert:
@tonysurma, @schuback  - This likely isn't helpful but I cannot reproduce this on my dev machine nor can I on http://allready-d.azurewebsites.net/

However, I do see some errors in the console (both on my dev machine and on the http://allready-d.azurewebsites.net/ site): 

![javscript console errors](https://cloud.githubusercontent.com/assets/688221/21630075/c0017cfc-d1fb-11e6-99a4-4210bfb4410c.gif)

Cleaner video if the above one isn't helpful enough: http://screencast.com/t/lGRCZptTY3C

johnmwright:
What's the status of this issue?  Is it still a valid issue for people looking to "jump in" (ie: take on first/second task)?

mk0sojo:
Tried this quickly and can't reproduce

stevejgordon:
[...]

PR #1724, opened 2017-01-16
Fixing issues when volunteering for a task:

mipre100:
#1506  Fixing a null reference that occurs when volunteering for a campaign that was created with the seed data without a campaign contact. The first ?. can result in a null, thus need the second ?.
Adding a default (empty) list of CampaignContacts to the Campaign EF model to prevent InvalidOperation occurring in an EF query when getting tasks.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Models/Campaign.cs
+++ b/AllReadyApp/Web-App/AllReady/Models/Campaign.cs
@@ -71,7 +71,7 @@ namespace AllReady.Models
 
         public Location Location { get; set; }
 
-        public List<CampaignContact> CampaignContacts { get; set; }
+        public List<CampaignContact> CampaignContacts { get; set; } = new List<CampaignContact>();
 
         public bool Locked { get; set; }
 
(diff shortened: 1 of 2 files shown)


mipre100:
Also associated with #1505

AllReadyApp/Web-App/AllReady/Models/Campaign.cs
```diff
@@ -71,7 +71,7 @@ public class Campaign
 
         public Location Location { get; set; }
 
-        public List<CampaignContact> CampaignContacts { get; set; }
+        public List<CampaignContact> CampaignContacts { get; set; } = new List<CampaignContact>();
```
BillWagner: Is this change necessary?  Looking at the code, it looks like EF will create this list on all queries, and the tests create the list for other items.

(If this fixed the issue, let me know, and I'll merge)

mipre100: This is what fixed part of the issue. There is a query that is getting the task the user is volunteering, including the campaign, and then in memory getting the campaign's first contact's email addresses. However, the in-memory component had null for the CampaignContact, thus throwing a NullReference.

BillWagner: Thanks!  I'll merge now.

Issue #641, opened 2016-03-31
finish unit tests/implement skipped tests in ManageControllerTests


mgmccarthy:
@tonysurma please label with "jump-in". Part of Issue #620

stubs are provided in:
ManageControllerTests.

mgmccarthy:
@tonysurma @BillWagner please label with "jump-in" when you get a chance

reecebradley:
Looking at it...we shall see how far I get :-)

mgmccarthy:
@reecebradley, good luck. It's one of the tougher controller to unit test.

Looking at unit tests from other controllers that use `UserManager<ApplicationUser>` and `SignInManager<ApplicationUser>` can help guide you b/c IMO, those are tough things to mock. Many of those unit tests have good setups either for the whole controller instance, or for creating mocks of `ApplicationUser` and `SignInManager` that are used in test setup.

Feel free to commit as many PR's against this issue as you need to keeping in mind smaller PR's make it easier for code reviewers.

Also, if you want to update this issue with the name of the unit tests you're working on, or plan to work on, that will keep other developers that want to contribute from working on the same tests you might be working on. I know it's not optimal, but with this many unit tests for a controller, it would be helpful.

If you have any question, feel free to ping me

reecebradley:
Thanks Michael. Presently I am working on IndexGetAddsCorrectMessageToViewDataWhenMessageEqualsChangePasswordSuccess

mgmccarthy:
@tonysurma, there is still a lot of work to do here.

Chris-Wooder:
Hi, I have have started on these and will submit a pr shortly before doing anymore.

mgmccarthy:
@Chris-Wooder, sounds good. Let me know if you have any questions.

@reecebradley, are you still working on any of these tests? I want to make sure we're not duplicating work here.

Chris-Wooder:
@mgmccarthy I am looking at the tests for the POST ResendEmailConfirmation() and have a couple of questions.

For test ResendEmailConfirmationInvokesFindByIdAsyncWithCorrectUserId()
I'm not sure which method 'InvokesFindByIdAsync' is referring to.  I am assuming it's something to do with retrieving the ApplicationUser object...

The method calls UserManager.GetUserAsync(ClaimsPrincipal), is it referring to this?

[...]

PR #1725, opened 2017-01-16
WIP - Started adding some missing tests to ManageControllerTests. 

DavidArno:
WIP for #641 

I have replaced six stubs with actual tests. I'd appreciate someone reviewing this to see if I'm heading in the right direction.

Of particular concern if the amount of duplicated code I'm creating for each test. I'd be tempted to merge a number of these tests and have multiple asserts, but that doesn't seem the project way, as far as I can see. I may have taken the one test/one assert rule too far by just fulfilling the existing test descriptions.


Chris-Wooder:
Note, to prevent duplication, I am currently working on the SetPassword tests, but haven't created a PR yet.

DavidArno:
This PR is to be superseded by a new PR that covers these tests and more and moves them to a new file

Ref: issue #1736

Issue #1699, opened 2017-01-11
Make List Of Approved Regions for API Request Config-based


mgmccarthy:
Tied to [this comment](https://github.com/HTBox/allReady/issues/1609#issuecomment-271890591) on Issue #1609

mgmccarthy:
@stevejgordon, I need to get through the code review for PR #1705, do you think you could take this task?

mgmccarthy:
most likley, it sounds like we'll need to have a hard-coded or database/table seeded list of approved regions which can be turned on via some type of config setting.

I prefer the hard-coded region list approach.... UNLESS it's something we need to add after deploying to live, unless we can redeploy the site to update the hard-coded list.

putting it in a table it a bit more flexible as we can write a sql script to populate it w/out deploying.

regardless of where it's kept, we need a way to turn it on and off via config which also makes me wonder if we want that flag to be a true config based value (which would require a redeployment or a change to the config file in live and a start/stop of the site) or if we also want to store this in the database.

stevejgordon:
@mgmccarthy Sure, I can got onto this pretty soon. Will digest the requirement and think it through a little.

mgmccarthy:
@stevejgordon definitely reach out to @tonysurma with questions on this one... he's the one that requested it be config-based in some way

stevejgordon:
@tonysurma I have a few questions after mulling this over which may guide the direction we go for this.

1. Do we know roughly how many regions we expect to manage?
2. How often will this list of regions change?

As Mike has suggested we have a few options to store the mappings. A Db table is an option. I'm not sure I like it since we're them putting in org specific database tables. I suspect this mapping requirement is very Red Cross specific in this case.

Therefore if the changes will be infrequent I'm more keen on a config setting. We could for example use a comma separated string of values on our current settings object, which can be updated in the Azure env. vars. This will likely need an app recycle to load in the latest value though.

We could also use a specific JSON file for RedCross mappings which could be updated and pushed up. This might require a customised deploy and again an app recycle. In 1.1 their is an autoReload on JSON change but we're not targeting current 1.1

We could also use a file stored in blob storage which the app could load from into a memory list. We could then have a management button to force a manual reload of the data. Or could even put in code to check this hourly using hangfire jobs perhaps.

Given the requirement is org specific I presume we want minimal engineering to implement it. So currently a single config setting of separated values might be quickest solution. If the list is long though, this gets a bit messy.

russdaygh:
[...]

PR #1726, opened 2017-01-16
List of approved regions for GASA endpoint #1699

russdaygh:
Included basic config settings for approved regions per #1699.

Included setting the appropriate acceptance argument on the GASA endpoint call if the region is (un)approved.


BillWagner:
Thanks @russdaygh for making this contribution. It looks great.  I really appreciate adding all the tests. 
:shipit:

Issue #1718, opened 2017-01-16
Editing a task with invalid data disables the Cancel button.


mjmilan:
Create an Itinerary event
Create an associated task that has a end date not on the same day as the start date.
When you try to save the task you will be given a helpful validation warning message - but if you now click the "Cancel" button, nothing happens...

We should be able to cancel out of the edit operation on the task regardless of the information on the task, valid or not.

stevejgordon:
Tested the steps and it blows with NullReferenceException inside the DetailsQueryHandler

joro550:
Do you mind me trying my hand at this? :)

shahiddev:
I can pick this up if its not being worked on - have some time on my hands at the moment so would be great to pick off some of these Jan v1 milestones ones

bergerb:
@shahiddev  I have a pull request for this bug.  Looking for feedback.  I have not worked a lot in MVC.

shahiddev:
@bergerb Haven't pulled the code and run it yet but change looks sensible to me

bergerb:
Incoming Revert?

c0g1t8:
@bergerb Oops. I did that in error.  I pulled your fix and likely clicked on something in error.  My apologies.

bergerb:
No problem.  👍

mjmilan:
I thought I had already fixed this - but nevermind - so long as it is
working now...

stevejgordon:
Checked today and this looks fixed

PR #1727, opened 2017-01-16
Enable user to cancel editing of invalid task.

mjmilan:
Populated the CancelUrl on the view model when a post is invalid, so that the user can use the cancel button to come out of the edit task page following an attempt to submit invalid data.

This is to address #1718 .

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/TaskController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/TaskController.cs
@@ -92,6 +92,7 @@ namespace AllReady.Areas.Admin.Controllers
                     RedirectToAction(nameof(Details), "Task", new { id = taskId });
             }
 
+            viewModel.CancelUrl = Url.Action(new UrlActionContext { Action = nameof(Details), Controller = "Task", Values = new { eventId = viewModel.EventId, id = viewModel.Id, area = "Admin" } });
             return View(viewModel);
         }
 


BillWagner:
@mjmilan When you have the chance, can you investigate why the automated tests are failing on this PR? Once the builds are green, I'll merge it.

mjmilan:
Will do...

naeemsarfraz:
I had a quick look at this @mjmilan, looks like the following three unit tests are failing

- EditPostAddsCorrectModelStateErrorWhenValidatorReturnsError
- EditPostInvokesValidateOnTaskSummaryModelValidatorWithCorrectParameters
- EditPostReturnsCorrectViewAndViewModelWhenValidatorReturnsError

Let me know if you're not able to look into this.

mjmilan:
I should be able to look into it at some point today...

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/TaskController.cs
```diff
@@ -92,6 +92,7 @@ public TaskController(IMediator mediator, ITaskEditViewModelValidator taskEditVi
                     RedirectToAction(nameof(Details), "Task", new { id = taskId });
             }
 
+            viewModel.CancelUrl = Url.Action(new UrlActionContext { Action = nameof(Details), Controller = "Task", Values = new { eventId = viewModel.EventId, id = viewModel.Id, area = "Admin" } });
```
naeemsarfraz: I spent a little time looking at this @mjmilan and this line is where an exception is thrown during the three failing tests I mentioned earlier.

Specifically when `Url.Action.(..)` is called an `Object reference not set to instance` exception occurs. @stevejgordon have you seen this before?

mjmilan:
I had a brief look at lunchtime... The test was hitting a certain point and
then dying...

I have my machine with me here today, so if I can swing a mid sized kitten
on the train on the way home, I'll take another look...

Otherwise, later tonight.

mjmilan:
OK - this should be ready to go...

The issue was that the Controller wasn't getting it's Url property initialized as part of the tests, so I've supplied a mock in the test to get around that.

tonysurma:
For some reason appveyor was having an issue building last night but verified today all built after the merge.  Thanks again!

PR #1728, opened 2017-01-16
Use case-sensitive path for showModal.js

ngm:
Linux is case-sensitive, so reference to showmodal.js results in a 404.

This was stopping the Volunteer modal from displaying when clicking the Volunteer button on a task in an event.


BillWagner:
Thanks @ngm I'm really happy to have this running on Linux now.
:shipit:

PR #1729, opened 2017-01-16
Use Local TimeZoneInfo in SampleDataGenerator

VishalMadhvani:
This is needed due to TimeZone Id's being different across operating systems and hence "Central Standard Time" does not exist on a mac.


BillWagner:
I like this fix @VishalMadhvani I'm glad we're moving away from more hard kinds of data.

PR #1733, opened 2017-01-17
adding text change to the required skill checkbox in the voluteer pop-up

Diana380:
applied text change as requested.


BillWagner:
@Diana380 Thank you for contributing.  This looks great.  Merging now.  :shipit:

PR #1735, opened 2017-01-17
#1707 social login error

### LINKED ISSUE #1707: Incomplete social login for Twitter results in 404
If you don't fully complete the social login registration (test with Twitter but may affect other services) it's possible to get into a position where you cannot use that provider any longer.

Steps to reproduce:

- Ensure you have user-secrets setup for Twitter with a valid key and secret
- On the login page choose to login with Twitter
- On redirect you will be taken to a page to complete your name and email etc.
- Navigate away from this page, without completing the process
- Go back to the login screen and choose Twitter again.
- You will get a 404 error from the site

The local solution is to delete the cookies for the site which removes the Identity.External cookie and allows you to retry the process.

The flow of the site may need review to see if we can prevent this situation.

shahiddev:
Fixes #1707 - Navigating away from additional information page after using social login no longer leaves the login process in invalid state requiring cookie clean up. Solution is identical to that made to the default account controller template in github (https://github.com/aspnet/Identity/issues/915#issuecomment-238939350)

Additional fixes added to handle remote failure of social logins where user is not authenticated (e.g. cancelled consent), user is now redirected back to login screen instead of crashing.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/AccountControllerTests.cs
@@ -31,12 +31,12 @@ namespace AllReady.UnitTest.Controllers
         public void LoginGetPopulatesViewDataWithTheCorrectTestUrl()
         {
             var sut = AccountController();
-
-            var result = (ViewResult)sut.Login();
+           
+            var result = (ViewResult)sut.Login().Result;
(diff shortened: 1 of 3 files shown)


mgmccarthy:
Thanks @stevejgordon for finding this bug and thanks @shahiddev for fixing this PR. I'll get eyes on this soon.

mgmccarthy:
@shahiddev SOLID PR! Thanks so much for taking the time to investigate and fix this. @tonysurma :shipit:

PR #1739, opened 2017-01-17
Update chromedriver to 2.27

recumbent:
The coded UI tests were failing when chromedriver.exe crashed.

Updating to the latest version of chromedriver (at the time of this PR) eliminates the crash and allows the tests to run. Some tests were failing in my context but I think that investigating that should be a separate is exercise


BillWagner:
Thanks for the update @recumbent I'll merge this now.  :shipit:

Issue #1736, opened 2017-01-17
Move unfinished/skipped phone number unit tests in ManageControllerTests to new class and implement those tests


DavidArno:
With reference to #641, it has been agreed with @mgmccarthy that we should split this class up into a number of separate classes, focused on email, phone number, password etc tests, respectively.

This issue relates specifically to the phone number tests:

1. Split out the phone number tests into a new `ManageControllerPhoneNumberTests` class,
2. Implement those tests.

@stevejgordon could you tag this with "NDC London", please?

stevejgordon:
@DavidArno done 😄

DavidArno:
One test remains marked as "skip" and fails if enabled. Needs a further PR to fix this.

mgmccarthy:
@DavidArno can this issue be closed seeing as PR #1740 was merged?

DavidArno:
@mgmccathy,

Unfortunately it needs to stay open for a little longer as I made a mistake and another small PR is needed to fix that. I'm aiming to create that PR tomorrow lunchtime (GMT) so ought to be ready for when you come online.

PR #1740, opened 2017-01-17
Moves all the phone number test stubs in ManageControllerTests to new class and implements them

DavidArno:
Ref issue #1736 

I have created a new `ManageControllerPhoneNumberTests` class and moved all the phone number test stubs from `ManageControllerTests` to that new class. Then all tests have been implemented.


BillWagner:
Thanks @DavidArno for all this work. It's a big help. 

This looks great.  :shipit:

PR #1741, opened 2017-01-17
#1597 - Add method of linking requests to events in the unlinked requ…

### LINKED ISSUE #1597: Organization Request List - Assign One or More Requests to an Event (Step 3)

chris-sargood:
Closes #1597 

In the 'Unlinked Requests' screen that available to Org admins, I've added the ability to link the requests to an event.

I've added a new combo box containing all events and a checkbox column into the request list as shown below:
image

The add button should have the appropriate form validation on submit - i.e. if the user hasn't selected an event and/or at least one request they should get a message telling them this. 

This screen could probably use a bit of FED love in future to make the validation a bit slicker!

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/UnlinkedRequestController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/UnlinkedRequestController.cs
@@ -1,5 +1,9 @@
-﻿using System.Threading.Tasks;
+﻿using System.Linq;
+using System.Threading.Tasks;
+using AllReady.Areas.Admin.Features.Events;
 using AllReady.Areas.Admin.Features.UnlinkedRequests;
+using AllReady.Areas.Admin.ViewModels.UnlinkedRequests;
+using AllReady.Areas.Admin.ViewModels.Validators;
 using AllReady.Security;
 using MediatR;
 using Microsoft.AspNetCore.Authorization;
@@ -12,10 +16,12 @@ namespace AllReady.Areas.Admin.Controllers
     public class UnlinkedRequestController : Controller
     {
         private readonly IMediator _mediator;
+        private readonly IUnlinkedRequestViewModelValidator _modelValidator;
 
-        public UnlinkedRequestController(IMediator mediator)
+        public UnlinkedRequestController(IMediator mediator, IUnlinkedRequestViewModelValidator modelValidator)
         {
             _mediator = mediator;
+            _modelValidator = modelValidator;
         }
 
         // GET: Admin/UnlinkedRequest/List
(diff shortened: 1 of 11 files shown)


mgmccarthy:
@stevejgordon any interest in doing the code review for this one?

chris-sargood:
@stevejgordon 
Hey Steve, 

So from what I can see in the Event model and the event table it doesn't seem that an event is associated with an Org ID directly. What I can see is that an event should always be associated with a campaign and a campaign should always be associated with a 'ManagingOrganizationId'.

It seems like a slight inefficient and long winded approach, but do you reckon it would be correct to update my query handler to build the events from the assiociated campaign's managingOrg?

stevejgordon:
@chris-sargood - I think that should be okay. There should be navigational properties between the entities so we should be able to do this in one query with an include. Do you want me to have a quick look with you tomorrow?

chris-sargood:
@stevejgordon 

Actually on second look there was nothing long winded about the changes to the query handler at all - I was just being a bone head!

I've made my changes now. I wanted to keep them in a separate commit to make them easier for you to review but --long story short-- my rebase from upstream got a bit messy and I ended up squashing everything into one commit. 

The changes I've made are in the following files
- UnlinkedRequestController.cs - you should see that the POST now does a check to ensure the passed in event ID belongs to the correct org
- UnlinkedRequestListQueryHandler.cs - theres now an addition WHERE clause on the events SELECT
- UnlinkedRequestControllerTests.cs - changed naming conventions to 'sut' and added an additional test to cover the updates to the POST method in UnlinkedRequestController
- UnlinkedRequestListQueryHandlerShould.cs - again updated the naming conventions and also tightened up my test data to reflect the changes to UnlinkedRequestListQueryHandler

Give it a look. If you reckon there's a way to make UnlinkedRequestListQueryHandler a bit tighter let me know. I'm new to this EntityFramework malarkey so I'm all ears!

mgmccarthy:
@chris-sargood new best saying on this repository: "EntityFramework malarkey"...

I couldn't' agree with you more coming from a RavenDb background myself.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/UnlinkedRequestController.cs
```diff
@@ -12,27 +16,75 @@ namespace AllReady.Areas.Admin.Controllers
...
+            if (ModelState.IsValid)
+            {
+                var eventSummary = await _mediator.SendAsync(new EventSummaryQuery()
+                {
+                    EventId = model.EventId
+                });
+
+                if (eventSummary == null || eventSummary.OrganizationId != organizationId.Value)
+                {
+                    return BadRequest("Selected event doesn't belong to the current users organization Id");
```
stevejgordon: In the case of Org id not matching I think we should return Unauthorized here.

chris-sargood:
@stevejgordon 
Hi Steve,

Updated as we discussed the other day

stevejgordon:
@chris-sargood Looks good, thanks.

PR #1742, opened 2017-01-17
Remove additional hardcoding of "Central Standard Time" from sample data

VishalMadhvani:
This should fix most of the timezone not found issues for linux and mac users when using the sample data.


dpaquette:
This looks good to me

PR #1746, opened 2017-01-18
Fixes #1736: Fixed one test that I previously missed and left "skipped"

### LINKED ISSUE #1736: Move unfinished/skipped phone number unit tests in ManageControllerTests to new class and implement those tests

DavidArno:
Fixes #1736

The previous PR missed one test - `VerifyPhoneNumberPostSendsUserByUserIdQueryWithCorrectUserId` and left it "skipped". Additionally, the test itself failed due to exercising parts of the  `VerifyPhoneNumber` method that aren't mocked.

Test now runs and logic corrected to minimise the execution path through the method under test to focus on "UserByUserIdQueryWithCorrectUserId" testing.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/ManageControllerPhoneNumberTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/ManageControllerPhoneNumberTests.cs
@@ -399,7 +399,7 @@ namespace AllReady.UnitTest.Controllers
             Assert.Equal(phoneNumber, viewModel.PhoneNumber);
         }
 
-        [Fact(Skip = "NotImplemented")]
+        [Fact]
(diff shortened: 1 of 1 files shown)


DavidArno:
@stevejgordon, @tonysurma 

Do I need to do anything else here for the PR to be merged and closed, or is it a case of Tony just having the time to do it?

It's just that I'd like to crack on with implementing more tests around this area, but would prefer not to have to start creating branches just to prevent those further changes leaking into this PR. So the sooner it can be processed, the better. 😁

MisterJames:
@DavidArno You're absolutely correct, better to work off a branch from master. All merged and good to go.

PR #1705, opened 2017-01-13
Unit tests for GoogleOptimizeRouteService

chinwobble:
Added unit tests for testing happy and sad path GoogleOptimizeRouteService.

Testing whether logging has occurred seems to be a bit difficult because all the logging methods are static extension methods. 

One of the tests also requires reading a json file to verify that the api response has been processed correctly. I'm not sure if this will work on the CI server. 

#1579

PR #1747, opened 2017-01-18
Adding optimize route status UI messaging

### LINKED ISSUE #1709: Improved UI notification for OptimizeRoute failures
Currently if the optimize route process fails (lack of connection to Google / some other error) we do not feed this information back to the end user. We simply return the requests in their original order (un-optimized but do not notify the user that no optimization occurred)

We should ensure that we include a response type that allows us to display an appropriate on screen message to the user in such cases.

This may require a simple messaging framework which will be reviewed. If so, we should split that requirement out into it's own issue.

stevejgordon:
Fixes #1709 

In order to display appropriate optimize route status messages to the end user in the UI I have refactored the service, command handler and controller. They now utilise a cache which temporarily stores optimize route statues (keyed to the user) so that on returning to the Itinerary details screen, we can render that result in the UI.

The various messages give the end user more information about any changes or failures so they can act accordingly.

image

NOTE: Please merge in after #1705 as those tests will likely need updating as part of finalising this PR. I can do that once it is merged.


mgmccarthy:
@stevejgordon do you want an extra set of eyes on this?

stevejgordon:
@mgmccarthy - Yes please, meant to tag you!

AllReadyApp/Web-App/AllReady/Caching/OptimizeRouteResultFromCacheQueryHandler.cs
```diff
@@ -0,0 +1,29 @@
+﻿using AllReady.Caching;
+using AllReady.Services.Mapping.Routing;
+using MediatR;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace AllReady.Areas.Admin.Features.Itineraries
```
stevejgordon: Will do - Good catch

stevejgordon:
@mgmccarthy - Thanks for the review. I'll look through more of these later, just heading out this morning.

AllReadyApp/Web-App/AllReady/Features/Users/GetUserIdQueryHandler.cs
```diff
@@ -6,16 +6,16 @@
 
 namespace AllReady.Features.Users
 {
-    public class GetUserIdCommandHandler : IAsyncRequestHandler<GetUserIdCommand, string>
+    public class GetUserIdQueryHandler : IAsyncRequestHandler<GetUserIdQuery, string>
```
stevejgordon: Yeah, fair point. I was originally assuming more work needed which I was trying to take away from the action and therefore keeping the controller light, but agree for the one line and dependency it saves, this handler is a bit unnecessary. Will refactor it.

AllReadyApp/Web-App/AllReady/Caching/OptimizeRouteResultFromCacheQueryHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public class OptimizeRouteResultFromCacheQueryHandler : IRequestHandler<OptimizeRouteResultFromCacheQuery, OptimizeRouteResultStatus>
```
stevejgordon: I went back on forward on this. I wasn't sure if I was misusing a handler here but I think in the end it's okay. Name wise, I'm happy to change it. I wasn't sure if we wanted to distinguish what to actually expect here. Perhaps... CachedOptimzeRouteResultQuery?

mgmccarthy: @stevejgordon, I don't think you're misusing a handler, I think you're mis-naming a handler.

Messages should be named in the language of the business. Technical implementations are a by-product of what we're trying to accomplish. When you think about it, the implementation is a really "technical artifact" of the business operation we're trying to perform.

Up to you in the end, but I'd still suggest `OptimizeRouteResultQuery`

stevejgordon:
@mgmccarthy - Updated with your PR feedback. Can you take a final glance over and if it looks good we can have this merged after #1705 is in and I've rebased this with any changes.

mgmccarthy:
@stevejgordon, one more thing. I was just looking at `OptimizeRouteCommandHandler` and it looks like if the UserId send on `OptimizeRouteCommand` is null, the handler doesn't really do anything at all except in this case:
```
else
{
    await ReOrderRequests(waypoints, requests, optimizeResult, message);
    SetOptimizeCache(message.UserId, message.ItineraryId, new OptimizeRouteResultStatus { StatusMessage = OptimizeRouteStatusMessages.OptimizeSucess });
}
```

In this case, it seems the only thing the handler is doing is reordering requests.

My question for you is, if the UserId on the message happens to be null, can we just check for that in `ItineraryController` and not dispatch the `OptimizeRouteCommand` command OR do we, in the case above, need still to re-order the requests regardless of the UserId being null?

If so, then this handler is doing work it shouldn't be doing? The the handler says it's "optimizing a route", so why in one specific case, is it reordering Requests as well (this is most likely due to my lack of knowledge as to what's involved in optimizing a route)? 

I'm thinking the whole "reorder request" concept needs to eventually become something we consolidate in code, and perhaps try to turn into something we do via a notification, b/c I swear I've seen this concept in code elsewhere. This is something we can take on later.

Anyhow, just wanted your thoughts on this. Thanks for the changes, but more importantly, thanks for figuring out what I'm sure was a fairly complex thing to figure out, which is how to optimize a route.

stevejgordon:
Hi @mgmccarthy - Yeah the code is a little complex. The handler does kinda do two things. I.e. optimize the route (calls into the service) and than re-order save results depending on the status we get back. The main reason it saves here is that we already have the request entities on our context (being change tracked) so we can update their order and save without having to re-query the DB. I agree we might be able to tidy things up a little in the future.

The handler should however always do it's work, regardless of the whether the UserId is set or not. The only difference that makes is that when it's not set (should never happen) we don't cache the result status. So in that case the UI would not show any feedback. As I say, that should never be possible since you have to be signed in to access the admin page.

mgmccarthy:
@tonysurma :shipit:

tonysurma:
once the conflicts/rebase are complete I will merge this.  Thanks!

stevejgordon:
@tonysurma Assuming it builds this is rebased. I had to update a couple of bits to account for changes made in other PRs. All tested locally. Not sure why my rebase has picked me up as committer and author so you get a wall of my face above!

tonysurma:
@stevejgordon let me know if master looks right after the merge I just did.  Thanks

PR #1752, opened 2017-01-21
`Startup.cs` Clean-Up

MisterJames:
This is a non-functional change to address the growing (and disorienting!) `startup.cs`.

 - moved configuration options objects and intialization to new configuration folder
 - broke out DI/autofac config (would prefer this to be a module, but this is a start)
 - moved security config
 - cleaned up namespaces


mgmccarthy:
YES! Thanks @MisterJames!

Issue #1501, opened 2016-11-17
Address verification process/click to show exact location


schuback:
Address field - needs to have a mapping button to confirm locations and to help pinpoint the location more.  Think like Uber and how it as you to verify your pick up location. Helps with big parking lots or buildings with multiple unmarked entries.

In the uber example it is like when it asks you to verify your location

tonysurma:
cc @stevejgordon 

Adding a bit more detail this is for where addresses are provided for events and/or tasks.  The update would be to both add lat/long to the models for those items as well as the UI to have the user enter an address (already there) and then have it display that on a map with a pin (new functionality) and optionally move/select a different pin spot to be the lat/long stored with the model.

stevejgordon:
@tonysurma We do have lat/long on some (maybe all) of the models. We have a geo-service that gets the lat/long from Google for the address. It's that lat/long we then store and use for the pins in other maps.

So I think really we just need to expose the lat/long details and as you suggest, make it possible to adjust using the map. I think this is possible with the APIs from either Google or Bing. Might need a bit of javascript expertise to make it happen dynamically.

tonysurma:
@stevejgordon is this handled by the other issues I saw you entered or is there still a gap?  THanks

stevejgordon:
@tonysurma I think there's still some work we can do here. We haven't exposed the lat/long details on the UI yet as far as I know.

dchristensen:
I can work on this, however I would like some clarification on where this feature is required.  Is this going to be added to the Event Add/Edit screen in the Admin section?  The Task screen was mentioned earlier as well, but I don't see an address associated with a Task.

@tonysurma @stevejgordon

dchristensen:
I have got the javascript working in a local branch to be able to move a point on the map and bind the lat/lon to a pair of hidden fields on a form.  I just need to know which screens need this functionality.

Right now the only place we're tracking Lat/Lon in the DB is on the request, so anything else will need to have changes made to the EF models.  @stevejgordon can you advise as to where this needs to be implemented?

stevejgordon:
@dchristensen I think the request add/edit page is most relevant to this requirement.

We also store the start and (optionally) end address lat/long for Itineraries. This is only on the edit view where the address(es) can be set. The flow is a little complex there though. Neither are required. You can optionally choose to one once location as the start and end of a itinerary route. Or you can select and enter different addresses.

We don't directly store the lat/long for the other models at the moment.

[...]

PR #1758, opened 2017-01-22
Add precise Lat/Lon selection to Request Add/Edit

dchristensen:
This PR takes care of the Request Add/Edit screen for #1501.  If everything looks OK I can continue with the Itinerary screen either as part of this PR or in a separate one.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Requests/EditRequestCommandHandlerShould.cs
@@ -99,7 +99,31 @@ namespace AllReady.UnitTest.Areas.Admin.Features.Requests
         }
 
         [Fact]
-        public async Task UpdateGeocodeWhenAddressChangedWhenUpdatingExistingRequest()
+        public async Task NotUpdateGeocodeWhenAddressChangedAndLatLonIsSetWhenUpdatingExistingRequest()
+        {
+            var mockGeocoder = new Mock<IGeocodeService>();
+            string changedAddress = "4444 Changed Address Ln";
+
+            var handler = new EditRequestCommandHandler(Context, mockGeocoder.Object);
+            await handler.Handle(new EditRequestCommand
+            {
+                RequestModel = new EditRequestViewModel
+                {
+                    Id = _existingRequest.RequestId,
+                    Address = changedAddress,
+                    City = _existingRequest.City,
+                    State = _existingRequest.State,
+                    Zip = _existingRequest.Zip,
+                    Latitude = 47.6,
+                    Longitude = -122.3
+                }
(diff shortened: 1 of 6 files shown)


dchristensen:
@stevejgordon can you take a look at this

stevejgordon:
@dchristensen Sure - I'll try to take a look later today

stevejgordon:
@dchristensen - Thanks for working on this. I've pulled it an had a look through. I'm a little unsure of the logic around the edit request. At the moment, if the request has Lat/Long and then the address is changed, the lat/long are not updated (no geocoding happens).

This makes sense in some scenarios but I can foresee occasional requirements to completely change an address. In the case of changing (rather than tweaking) an address we might want to geocoding to occur.

This is a little tricky to determine so we might need to consider something in the UI to determine if lat/long should be updated or not. We need to figure that out I think.

I also noticed that you had some js for the mapping. Some of that looks to repeat code we have in site.js. It would be nice to try and reuse/update site.js code so that any mapping code is centralised. Could you take a look at if that could work here? Obviously we need to be vary of breaking the pages using site.js currently!

dchristensen:
@stevejgordon Thanks for looking at this.  I agree this is tricky trying to figure out how to handle these changes.  

Currently, when any of the address fields change I do a geocode lookup on client side:

````javascript
      $("#Address, #City, #State, #Zip").on("change", function () {
            var address = getAddress();
            if (!isAddressEmpty(address)) {
                lookupAddress(address);
            }
        });

   function lookupAddress(address) {
        getGeoCoordinates(address.address1, address.address2, address.city, address.state, address
            .postalCode, address.country, function(l) {
                var location = setLocation(l.latitude, l.longitude);
                if (location !== null) {
                    setMapCenterAndZoom(map, [location], 15);
                }
            });
    }
````
This should update the Latitude & Longitude fields any time the address is changed, but there are no guarantees that request succeeds.  I could add code to the change handler function to set the Latitude & Longitude fields to zero before issuing the request to lookup the address, that way if the geocode fails the request will be geocoded by the back end.

As far as the mapping code, was there any specific code you saw? I have been trying to use what's in site.js when it makes sense (creating the map, center and zoom, geocoding).  Most of the mapping code is to handle manipulating the pushpin.  I could possible try to make it so it could be shared in site.js, but I don't think there is anything in there currently that would provide the control need to handle the movements.

tonysurma:
@stevejgordon any more thoughts on this PR?

stevejgordon:
@dchristensen @tonysurma - I'm not great with JS so will defer to anyone who is a bit more experienced with it. I'd suggest, if we can, we put anything around the ping that might be reusable into the site.js file.

When I tested locally the address change didn't seem to result in the lat/long updating, but I might have been unlucky. I can try and pull later and check again.

tonysurma:
@dpaquette can you take a review?

dpaquette:
Sure, I will review either tonight or tomorrow

Issue #1387, opened 2016-10-20
rename AllReadyTask to Task (low priority)


mgmccarthy:
I know `Task` is a reserved word in .NET framework, but to declare a var of `Task`, you can always use:

```
var @task
```

I use `var @event` when querying for event data.

aastha93:
I am taking this issue .

tonysurma:
Hello @aastha93  I wanted to check in after the event and see if there was a pull request to update this - knowing GitHub was down during the hackathon.  Thanks!

mgmccarthy:
just to clarify, this deals with renaming the AllReadyTask model (aka ,EF-tracked entity to Task). In this folder \Web\Model\AllReadyTask.

we can use `var @task` b/c task is a reserved keyword in .NET.

mgmccarthy:
when I tried to do this on my own, the renaming went fine, the migration went fine, I rebuilt succesfully and ran the site.

However, when looking the Index action method on the HomeController, I received this message

![image](https://cloud.githubusercontent.com/assets/2428993/21084964/0653af84-bfd9-11e6-84e5-1364b4895f26.png)

So it looks like the exception for the ambiguous reference is coming from the MVC.Rendering code in the .NET Core framework.

That being said, I'm thinking this issue should really not be worked on, unless someone wants to go in to the .NET core code and find a way to specific the `System.Threading.Tasks.Task` for whatever code implements `public Microsoft.AspNetCore.Mvc.Rendering.IHtmlHelper<IndexViewModel> Html { get; private set; }`

tonysurma:
Hey @mgmccarthy didn't this get completed during the London Ontario Hackathon via @TomWalkerCodes  ? Is the error above still an error we need to worry about?

mgmccarthy:
@tonysurma, I don't know. Either way, it's low priority. I do know that the name of a "Task" in allReady is still `AllReadyTask`.

stevejgordon:
@tonysurma I haven't seen anything come in to change the naming. I think it's a post v1 item now and if it's going to cause pain maybe leaving as it is would be for the best.

tonysurma:
did a little research and I think this is the one I am remembering https://github.com/HTBox/allReady/pull/1625

stevejgordon:
[...]

PR #1759, opened 2017-01-23
Rename task 1387

naeemsarfraz:
Changes discussed  in #1387 

Code compiles, yet to check Unit tests all still hold together.

Keen to hear feedback on the files that have been changed. And a discussion is probably needed on syncing db changes.


stevejgordon:
@naeemsarfraz After an initial scan this is looking good. Having done a few renamings myself it's a bit of thankless task and can be quite complex. I haven't pulled code to exercise it yet but I think we can do that right at the end.

We will need a migration to handle the DB changes. It's likely to be destructive and blow away the existing AllReadyTask table. Given we're not v1 and we're likely to stand up a new environment once the deployments are fixed I would suggest this is okay. Had we been live we'd have had to manage things a little differently. @tonysurma - Do you think loosing the tasks table will be an issue if we get this in soon?

Are you happy with creating an EF migration? I can assist if you need any advice.

naeemsarfraz:
@stevejgordon I'll have a stab at it and I'll give you a shout if I get stuck

There's going to be some pain for the outstanding PR's to rebase against this if we get in soon so I'll make myself available if help is needed.

stevejgordon:
@naeemsarfraz - As we're not "live" we might not need to worry about data migration here. We haven't in the past as it's just the sample data.

@tonysurma - Would you be okay with loosing any task related data from the db?

naeemsarfraz:
What do you think about renaming the following classes @stevejgordon? There are a few more examples like this.

image

stevejgordon:
@naeemsarfraz - It's probably a good idea to try and rename them to match the new domain language we're using. Easier for someone stepping into the project to find things. Some of them might get a bit long though so we might need to try and simplify where possible.

TasksByApplicationUserIdQuery for example could probably become VolunteerTaskByUserIdQuery

naeemsarfraz:
The changes I described in my last comment is still outstanding so if you could hold off merging for a day or so that would be great @tonysurma.

I'll remove WIP from the title when I'm done.

naeemsarfraz:
@stevejgordon this is ready for a final review

There are two changes I held off from doing as I just think they are a bit risky though not impossible:
1. Renaming the `Task` controller (including the `TaskAPI` controller)
2. Renaming `Features/Task` folder

tonysurma:
Thanks @naeemsarfraz as I see the WIP is removed, is this ready to go?  @stevejgordon any thoughts?

naeemsarfraz:
Yes, it's ready to go

tonysurma:
Thanks @naeemsarfraz .  @stevejgordon if you are good with latest changes and there are no open issues I will merge in

naeemsarfraz:
@stevejgordon Yes, it was as you suspected. A few tweaks were needed in javascript to properties on the model used at that point. I found the same problem with Message All Volunteers and that's been fixed.

Any other areas you suggest I check? I've had a click round a few links (Mark task as complete, unenroll) and it seems ok but I'm not entirely sure of the functionality.

I can do with your view on [this file](https://github.com/naeemsarfraz/allReady/blob/rename-task-1387/AllReadyApp/Web-App/AllReady/Views/TaskApi/Details.cshtml). There are two lines in particular ([this](https://github.com/naeemsarfraz/allReady/blob/rename-task-1387/AllReadyApp/Web-App/AllReady/Views/TaskApi/Details.cshtml#L42) and [this](https://github.com/naeemsarfraz/allReady/blob/rename-task-1387/AllReadyApp/Web-App/AllReady/Views/TaskApi/Details.cshtml#L49)) that use a tag helper but I can't find the two actions on the [TaskAPI controller](https://github.com/naeemsarfraz/allReady/blob/rename-task-1387/AllReadyApp/Web-App/AllReady/Controllers/TaskApiController.cs) they reference. Visual Studio also highlights these actions as missing. I was thinking the attribute `asp-route-taskId` might need renaming to `asp-route-volunteerTaskId`.

image


**Note: I've squashed the commits so you might have an issue if you simply pull down the changes. Might be better to blow away your copy and then fetch the branch again.**

stevejgordon:
@naeemsarfraz Sorry for the delay. Been out sick for a few days. Still not great but will try to find some time to review this again.

stevejgordon:
@naeemsarfraz - This looks good now thanks. I was able to test through the main flows manually without any issues.

In regards to that view and those referenced actions, not sure I'm afraid. I can't really see why this view exists either so it might be some legacy that's hanging around. I think for this PR we can ignore but it might warrant a little more digging and possibly we can scrap the views if not used.

@tonysurma - I think this is now good to merge.

PR #1767, opened 2017-01-24
Do not write Request to db when API Request is outside of supported regions

### LINKED ISSUE #1757: Should we still record an ApiRequest in our db when the ApiRequest is not in the list of supporeted regions?
@billwagner @OhMcGoo @tonysurma  cc @stevejgordon @MisterJames

I'm looking at the code for [this closed PR now](https://github.com/HTBox/allReady/pull/1726) which has to do with the "approved" list of regions that we service based on incoming requests from GASA (at least for v1 launch). 

Based on the implementation of this PR, it looks like we're still writing the Request to our db but then reporting back to GASA that we will not service it? Is this correct? If the code is correct, for all these Requests we're writing to our db, what is the plan to eventually be able to service them?

my understanding is if the Request is not serviceable b/c of its region, we don't write the Request to our db at all. If so, we could simply return a 400 to GASA from RequestApiController

mgmccarthy:
Fixes #1757

Ignore the name of the branch, b/c what this code does is different from what the branch name is

- does not write a Request to our db when the API Request is outside of a supported region
- removed ApiRequestProcessedNotification/Handler and instead enqueue ISendRequestStatusToGetASmokeAlarm directly in ProcessApiRequests
- added new GasaStatus static class along with public fields that are the string values of what is expected back as status to the GASA endpoint
- updated both SendInProgressStatusToGetASmokeAlarmHandler and SendRequestStatusToGetASmokeAlarmHandler to use GasaStatus insteaed of hard coding the strings.
- corrected the Settings class's namespace to match where it is in the project's diretory structure (this is acually what most of the changes are from
- fixed/updated ProcessApiRequestsShould unit tests


tonysurma:
after a rebase I will merge this. Thanks!

mgmccarthy:
@tonysurma, rebased.

Issue #1769, opened 2017-01-25
Add httpClient abstractions SendRequestStatusToGetASmokeAlarm


chinwobble:
At the moment we are missing proper abstractions `SendRequestStatusToGetASmokeAlarm` class.

mgmccarthy:
@chinwobble, the problem with using the IHttpClient abstraction on this one (and why I haven't moved to it yet) is b/c of the http headers involved in putting the auth token onto the POST request. At least that was my reasoning when I wrote it.

I just did a quick search around and it looks like TwitterService uses the IHttpClient static instance in a POST scenario. Instead of attaching the http headers to the client, it builds an HttpRequestMessage and attaches the headers to the request, then sends the request via HttpClient. In other words, HttpClient stays "header agnostic", which it should as it's a static instance, it will be used for both GET and POST scenarios and we would not want to attach headers to the client for an instance where it's being used in the wrong call.

If you'd like to have a go at changing SendRequestStatusToGetASmokeAlarm to utlizie the static IHttpClient like TwitterService does, please do and let me know how it goes.

mjmilan:
Forgive me if this is already taken care of architecturally, but that sounds like a very campaign specific method name to me.

Should we be placing such campaign specific code in the application, and if so, should we be treating it as a campaign specific add-in?

I would check, but I am using a mobile phone at the moment...

mgmccarthy:
@mjmilan  cc @stevejgordon @tonysurma 

Yes, it is campaign specific, but until we examine some real world usage, and b/c the launch of v1 is presumably going to happen this month, we had to create the first delivery accepting some Red Cross/GetASmokeAlarm-specific code/verbiage.

As we better understand usage of the system, we'll be able to start to generisize some of this functionality, but make no mistake, there is no 100% generic solution that can handle any and all requirements any and all integrations, and trying to build something like that for the first release of the software would be a mistake.

Keep in mind this is integration code (not business code), and since each integration is different, it requires different code in order to process incoming data to our model (incoming integration), or outgoing to to the integration's model (outgoing integration).

chinwobble:
Could someone please assign this to me?

mgmccarthy:
@chinwobble, assign what? I need feedback from project owners first as to when we're officially launching v1 so we don't start to refactor/change code to something more generic. We first need to figure out what that actually means.

cc @tonysurma @MisterJames

chinwobble:
I think this can be closed now that #1776 has been merged.

PR #1776, opened 2017-01-25
#1769 - added http abstraction for SendRequestStatusToGetASmokeAlarm and added tests

chinwobble:
updated SendRequestStatusToGetASmokeAlarm to use IHttpClient.

Ideally I think each web service should have its own persistent HttpClient instance, however this solution should be working for now to allow them to be tested. 

#1769

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarmShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarmShould.cs
@@ -5,6 +5,7 @@ using Moq;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Xunit;
@@ -26,13 +27,9 @@ namespace AllReady.UnitTest.Hangfire.Jobs
         }
 
         [Fact]
-        public void SendJsonService()
+        public void SendsValidHttpRequestMessage()
         {
-            HttpRequestMessage requestMsg = null;
             var mockedHttpClient = new Mock<IHttpClient>();
-            mockedHttpClient.Setup(x => x.SendAsync(It.IsAny<HttpRequestMessage>()))
-                .ReturnsAsync(new HttpResponseMessage())
-                .Callback<HttpRequestMessage>(request => requestMsg = request);
             var sendRequestStatusToGetASmokeAlarm = new SendRequestStatusToGetASmokeAlarm(_getASmokeAlarmApiSettings, mockedHttpClient.Object);
             try
             {
@@ -40,11 +37,26 @@ namespace AllReady.UnitTest.Hangfire.Jobs
             }
             catch { }
 
-            //mockedHttpClient.Verify(x => x.SendAsync(
-            //    It.Is<HttpRequestMessage>(request =>
-            //        request.RequestUri == new Uri("https://demo.getasmokealarm.org/admin/requests/status/request1")
-            //)));
+            mockedHttpClient.Verify(x => x.SendAsync(
+                It.Is<HttpRequestMessage>(request =>
+                    request.RequestUri == new Uri("https://demo.getasmokealarm.org/admin/requests/status/request1")
+                    && request.Headers.Authorization.Scheme == "someToken"
+                    && request.Content.ReadAsStringAsync().Result == "{\"acceptance\":false,\"status\":\"new\"}"
+                    && request.Content.Headers.ContentType.ToString() == "application/json; charset=utf-8"
+            )));
+        }
 
+        [Fact]
+        public void ThrowsException_WhenOnNonSuccessHttpStatus()
+        {
+            var mockedHttpClient = new Mock<IHttpClient>();
+            mockedHttpClient.Setup(x => x.SendAsync(It.IsAny<HttpRequestMessage>()))
+                .ReturnsAsync(new HttpResponseMessage(HttpStatusCode.BadRequest));
+            var sendRequestStatusToGetASmokeAlarm = new SendRequestStatusToGetASmokeAlarm(_getASmokeAlarmApiSettings, mockedHttpClient.Object);
+            
+            Assert.Throws<HttpRequestException>(
+                () => sendRequestStatusToGetASmokeAlarm.Send("request1", "new", false)
+            );            
         }
     }
 }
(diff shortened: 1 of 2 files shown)


AllReadyApp/Web-App/AllReady.UnitTest/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarmShould.cs
```diff
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
```
mgmccarthy: please delete all unused usings

AllReadyApp/Web-App/AllReady.UnitTest/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarmShould.cs
```diff
@@ -26,25 +27,36 @@ public SendRequestStatusToGetASmokeAlarmShould()
...
-        public void SendJsonService()
+        public void SendsValidHttpRequestMessage()
         {
-            HttpRequestMessage requestMsg = null;
             var mockedHttpClient = new Mock<IHttpClient>();
-            mockedHttpClient.Setup(x => x.SendAsync(It.IsAny<HttpRequestMessage>()))
-                .ReturnsAsync(new HttpResponseMessage())
-                .Callback<HttpRequestMessage>(request => requestMsg = request);
             var sendRequestStatusToGetASmokeAlarm = new SendRequestStatusToGetASmokeAlarm(_getASmokeAlarmApiSettings, mockedHttpClient.Object);
             try
```
mgmccarthy: why the try/catch here? The test you're setting up is not for the exceptional path, so you can get rid of try/catch and just write your test. Please remove try/catch.

AllReadyApp/Web-App/AllReady.UnitTest/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarmShould.cs
```diff
@@ -26,25 +27,36 @@ public SendRequestStatusToGetASmokeAlarmShould()
...
+        [Fact]
+        public void ThrowsException_WhenOnNonSuccessHttpStatus()
+        {
+            var mockedHttpClient = new Mock<IHttpClient>();
+            mockedHttpClient.Setup(x => x.SendAsync(It.IsAny<HttpRequestMessage>()))
+                .ReturnsAsync(new HttpResponseMessage(HttpStatusCode.BadRequest));
+            var sendRequestStatusToGetASmokeAlarm = new SendRequestStatusToGetASmokeAlarm(_getASmokeAlarmApiSettings, mockedHttpClient.Object);
+            
+            Assert.Throws<HttpRequestException>(
+                () => sendRequestStatusToGetASmokeAlarm.Send("request1", "new", false)
```
mgmccarthy: please replace all the args to `.Send` with `It.IsAny<T>`. We don't care what the values are, the values being something explicit makes the unit test a little unclear as to whether we're relying on those values being passed in to be part of the condition that throws the exception.

AllReadyApp/Web-App/AllReady/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarm.cs
```diff
@@ -15,14 +15,17 @@ public class SendRequestStatusToGetASmokeAlarm : ISendRequestStatusToGetASmokeAl
...
         {
             _getASmokeAlarmApiSettings = getASmokeAlarmApiSettings.Value;
+            _httpClient = httpClient;
         }
 
         public void Send(string serial, string status, bool acceptance)
         {
-            var updateRequestMessage = new { acceptance, status };
-
+            var str = JsonConvert.SerializeObject(new { acceptance, status });
```
chinwobble: Regarding NewtonSoft.Json, previously we used extension method `httpClient.PostAsJsonAsync()` which will internally call the `JsonContractorResolver` to serialise the object to a json string. So we could either add that as another abstract method to the `IHttpClient` interface or use json.net.

mgmccarthy: @chinwobble, thanks for the explanation. In the end, it's probably better to have it there as it could potentially catch bad JSON before i goes out the door to the web call, and like you said, HttpClient was already doing it. Thanks!

AllReadyApp/Web-App/AllReady.UnitTest/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarmShould.cs
chinwobble: Because the response object is not mocked, the send will throw an exception on `response.EnsureSuccessStatusCode();` . That's why try-catch is used. We could either mock a response or use try catch.

mgmccarthy: thanks. keep as-is for now.

chinwobble:
@mgmccarthy I've made all the requested changes.
I was wondering if the constructor for SendRequestStatusToGetASmokeAlarm should also check whether the settings values are null or empty and throw a more helpful exception message there than a `NullReferenceException` later on. This could potentially help with identifying misconfiguration issues much earlier on.

AllReadyApp/Web-App/AllReady/Hangfire/Jobs/SendRequestStatusToGetASmokeAlarm.cs
```diff
@@ -15,14 +15,17 @@ public class SendRequestStatusToGetASmokeAlarm : ISendRequestStatusToGetASmokeAl
...
             var request = new HttpRequestMessage(HttpMethod.Post, $"{_getASmokeAlarmApiSettings.BaseAddress}admin/requests/status/{serial}");
-            request.Content = new StringContent(JsonConvert.SerializeObject(updateRequestMessage), Encoding.UTF8, "application/json");
+            request.Content = new StringContent(
+                JsonConvert.SerializeObject(new { acceptance, status }), 
+                Encoding.UTF8, 
+                "application/json");
             request.Headers.Authorization = AuthenticationHeaderValue.Parse(_getASmokeAlarmApiSettings.Token);
```
chinwobble: I was thinking its more difficult to read when you have parameters with nested functions all in one line.

mgmccarthy:
@chinwobble, we don't need those checks. Either way, we'll get an exception. Guard clauses on constructors are a bit more problem than their worth. Based on the captured stacktrace of the unhandled exception, we'll be able to figure out what's going on.

mgmccarthy:
awesome effort here @chinwobble! Thanks for finding this and fixing it.

@tonysurma @MisterJames :shipit:

Issue #1779, opened 2017-01-26
Ensure Twitter users with no verified email can still login with Twitter login


shahiddev:
Follows on from conversation in #1763 
Handle case where twitter user doesn't have email address (or not verified). 
Use the twitter Id from IdentityUserLogin<string> table to identify registered user who has just logged in

shahiddev:
Happy to tackle this given its a follow on from the other issue

EmilyLuijbregts:
@shahiddev  - just wanted to follow up on this ticket. Are you working on this? Shall I assign you to this ticket? Thanks for your reply!

shahiddev:
Please assign it, I'll not start on this until we have a direction for the identity server item because the changes required here will become obsolete with that item.

PR #1780, opened 2017-01-26
Fixes #1763

### LINKED ISSUE #1763: Login via Twitter Fails
After an initial successful login via twitter if you logout and then try logging in the next time we get an application error.

AllReady.Features.Login.ApplicationUserQueryHandler+<Handle>d__2.MoveNext() in ApplicationUserQueryHandler.cs
            var normalizedUserName = message.UserName.ToUpperInvariant();
System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
System.Runtime.CompilerServices.TaskAwaiter.GetResult()
AllReady.Controllers.AccountController+<ExternalLoginCallback>d__20.MoveNext() in AccountController.cs
                var user = await _mediator.SendAsync(new ApplicationUserQuery { UserName = externalUserInformation.Email });

shahiddev:
Twitter login now works for majority of users
Removed Twitter service as once Twitter app registration is configured correctly you can retrieve the user email address directly from the claims returned by the Twitter middleware by setting the correct bool property in the options.

Any users who do not have a twitter email will currently not be able to log in (#1779 opened to address this)


mgmccarthy:
@shahiddev I just pulled this local and ran it, and it works great. Thanks for digging into this problem and finding a viable solution.

cc @stevejgordon I'm going to approve the code review.

PR #1732, opened 2017-01-17
P.M-A - Issue 1585

pmarshallandrew:
Fixes #1585 
Implemented functionality to display request status information , detailing status and planned deployment date . Added RequestStatus Controller,View, QueryHandler as well as accompanying Unit Tests both for handler and controller.

PR #1781, opened 2017-02-01
remove unneeded bit to fix appveyor builds

tonysurma:


tonysurma:
going to wait for build to complete but already see that this removes the feed for nuget causing the issue.  After this is merged rebasing on this will fix appveyor builds for the open PRs like #1732 as well

PR #1793, opened 2017-02-10
Cleanup for project.json

### LINKED ISSUE #1792: Project.json Cleanup
On reviewing the project.json changes I made for our .NET Core re-target, I have found a few areas we can further improve/cleanup.

stevejgordon:
Fixes #1792 

Minor cleanup in project.json files

### DIFF
--- a/AllReadyApp/AllReady.Core/project.json
+++ b/AllReadyApp/AllReady.Core/project.json
@@ -2,7 +2,6 @@
   "version": "1.0.0-*",
 
   "dependencies": {
-    "Microsoft.NETCore.Portable.Compatibility": "1.0.2",
     "NETStandard.Library": "1.6.0"
   },
 
--- a/AllReadyApp/Web-App/AllReady.UnitTest/project.json
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/project.json
@@ -8,6 +8,10 @@
   },
 
   "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.3"
+    },
     "Microsoft.EntityFrameworkCore.InMemory": "1.0.2",
     "Microsoft.AspNetCore.TestHost": "1.0.1",
     "dotnet-test-xunit": "2.2.0-preview2-build1029",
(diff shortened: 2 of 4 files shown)


stevejgordon:
Hmm, AppVeyor doesn't like this. Seemed to be working locally. Checking into it now

naeemsarfraz:
Hmmm, when I run the following commands I get an error

- `dotnet restore` OK
- `dotnet build` OK
- `dotnet run` Throws an error (see below)

htbox allready

stevejgordon:
@naeemsarfraz - Strange. I've just pulled it onto another machine and it seems to build and run okay for me there too. Could you please try deleting the bin and obj folder manually, the restoring and running? I've sometimes had the tooling hang onto things it shouldn't.

naeemsarfraz:
@stevejgordon can you confirm your `dotnet --version`? mine is 1.0.0-preview2-003156

Just cloned into a new directory and still getting the same result. If I rollback the recent PR it works fine.

:puzzled:

stevejgordon:
I have 1.0.0-rc3-004530 as I've been playing with VS 2017 and the current SDK in the past. But as the global.json specifies 1.0.0-preview2-003156 then it should be using that version of the SDK for me.

Looking at the error again, it's blowing inside the ConfigureAuthentication method. I can't see any direct calls to load files, but can you double check you have the version.json and config.json files in your web app directory?

Might be that I have a user secret set for some of the settings, that you don't. I'm installing VS 2017 and upgrading VS 2015 at the moment, so can't run anything up to investigate right now.

dchristensen:
I can confirm the same issue on my machine here with 1.0.0-preview2-003156.

It appears to be an issue with the Microsoft.AspNetCore.Authentication.Facebook library (removing the code accessing that library from `Authentication.ConfigureAuthentication` prevents the error).  

I think it comes down to an issues with Nuget not resolving dependencies correctly for libraries depending on NetStandardLibrary1.6.1, as you can see here:  https://github.com/aspnet/Security/issues/1046

The Facebook library depends on NetStandardLibarary1.6.1.

stevejgordon:
Thanks @dchristensen - I'll try and dig into this later.

stevejgordon:
@naeemsarfraz and @dchristensen - Replicated it and I believe located the issue...

```
"imports": [
  "portable-net45+win8+wp8+wpa81"
]
```
Seems having wp8 and wpa81 causes the issue - removing wp8 seems to solve it for me. Pushing a PR in 2 mins

PR #1798, opened 2017-02-23
#1718 - Check Model ID and redirect accordingly.  There are two use-c…

bergerb:
…ases: one when the user is entering a new task and another when the user is editting a task.  The cancel url needs to change depending.

See #1718 

This is a fix for the nullreference bug.


tonysurma:
@shahiddev @stevejgordon can you take a quick review and then I will merge if you have time?  Thanks @bergerb

bergerb:
Yes can do.  Thanks for taking a look!

Issue #1800, opened 2017-02-24
DateTimePicker formats date with the default en-US format


danielepo:
Whenever the browser is set to a locale different than 'en-US' the validation of the date may not work depending on the locale of the browser. For example using a browser with locale 'it' inverts day and month and you get like validation errors:
![image](https://cloud.githubusercontent.com/assets/6214856/23308266/e90c3a60-faaa-11e6-9098-5c61bd1edfdb.png)

PR #1801, opened 2017-02-24
Use window.navigator.language to set datetimepicker's locale

danielepo:
fix for #1800


tonysurma:
@stevejgordon can you take a quick review?  thanks @danielepo

stevejgordon:
Would @dpaquette mind getting eyes on this one? As it's JS and date related we need datetime-paquette on the case!

PR #1810, opened 2017-02-24
Sync script/css versions

BBosman:
While setting up my dev environment I noticed that some script and style sheets included in `_Layout.cshtml` differ in version depending on whether it's Staging/Production or Development.

This PR brings them back in line.

**bootstrap**; Local (Bower) was giving `3.3.7`, while the CDN was pointing at `3.0.0`. Set them both at `3.3.7` and made the Bower version fixed.

**bootstrap-toggle**; local (Bower) was set to `2.2.1`, while the CDN was pointing at `2.2.0`. Updated them both to `2.2.2` as `2.2.1` isn't on the CDN for some reason.

**tinymce**; local (Bower) was set to `4.2.6`, while the CDN gives `4.5.4`. I set Bower to `4.5.4` as well, but added a caret, as the CDN one doesn't contain a version number and will always give the latest in the `4.x` range.

**d3**; local (Bower) was set to `3.5.9`, while the CDN gives `3.5.17`. I set Bower to `3.5.17` as well, but added a caret, as the CDN one doesn't contain an explicit version number and will always give the latest in the `3.x` range.

**jquery-validation-unobtrusive**;  local (Bower) is set to `3.2.2`, while the CDN was giving a different version, for which I was unable to determine the actual version. I switched to `3.3.2` on a different CDN.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Views/Shared/_Layout.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Views/Shared/_Layout.cshtml
@@ -17,10 +17,10 @@
         <link rel="stylesheet" href="~/lib/fullcalendar/dist/fullcalendar.css" />
     </environment>
     <environment names="Staging,Production">
-        <link rel="stylesheet" href="//ajax.aspnetcdn.com/ajax/bootstrap/3.0.0/css/bootstrap.min.css"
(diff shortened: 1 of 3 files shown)


tonysurma:
thanks for the PR.  @stevejgordon @dpaquette can you give this a quick review?

PR #1832, opened 2017-03-02
Ability to view and set team leads on itineraries

### LINKED ISSUE #1813: Ability to mark an itinerary team member as Team Lead
Breakout from #1787

We need to support a single team lead within an itinerary. This feature will include 

- The database changes required to store team lead data.
- A UI update to set a team lead
- A UI update to show the currently configured team lead
- A UI update to change the team lead

I will limit the functionality as basic as possible in order to support more future features that can be worked on by multiple contributors.

stevejgordon:
Fixes #1813

Initial basic implementation to allow setting a team lead / changing a team lead and displaying the current team lead on the itinerary details page.

A team lead dropdown is included when potential team leads are available...

image

Once submitted a success message is shown and the team lead is displayed in the list

image

There are some future enhancements / known issues with the current implementation which I will address as separate issues.

1. It would be useful to make the team lead stand out more in the lister - Opened #1833
2. If you add a team member to the team, the team lead dropdown is not shown/updated without a full page reload - this is because some of the page functions via ajax/js code. We need to either refactor to update the knockout model and code or remove the ajax code for now.
3. Currently I use an optional value on the details page querystring to show the success/failure message for adding a team lead. This would be better achieved using TempData which requires Sessions to be enabled.


tonysurma:
thanks @stevejgordon hoping to get another set of eyes from @MisterJames or @mgmccarthy before we merge (or myself but time rarely permits such endeavors)

PR #1845, opened 2017-03-02
Add missing "Delete" tooltip on Admin/Campaign/Details/x

BBosman:
While browsing the site, getting a feel for it, I noticed that the Delete button was missing a tooltip due to a missing `title` attribute, while the other buttons in that row have one. This PR adds it.


tonysurma:
Going to go ahead and merge assuming the build will work 😉

PR #1846, opened 2017-03-02
Fix Aesop layout

BBosman:
The styles on the `logo-background` class are mangling the page layout.

As we only need to set a `background-color` after dropping that class I decided to use an inline style, but am happy to move it to `site.css` if that's preferred.


MisterJames:
It's a one-off on a stand-alone page, I'm okay with inline. Just don't tell anyone I said that. ;)

#pragmatic

PR #1847, opened 2017-03-02
Show fail reason when volunteering fails

BBosman:
When volunteering for an event and the request to the server fails for whatever reason there is no visual indication that something went wrong at all.

This PR adds the `statusText` from the failed request to the `validationErrors` to at least show that something went wrong.


MisterJames:
Thanks @BBosman, good addition.

Issue #284, opened 2015-11-08
Up for Discussion: TypeScript?


BillWagner:
Currently, all our client side code is written in JavaScript.

Do we want to switch to use TypeScript?

Comments?

dpaquette:
My vote is that we do move to TypeScript. TypeScript will help us to add some structure to our javascript. It will also give us IntelliSense support for both our javascript and for popular libraries (like jQuery and Knockout) using definition files from Definitely Typed. In my experience having the strong typing and IntelliSense support makes it a lot easier for new developers to jump in and be productive.

mheggeseth:
I support this, too. This will help us get nice intellisense in places that JS can't like parameters to callback functions. And because TS is a superset of JS, you can always drop down to raw JS if you need to be "clever".

martynfrank:
Hi. I have been following allReday from the side lines for some time, and this is the first OS project I would like to contribute to (so please be patient with me). I have now forked it and have it all set up locally.

Looking through the issues this caught my attention as I have recently worked on a project where we moved away from JS to TS. I would be keen in helping allReady to make the move. If everyone is happy for me to work on this I can share some implementation ideas and raise some corresponding issues.

I look forward to hearing from you.

BillWagner:
@martynfrank I would be happy to see you pick this up. This fork https://github.com/SteveStrong/allReady has a start, but the effort was dropped months ago. It may be very out of date, but it may help you.

martynfrank:
Thanks @BillWagner I will take a look through that fork. I was thinking of using the [TypeScript Definition Manager](https://github.com/typings/typings) (tsd) because it appears all the d.ts files we need are not managed on bower.

I will then move over to typescript a page at a time, keeping the structure the same where possible. Does it make sense for me to create an issue for each of the distinct area I migrate?

BillWagner:
@martynfrank I think that a series of small PRs would make this much easier to integrate. Pinging @stimms who should weigh in on the strategy as well.

stimms:
On some other projects I've slowly been transitioning to typescript. The approach I've taken is 

1. rename files to .ts
2. fix any minor compiler issues. Feel free to solve them by just casting stuff to any at this juncture `(<any>Blah).function)`
3. slowly start using actual types as you encounter them

A wholesale transition is, I think, less useful than changing stuff over as you encounter it. However the latter approach requires buy in from the team which is harder in an OSS project.

martynfrank:
[...]

PR #1851, opened 2017-03-03
Make typescript compile again

BBosman:
After reading #284 I decided to take a stab at seeing if I can get Typescript to work (again).

Although this PR is mergable, I consider it a bit of a Work in Progress. But I think the discussion about maybe switching to Typescript works better if you have some actual code to point at. :)

Things I did:
- Removed building of the `site.min.js` from the gulpfile. It isn't used anywhere in the site and it was confusing that you have a `site.js` (which is used) that doesn't correspond with `site.min.js`, but is actually just a small part of that file's contents.
- Fixed `tsconfig.json` so it doesn't give an error anymore and actually compiles/transpiles stuff.
- Added compiled/transpiled versions of the files in `wwwroot/ts`.
- Renamed `wwwroot/js/site.js` to `wwwroot/js/site.ts` as a sample. You can see that the now generated `site.js` closely matches the previous version.
- Tweaked `wwwroot/js/site.ts` a bit to sprinkle some Typescript goodness on it. (And to get rid of the many warnings it was giving.)
- Added Typings for `jquery` and `bingmaps`. Had to reference the latter one manually from `site.ts` as it delivers it `.d.ts` in a non-standard way.
- Put the `devDependencies` in `packages.json` in alphabetical order, updated `gulp-typescript`, added the `typescript` dependency and dropped the now unused `gulp-minify` dependency.
- Made a new `gulp build` task that transpiles the `.ts` files (and incorporates the old task that minifies the `site.css` file), so people who don't use Visual Studio can use it as well.
- Aliased the old `gulp min` task to `gulp build` for backwards compatibility in case people depend on the old name.

Questions I have:
- Are the files in `webroot/ts` used? I can't find any reference to them and they were missing their `.js` counterparts
- Do we want bundling and minification for the resulting `.js` files?
- What settings do we want to use in `tsconfig.json`?
- Do we want the generated `.js` files in sourcecontrol?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,10 @@
 *.min.js
 *.min.css
 
+#ignore map files
+*.js.map
+*.css.map
+
 #Ignore any leftover PrepOpsApp files
 PrepOpsApp
 
--- a/AllReadyApp/Web-App/AllReady/gulpfile.js
+++ b/AllReadyApp/Web-App/AllReady/gulpfile.js
@@ -3,59 +3,44 @@ var gulp = require("gulp"),
     rimraf = require("rimraf"),
     concat = require("gulp-concat"),
     cssmin = require("gulp-cssmin"),
-    uglify = require("gulp-uglify"),
     ts = require('gulp-typescript'),
     project = require("./project.json");
+
+// https://www.npmjs.com/package/gulp-typescript
+// http://weblogs.asp.net/dwahlin/creating-a-typescript-workflow-with-gulp
+var tsProject = ts.createProject('tsconfig.json');
 
 var paths = {
     webroot: "./" + project.webroot + "/"
 };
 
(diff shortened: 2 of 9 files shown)


stimms:
Might want to exclude the `.map` files as they are generated artifacts and will change every build.

MisterJames:
@BBosman thanks for this! As per @stimms suggestion, let's remove the `map` files from the PR and perhaps even add an exclude in the gitignore for `*.map` under any related JS trees. That should seal this one up.

BBosman:
I dropped the `*.map` files and added an entry for them in the `.gitignore` file.

tonysurma:
@MisterJames can you review after the edit per your and @stimms comment? Thanks

stimms:
:shipit:

Issue #1818, opened 2017-02-28
Itinerary Details Page - Hide request filter when no requests


stevejgordon:
It would be nice to hide the requests filter on the Itinerary Details page when there are no requests assigned.

e.g. http://localhost:48408/Admin/Itinerary/Details/3

stevejgordon:
Looks like we can close this.

PR #1854, opened 2017-03-08
#1818 Hide request filter when no requests

arst:
I've reverted my previous pull request for this one to be sure that it is not going to affect the build.


stevejgordon:
Hi @arst - Looking at the error, it likes like the build is failing in a similar way to issue #1855

This was fixed in #1856 and merged. If you can rebase from the latest master, hopefully your PR will work fine.

arst:
Tnanks @stevejgordon it works after rebase!

Issue #1860, opened 2017-03-10
TestHangfireController.cs failing to build in release mode.


c0g1t8:
Additional issue found when working on #1859.

```
1>------ Build started: Project: AllReadyApp, Configuration: Release Android ------
2>------ Build started: Project: AllReady, Configuration: Release Any CPU ------
2>  C:\Program Files\dotnet\dotnet.exe build "C:\Repos\allReady\AllReadyApp\Web-App\AllReady" --configuration Release --no-dependencies
1>  ------ Ensuring correct global installation of package from source package directory: C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\ApacheCordovaTools\packages\vs-tac
1>  ------ Name from source package.json: vs-tac
1>  ------ Version from source package.json: 1.0.42
1>  ------ Package already installed globally at correct version.
2>  Project AllReady (.NETCoreApp,Version=v1.0) will be compiled because inputs were modified
2>  Compiling AllReady for .NETCoreApp,Version=v1.0
1>  ------ Cordova tools 6.3.1 already installed.
1>  ------ Build Settings:
1>  ------ Build Settings:
1>  ------    platformConfigurationBldDir: C:\Repos\allReady\AllReadyApp\Mobile-App\bld\Android\Release
1>  ------    platformConfigurationBinDir: C:\Repos\allReady\AllReadyApp\Mobile-App\bin\Android\Release
1>  ------    buildCommand: prepare
1>  ------    platform: Android
1>  ------    cordovaPlatform: android
1>  ------    configuration: Release
1>  ------    cordovaConfiguration: Release
1>  ------    projectName: AllReadyApp
1>  ------    projectSourceDir: C:\Repos\allReady\AllReadyApp\Mobile-App
1>  ------    npmInstallDir: C:\Users\gchan\AppData\Roaming\npm
1>  ------    language: en-US
1>  ------ Platform android already exists
1>  ------ Updating plugins
1>  ------ Currently installed plugins: cordova-plugin-compat@1.0.0,phonegap-plugin-barcodescanner@6.0.1,cordova-plugin-device@1.1.5,cordova-plugin-console@1.0.6,cordova-plugin-statusbar@2.1.3,ionic-plugin-keyboard@2.2.1,cordova-plugin-whitelist@1.2.2,cordova-plugin-splashscreen@3.2.2
1>  ------ Currently installed dependent plugins:
1>  ------ Currently configured plugins:
[...]

PR #1861, opened 2017-03-10
TestHangfireController.cs failing to build in release mode. #1860

c0g1t8:
Fix for #1860.


tonysurma:
Thanks again @c0g1t8 .  ping @mgmccarthy as it is a small change but in Hangfire

mgmccarthy:
@c0g1t8 cc @tonysurma

A better fix would be to delete this file. We don't need it. I had initially set it up when working with Hangfire and forgot to delete it. It doesn't do anything.

tonysurma:
thanks @mgmccarthy , @c0g1t8 if you would like to delete the file (and your change 😄 ) then we can pull this in and get it compiling in release!

tonysurma:
@c0g1t8 is it just my eyes or does it seem like the 'change' for the file is removing all the lines vs removing the file?  cc @stevejgordon @BillWagner to double check as I may just be going git blind :)

stevejgordon:
@tonysurma I think that's just how the compare view handles the deletions. I pulled the PR locally and indeed the file is gone so this looks ok to me. :shipit:

Issue #1803, opened 2017-02-24
Event Managers (EPIC)


stevejgordon:
Initial stub issue around requirement for more granular authorisation levels for allReady after discussions with @schuback

Event Managers should be able to perform the following actions within the scope of a specific event

- [x] able to view an event
- [ ] able to create tasks - pending new auth classes
- [ ] create itineraries  - pending new auth classes
- [ ] create manual requests - pending new auth classes
- [x] edit event details
- [ ] close out requests - pending new auth classes
- [ ] send messaging to all event volunteers - pending on updated messaging flow
- [ ]  Update event lister / add a new lister page where an event manager can see links to events(s) they have access to manage - Included in #1898
- [ ]  Update menu rules so that anyone with event manager access has an admin menu with a link to the admin event list - Included in #1899

We will need to include a way or storing details of which users are event managers in our database. We will also need to include some changes within our authorisation code to ensure we allow the correct access to manage events for the user's who should have it.

An org admin will invite someone to become an event manager - see #1804

- [ ]  Include a page which will lists all authorized event users - Included in #1900
- [ ]  Add a button with a confirmation screen to allow an org admin (maybe also campaign manager) to revoke event manager access from a user - Included in #1901

stevejgordon:
Working on a PR to enable easier authorization logic to support this "role"

stevejgordon:
We have some of the auth in place now with the AuthorizableEvent code. The controllers / views can start to be hooked up to start enabling the access as defined above.

stevejgordon:
@c0g1t8 As you will be looking at some of the items in here, just let me know if you have any questions around the auth stuff. It's similar to resource based auth but using a custom set of classes to wrap up the code which we need to work out the parent layers for authorization.

c0g1t8:
@stevejgordon Planning to merge PR #1968 locally to get a better perspective.  Is this advisable?

tonysurma:
@c0g1t8 I just merged to master so you no longer need to :)

c0g1t8:
@tonysurma That makes things easier 😄 

@stevejgordon I've done a preliminary review of the latest bits.  I think I understand the pattern you are using.  I'll do a  deeper dive tomorrow.

stevejgordon:
@c0g1t8 If you look at most (not all) of the admin event controller actions you can see examples of how I've used the IAuthorizableEvent bits.

c0g1t8:
[...]

Issue #1802, opened 2017-02-24
Campaign Managers


stevejgordon:
Initial stub issue around requirement for more granular authorisation levels for allReady after discussions with @schuback

Campaign Managers should be able to perform the following actions within the scope of a specific campaign

- create events 
- create tasks 
- create itineraries 
- create manual requests 
- edit campaign details 
- close out task assignments 
- send messaging to all campaign levels

A user will be invited by an org admin to become a campaign admin

We will need to include a way or storing details of which users are campaign admins in our database. We will also need to include some changes within our authorisation code to ensure we allow the correct access to campaigns for the user's who should have it.

I will try to add some additional details / issues to this story soon.

stevejgordon:
Working on a PR to enable easier authorization logic to support this "role"

c0g1t8:
I'm going to take this one if no one minds.  I've got a lot practice with #1803.

c0g1t8:
@stevejgordon @tonysurma 

I've got the _enable campaign manager_ part done but it's dependent on #1803.  I'll create a PR once #1803 is merged.

@stevejgordon In making these changes, it feels like there should be an `IAuthorizableOrganization `interface with associated support classes for consistency purposes.  I ended applying an `[Authorize("OrgAdmin")]` attribute to some methods.  Thoughts?

stevejgordon:
> it feels like there should be an IAuthorizableOrganizationinterface with associated support classes for consistency purposes

@c0g1t8 Yep, that's my thought too. It makes sense to have a standard approach.

c0g1t8:
@stevejgordon I'll take a crack at `IAuthorizableOrganization`

PR #1862, opened 2017-03-13
Campaign Manager and Event Manager DB Schema

stevejgordon:
Initial db schema groundwork for #1802 and #1803 

This adds two many-to-many tables to support the storage of which campaigns and events users can manage.

They may need to be extended further in the future to support the invite functionality, but this will allow further work on the issues above.


tonysurma:
@MisterJames or @mgmccarthy any chance one of you could put a second set of eyes on this?  Thanks!

MisterJames:
@tonysurma is it possible to run a deployment off of this as a trial? I hit a snag with my DB on the first migration, but I don't know if it was a tooling thing. I killed the DB and reran, and everything worked fine.

tonysurma:
@MisterJames the deploy to allready-d worked and is live?  Is that what you were looking for?

PR #1868, opened 2017-03-14
Initial Classes for new Authorization Control Concept

stevejgordon:
This is a concept at this stage to solve what are now more complex authorization requirements for allReady.

Currently we have the concepts of a SiteAdmin and OrgAdmin. After reviewing v1 release requirements we now need to incorporate finer grainer access. This includes the additional of campaign and event managers. These can be volunteer users who have been invited by an org admin to manage specific campaigns or events.

As a result we need to be able to manage access to actions based on additional tables which store which events and campaigns a user can manage.

The PR attempts two things.

1. An addition of a UserAuthorizationService which is a class that holds onto User info, and when required loads additional data from the new tables to valid if a user should be able to perform actions on a given event/campaign/etc - It is initialised during the middleware pipeline with a custom middleware component which passes in the claims principle.

We can then call methods on this class from our controllers to validate if access is permitted. I would suggest we try to aggregate all authorization code into this class.

Any db actions are lazy and only performed if required to validate the user's access. 

2.  I have looked at improving the issue we have on actions around determining authorization. Where we have an event for example, with our additional "roles" we might have a user who has explicit access to manager the event, which can be compared by eventId. However, we might also have a campaign manager who should also be able to manage the event due to their access to a given Campaign Id. As we need to lookup and use the various hierarchy of Id's in the authorization, I've added the concept of an IAuthorizableEvent which can be built using an IAuthorizableEventBuilder.

This builder can load the additional Id's if we haven't already been able to pass them in. It will also cache objects to prevent unnecessary db calls. I've added this to one action for now as a demo of how it could work.

These changes are quite large and there would be work involved to implement them, but this could be the focus of many smaller issues and PRs once the framework is in place.

If this looks reasonable to pursue, I'd like to try and get some of this in before our code-a-thon.

@MisterJames @tonysurma - Looking for your thoughts on the code at this stage and any alternative suggestions. I played with a few thoughts while putting this together and settled on this for the time being.


stevejgordon:
Ignore any failing tests which I expect there will be!

stevejgordon:
@MisterJames I'm keeping this WIP until I have a chance to test everything again and also review the code completely, but I think this is basically where I want it now. If you do get chance to get eyes on it, let me know if you think we need any changes to the concept.

stevejgordon:
@tonysurma I think I'm now happy with this and once reviewed it can be merged in. It's mostly new classes to apply a new concept for the authorization. I've implemented in on one action for testing and as an example of how it simplifies the code.

Once this is merged I can move more code over to this pattern. It might need some tweaking as I roll it out but hopefully the core is now solid.

It will then start to allow the support for Event Managers within the system.

stevejgordon:
Had a few extra ideas so moving back to WIP

stevejgordon:
I think I now have this in a good state again. My most recent change of thinking was that it made sense to give the authorizable event more specific methods around the functionality available. This allows us to create differing rules as required. This all remains the responsibility of this object so that we have one place controlling what different user types can do to those manageable objects.

cc @mgmccarthy and @MisterJames for review and thoughts.

If it looks okay as a starting point I think we can merge and then it use can be expanded.

c0g1t8:
For those who are new like me, you may find _Resource Based Authorization #982_ a good read.  It's a good background and rationale for the intention behind this PR. 

@stevejgordon Looking forward to when this PR is approved and merged. 😄

tonysurma:
@stevejgordon looks like it needs a rebase

@mgmccarthy and @MisterJames any chance you can have a review of this soon?  Thanks!

stevejgordon:
@tonysurma rebased for reveiew

stevejgordon:
Thanks for the review @MisterJames - I suspect this will evolve and change as we move through. Once merged I can start on expanding where needed to support some of the new event and campaign managers.

PR #1911, opened 2017-03-18
Fixes #1835

### LINKED ISSUE #1835: Itinerary Details Page - Add Team Member AJAX
If you add a team member to the team for an itinerary, the team lead dropdown is not shown/updated without a full page reload - this is because some of the page functions via ajax/js code. We need to either refactor to update the knockout model so we can show/hide the team lead section as part of the ajax result  or remove the ajax code on this page.

c0g1t8:
Itinerary Details Page - Add Team Member AJAX #1835
- removed AJAX and knockout used in ItineraryDetails view
- cleaned up view


c0g1t8:
Oops - forgot to update unit tests. Coming soon.

c0g1t8:
@stevejgordon Wondering if I should add a **AddTeamMemberResult** to be consistent.

stevejgordon:
@c0g1t8 Let me pull the code and have a look. It'd be nice to be consistent, but I think we might change the way we do the notifications/messaging.

stevejgordon:
Had a quick play and I think in the future, some more bootstrap messaging would be nice, but I think it will be easier with a revised notification approach. I'd say for now, this probably meets the key requirement. We can catch up with messaging a little later. Works fine for me. If you're happy with where you have this I can do a final full check over and we can look to get it merged in.

c0g1t8:
I'm good with it. 😄

PR #1934, opened 2017-03-23
Fix appveyor dotnet restore for the solution

tonysurma:


tonysurma:
closing as fixed in #1935 more directly/cleaner

PR #1935, opened 2017-03-23
remove new(er) sdk so we can continue to build

tonysurma:
per this document https://github.com/appveyor/ci/issues/1377#issuecomment-287644271 using the total hack of removing the new sdk before we build.  This will buy us time before we move to vs2017 which is on the horizon for the next 30 days or so


BillWagner:
closing and reopening to force AppVeyor to see this.

PR #1936, opened 2017-03-23
Fixes #1785 notes on request

### LINKED ISSUE #1785: Request - Add a "Notes" Property
Add a notes property to requests so that any special requirements or request specific information can be stored against the request.

This will need to be added to the data model, db migration and the UI for managing requests.

@schuback - We don't receive special instructions from the 3rd party (GetASmokeAlarm.org) integration. There I presume we consider this an org admin function to update a specific request's instructions if required?

shahiddev:
Addresses issue #1785 
Requests now have a non mandatory Notes field which can be set/edited when creating/updating a request.

Currently not showing the notes property in the various grids.

Contains a db migration


tonysurma:
Thanks @shahiddev I will ping @stevejgordon and @MisterJames to review when/as soon as they can.

One note, we have merged in a PR recently that included a migration and our experience is that the best 'experience' for the app is if you rebase and then recreate your migration on top of the other one.  Unfortunately we have seen experiences where non-linear history on migrations confuses EF even though they are on unrelated tables etc.

Thanks again

shahiddev:
@tonysurma Thanks, I'll re-do the migration, I had already re-based and was expecting a conflict in the migration file which would have told me I needed to recreate the migration. But as you say it might be better to avoid any issues so I'll rebase and recreate the migration.

shahiddev:
Re- rebased and migrated :)

tonysurma:
@stevejgordon can you give this a quick review so we can merge it in?  thanks!

PR #1945, opened 2017-04-01
VS2017 and csproj migration

stevejgordon:
I have migrated the web solution to VS 2017 and updated as required to get this running locally.

One gotcha was many, many failures when trying to build it the first time. >200

Seems that the easiest way to solve this is to complete clear the old nuget cache at USERPROFILE%/.nuget

Close VS 2017 first and afterwards run a clean and build.

I've upgraded a few package versions as needed. There's still some places we could take newer dependencies but I have left those that I can until a future PR.

I've made the changes on the latest master branch, that is why this includes many commits from other contributors since the vs2017 branch on GitHub was made.

@tonysurma - Could you give this a try locally
@MisterJames - Perhaps you could try it also to make sure this works for more than just me :-)

If it looks good I can make a fresh PR against the master if we want to bring it in there.


tonysurma:
I will test later today or tomorrow AM

tonysurma:
@stevejgordon quick dumb question for me and future generations, why does this PR contain a bunch of commits from other PRs etc.?  I was trying to look at which files had to change and it makes it look like a bunch of unrelated files had to change?

Testing this later today

stevejgordon:
@tonysurma Not dumb at all. It's because the vs2017 branch we have here is quite a bit behind the main master branch. However, to ensure I had the most up to date code possible I used the current master when starting my work. Since I've made the PR against the older vs2017 it's showing a bunch of prior commits as changed.

I can make the PR against the main master if that's easier for review?

Not sure if there's a way to catch vs2017 back up to date with master or recreate and I can PR again after that.

tonysurma:
@stevejgordon I totally forgot we made that branch, since it appears we can make this work in one pr let's do it off of master.  My local build worked, I just want to verify a deploy can work.  Figure we need to give everyone a bit of notice as well.

stevejgordon:
@tonysurma I have opened #1950 against master so will close this against the branch

PR #1949, opened 2017-04-03
Removed empty space around hero image

### LINKED ISSUE #1869: shrink hero block size on homepage to allow data shown higher up [JT]

dchristensen:
Fixes #1869

I was able to remove about 125px worth of space by just removing padding around the image and text.  We could probably get a bit more by actually changing the image size if we need it.

For reference here are to top 600px before/after

Before:
Before

After:
After


stevejgordon:
@dchristensen Personally, I think we should aim to gain more space by reducing the image size a bit as you suggest. I think we could go to about 50-60% without loosing value.

@tonysurma Thoughts?

tonysurma:
Agree

dchristensen:
I've shrunk the hero image by 50%.

Reduced by 50%

dchristensen:
I've also been thinking, if we're trying to get more visibility to the events, there is a lot of empty space on the right side of the page.  Could we fit two or three events per row, rather than spanning them the full width?  As long as the descriptions aren't too long I think this could help bring more events into view.

Thoughts @stevejgordon @tonysurma ?

stevejgordon:
@dchristensen Looks like a good idea to me if we can preserve the logo scaling.

Personally for my eye I'd suggest a little more padding around the allReady logo (top and bottom) now we have it smaller. But I'm not a designer and that's just an opinion.

dchristensen:
@stevejgordon do you mean preserve the event logo scaling?  I think we should be able to do that, I'll give it a shot.

I think I agree with you on adding a bit of padding back in around the hero image.

tonysurma:
agree @stevejgordon and @dchristensen with a bit more vertical padding this will look great

dchristensen:
Ok, so I implemented the 2 column layout and added a bit of padding back around the hero image. All of the images except the hero are responsive now, so they scale nicely and eventually disappear as the viewport gets smaller.  Here is the current view:

Homepage

Issue #1905, opened 2017-03-16
No "friendly" 404 page for missing content


stevejgordon:
If I visit a page which does not exist, I do not get a friendly 404 page.

We used to display a nicer page and should certainly add something using the error pages flow.

e.g. http://localhost:48408/sometiing

babagy:
Hi, 

Looking into this issue at the moment. 

Thanks, Baba.

babagy:
Fixed. Pull request created. https://github.com/HTBox/allReady/pull/1916

PR #1953, opened 2017-04-05
Display Friendly Error Pages

dchristensen:
I noticed as I was working on other issues that the 404 handler was not working.  I remembered seeing an issue closed  a couple weeks ago (#1905) dealing with this, so I'm not sure why its still not working.

As part of this PR I validated that the correct pages are displayed for 404 and other errors as well.  I also centralized all of the error page handling to the ErrorController.

@stevejgordon please take a look, as I had to make an adjustment to the UserAuthorizationService in order to handle rendering the error pages via a "re-executed" request.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/ErrorControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/ErrorControllerTests.cs
@@ -1,0 +1,25 @@
+﻿using AllReady.Controllers;
+using Microsoft.AspNetCore.Mvc;
+using Xunit;
+
+namespace AllReady.UnitTest.Controllers
+{
+    public class ErrorControllerTests
+    {
+        [Fact]
+        public void ErrorReturnsTheCorrectView_WhenStatusCodeIsNull()
+        {
+            var controller = new ErrorController();
+            var result = (ViewResult) controller.Error(null);
+            Assert.Equal("~/Views/Shared/Error.cshtml", result.ViewName);
+        }
+
+        [Fact]
+        public void ErrorReturnsTheCorrectView_WhenStatusCodeIs404()
+        {
+            var controller = new ErrorController();
+            var result = (ViewResult) controller.Error(404);
+            Assert.Equal("404", result.ViewName);
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


MisterJames:
:shipit: 

@stevejgordon I think everything's fine here, but have a look at this one if you like as it relates to `UserAuthorizationService`.

stevejgordon:
@dchristensen and @MisterJames - Had one thought to avoid the direct change to the AssociateUser call that might be a bit "safer". After a quick scan of the UseStatusCodePagesWtihReExecute, it looks like they add a feature to the HttpContext features collection.

So I think we should be able to adjust the Invoke method for AssociateUserMiddleware as follows to prevent it re-running the Associate on the re-executed flow.

```
   public class AssociateUserMiddleware
    {
        private readonly RequestDelegate _next;

        public AssociateUserMiddleware(RequestDelegate next)
        {
            _next = next;
        }

        public async Task Invoke(HttpContext context, IUserAuthorizationService userAuthorizationService)
        {
            if (context.Features.Get<IStatusCodeReExecuteFeature>() == null)
            {
                await userAuthorizationService.AssociateUser(context.User);
            }

            await _next(context);
        }
    }
```
I didn't get chance to pull down the code to try this, but from the debugging I've been doing for my blogging I think this should work. Various ways to skin a cat on this one.

dchristensen:
@stevejgordon my only concern with that implementation is that it is specific to the StatusCodePages middleware.  Any other upstream middleware that causes an action to be re-executed would still throw the exception.  I would have to do some digging to verify, but I am pretty sure the `UseExceptionHandler` middleware re-executes the request, so we would likely see the problem when that middleware runs.

tonysurma:
@stevejgordon @dchristensen let me know if we are good to go from this discussion?

stevejgordon:
Sorry for the delay @dchristensen - Fair point, although I only expect us to have 1, maybe 2 that re-execute. It might be better to review us adding a feature/header as part of our user assignment middleware which we can pick up later.

@tonysurma - Let's get this in as it stands since I'd rather us have time to review if we want to go down that road. This is sound and will solve the main issue of errors which is more pressing I think.

Thanks @dchristensen for the PR

PR #1955, opened 2017-04-07
Local filesystem IImageService implementation

dchristensen:
This PR is just a quick-and-dirty implementation of the `IImageService` so that you don't have to have an Azure account setup to upload images during development.  I have found it very useful in testing the homepage, and have found some bugs because I've been able to actually render images.  The image service stores the images in an "upload" folder in `wwwroot`, so this is definitely not something that should be used in production.  I have added appropriate flags to toggle this vs the Azure Blob Storage implementation, but I believe there would need to be a config change in the Azure deployment to make sure the right service is used.

Hopefully this is something useful, if not feel free to close the PR.  Thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -238,3 +238,4 @@ bower_components
 /smoke-alarm-requests-sample-data-INVALID.csv
 /smoke-alarm-requests-sample-data-EMPTY.csv
 /smoke-alarm-requests-sample-data-DUPLICATE.csv
+upload/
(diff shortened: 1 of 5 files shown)


tonysurma:
@dchristensen - I presume it defaults to local file and I have to configure differently for the live deployment?

Two thoughts - a) can you document what the config is here in the issue and b) can we make it dependent on environment? For example, for development it uses local file but for all other environments it uses the azure based one?  That way the 'config' is automatic

Thanks again for your contributions!

tonysurma:
@dchristensen okay let's keep it as is then.  Following the pattern makes sense.  I will update production before I merge this PR then.  Thanks!

Issue #1293, opened 2016-10-04
AllReady.ScenarioTest not run as part of appveyor builds


digitaldrummerj:
Looking at the appveyor.yml file it appears that the AllReady.ScenarioTest are not actually run as part of the build.  

I can see that they were  turned off back on Feb 20th as part of this [commit](https://github.com/HTBox/allReady/commit/988d41bc6e9df5e66c2b79c01e31b17dd9d0000c).

Are they still useful or should we remove them along with the need for F#?

tonysurma:
We had plans to implement scenario tests for common scenarios so we could test them as part of builds.  Not sure why they got turned off cc @BillWagner @MisterJames

BillWagner:
These were turned off because they were taking enough time that they caused build failures.

We can look at AppVeyor config to see if our account lets us configure the failure timeout. Then, we could turn them on.

digitaldrummerj:
I would assume you have longer than 3 minutes for a process to run during the builds but maybe not.  The scenario tests run in about 3 minutes on my machine.  However, if the site changes and Canopy can't find an element on the page that it is looking for, then Canopy will throw a time out error.  It can be a little confusing to realize that the time out is an indication that it can't find an element more than it took too long to run a working test.

Also, I do expect the test running to get longer though as we add in other browsers.  Right now it is just running in Chrome but I would like to see them running in Firefox, Internet Explorer, and Edge as well.

BillWagner:
The scenario tests by themselves aren't the issue, it's the total time:
. Start the VM
. Install pre-release .NET Core SDK (which may now be a standard image. If so we, can simplify the script)
. clone the repo
. restore all nuget packages.
. do the builds
. run the unit tests
. Run the scenario tests.

All up, it was often timing out.

tonysurma:
actually in the recent past we have gotten tighter so all the overhead is < 3min 30sec, we should be good for scenario tests

PR #1962, opened 2017-04-11
Fixing Integration Tests

### LINKED ISSUE #1938: Integration Tests Broken
I tried to run the integration tests based on the [Prerequisite Install Guide](https://github.com/HTBox/allReady/blob/master/docs/prerequisite_install_guide/prerequisite_install_guide.md) and got the following:

image

I dug further and found _comment out the integration tests for now #542_, _AllReady.ScenarioTest set only to test against Chrome #1290_ and _AllReady.ScenarioTest not run as part of appveyor builds #1293_.  There also does not appear to be any _new work_ in the last few months.

What's the status on the integration tests?  Are there plans to continue using F# and Canopy for UI testing?  I've not worked with either but would be willing to take it on 😄.

dchristensen:
Fixes #1938

I spent some time wrestling with the F# integration/scenario tests.  I believe I've got them working again.  I would like someone else to try these changes and make sure its not just something working on my machine now.  I'm assuming if these end up working we could update the `appveyor.yml` to run them to close out #1293 as well.

F# is completely new to me, so if there are better ways to do any of these things I'm open to suggestions.

### DIFF
--- a/AllReadyApp/AllReady.IntegrationTest/CreatingActivitiesForCampaigns.fs
+++ b/AllReadyApp/AllReady.IntegrationTest/CreatingActivitiesForCampaigns.fs
@@ -38,19 +38,7 @@ let All baseUrl =
         TopMenu.SelectCampaigns()
         "td a" *= testCampaignName
 
-    "The Org Admin can create a campaign from public campaign page" &&& fun _ ->
-        let testCampaignName = Utils.GetScenarioTestName "Test Campaign for Org Admin From Public Page"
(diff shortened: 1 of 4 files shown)


c0g1t8:
@dchristensen I'll give your changes a try.

c0g1t8:
@dchristensen I checked out your branch, built and ran the integrations tests.  Some tests failed only when I loaded the system down with processes.  They were related to timeouts.  So it looks like you've corrected the issue.  

I'm relatively new to F# as well but your changes look good to me.

PR #1965, opened 2017-04-12
WIP - Fixes #1896 - Campaign Access Lister

### LINKED ISSUE #1896: Campaign Access List
Include a page which will lists all authorized campaign users - i.e. org admins and campaign managers

We could possibly also show the event managers in this list?

This page is accessible to org admins.

@schuback - Should campaign admins also be able to view this?

If we include different user types we should include a way to sort/filter the list

Should also include a link to the invite campaign manager page which trigger the invite of a new campaign manager. This page will be added in #1886

c0g1t8:
Hoping this addresses **Campaign Access Lister #1896**.

I added a method to the `CampaignController `to access the authorized users -**Admin/Campaign/[campaignId]/Users**.

Being relatively new to this problem domain, I made some assumptions.  Hopefully correctly 😄.
- assumed that a campaign may have more than on manager
- assumed that an event may have more than on manager

@stevejgordon Am I on the right track? If so, I can apply the same pattern to #1900

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Threading.Tasks;
 using AllReady.Areas.Admin.Features.Campaigns;
 using AllReady.Extensions;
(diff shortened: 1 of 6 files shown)


stevejgordon:
@c0g1t8 Looks like you're on the right path here and your assumptions are correct.

stevejgordon:
@c0g1t8 Is this one ready for a final review?

c0g1t8:
>@stevejgordon I've been looking into #1900 as well and can see that it has a lot in common with this issue. This makes me want to re-think what I've done in PR #1965. I'm now thinking there are two other options.

>- Create a new User controller and provide a partial view that can be called from the Campaign or Event controller.

>- Create a partial view associated with the Organization controller that can be called by the Campaign or Event Controller.

>I'm favoring the second option. I am assuming that a user/volunteer may be assigned to multiple events in multiple campaigns.

>Thoughts?

I made this comment originally in #1896 a couple of weeks ago.  😏  I'm still not comfortable enough with the domain model to decide.

PR #1966, opened 2017-04-13
Fixed Issue 1348

wmay1991:
Remove databinding from My Assigned tasks parent div


wmay1991:
will remove the solution file

Issue #1877, opened 2017-03-14
navbar - make logoff as an item 'under' your user name [JT]


tonysurma:
from a UI point of view don't take space home navbar and make it a float sub menu for logoff (can later be used for other items as well)

PR #1977, opened 2017-04-24
Fixes #1976 broken integration test

### LINKED ISSUE #1976: PR #1947 Breaks integration test
The fix for #1877 changed a tag in `_AdminNavigationPartial.cshtml` on which the integration tests requires.

c0g1t8:
Fixes #1976 

The fix for #1877 changed a tag in `_AdminNavigationPartial.cshtml` on which the integration tests requires.


c0g1t8:
Fix is now redundant.  #1978 works.

PR #1978, opened 2017-04-24
Fix admin menu in integrations test

### LINKED ISSUE #1976: PR #1947 Breaks integration test

dchristensen:
Fixes #1976

This is a pretty minor change, just adding a class back that was mistakenly removed.  The class is only used in the integration test to locate the Admin menu.


c0g1t8:
@tonysurma You can close this 😄

PR #1979, opened 2017-04-26
Fixes  #1961 Replacing Image Deletes Updated Image

### LINKED ISSUE #1961: Replacing Image Deletes Updated Image
There is an error in the logic around replacing images.  I believe it exists for all images types (organization, campaign & event).  The following steps will reproduce the error (note this must be done on a copy that is using the Azure based `IImageService` such as http://allready-d.azurewebsites.net/):

1. Upload an image
1. Verify the image is visible
1. Edit the org/campaign/event again and upload the same image again.
1. After saving the image is now broken and throws an exception if you try and update again.

This will occur with any two images that use the same filename.  The problem stems from the way we are currently naming our blobs and the fact that we always upload and then delete the old one.  If they are using the same filename, we upload the replacement and then immediately delete it.

c0g1t8:
Fixes #1961 Replacing Image Deletes Updated Image

- Fixes  image deleted when image is updated for CampaignController
- Fixes image deleted when image is updated for EventController
- Fixes wrong parameters sent to UploadEventImageAsync method

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignController.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Threading.Tasks;
 using AllReady.Areas.Admin.Features.Campaigns;
 using AllReady.Extensions;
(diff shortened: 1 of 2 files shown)


c0g1t8:
@tonysurma I've just rebased this and it's ready for review.

PR #1985, opened 2017-05-03
Added unit test to cover the SendInProgressStatusToGetASmokeAlarmHandler

### LINKED ISSUE #1765: Write unit tests for SendInProgressStatusToGetASmokeAlarmHandler

YannickMeeus:
Fixes #1765.

I settled on only one Unit Test for this. I'm not 100% convinced about it, but it seemed overkill to cover separate use-cases for a single request *and* multiple requests.

One thing I can think of that might be worth covering is a negative case, to verify that if no requests are found then nothing is queued up. One would argue we'd be testing the .NET framework, but at the very least it would verify the behaviour of this particular handler, instead of the implementation.

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Notifications/SendInProgressStatusToGetASmokeAlarmHandlerShould.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Notifications/SendInProgressStatusToGetASmokeAlarmHandlerShould.cs
@@ -1,11 +1,54 @@
-﻿using System;
+﻿using AllReady.Areas.Admin.Features.Notifications;
+using AllReady.Hangfire.Jobs;
+using AllReady.Models;
+using Hangfire;
+using Hangfire.Common;
+using Hangfire.States;
+using Moq;
+using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
+using Xunit;
 
 namespace AllReady.UnitTest.Areas.Admin.Features.Notifications
 {
-    public class SendInProgressStatusToGetASmokeAlarmHandlerShould
+
+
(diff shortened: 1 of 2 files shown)


AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Features/Notifications/SendInProgressStatusToGetASmokeAlarmHandlerShould.cs
```diff
@@ -1,11 +1,54 @@
...
+                    secondRequestId
+                }
+            };
+
+            var sut = new SendInProgressStatusToGetASmokeAlarmHandler(Context, backgroundJobClient.Object);
+            await sut.Handle(notification);
+
+            backgroundJobClient.Verify(s => s.Create(It.Is<Job>(job => (string)job.Args[0] == firstProviderrequestid && (string)job.Args[1] == GasaStatus.InProgress && (bool)job.Args[2] == expectedAcceptance), It.IsAny<EnqueuedState>()), Times.Once);
+            backgroundJobClient.Verify(s => s.Create(It.Is<Job>(job => (string)job.Args[0] == secondProviderrequestid && (string)job.Args[1] == GasaStatus.InProgress && (bool)job.Args[2] == expectedAcceptance), It.IsAny<EnqueuedState>()), Times.Once);
+        }
```
mgmccarthy: @YannickMeeus one more thing you could add here is to make sure the method we're testing is correct:
```
backgroundJobClient.Verify(x => x.Create(It.Is<Job>(job => job.Method.Name == nameof(SendRequestStatusToGetASmokeAlarm.Send) 
```

besides that, the test looks good. Thanks!

YannickMeeus: Ah cool, forgot about that check. That's added now.

PR #1986, opened 2017-05-04
Abstract azure storage / inject abstraction into ImageService and TaskAttachmentService to make them testable

mgmccarthy:
ImageService and TaskAttachmentService have a lot of repeated code, but the most important elements of those classes is there is no way to test we're building the correct paths to the azure blob storage, or in the case of ImageService, there is no way to test the exception that will be thrown when trying to upload an image extension we don't support (although, I think this should be done in the Model validation in the controller, not in the ImageService itself).

I've added an IBlobStorage/BlobStorage implementation which encapsulates the calls to the azure api which should not require testing, and now both ImageSerive and TaskAttachmentService have IBlobStorage injected.

We'll see if the IBlobStorage API/abstraction holds up against future needs for azure storage types and how AllReady will need to work with those types. For now, I consider this a step in the right direction.


tonysurma:
@mgmccarthy since I don't see WIP in the title I will assume this is good to go and merge later today, let me know if otherwise

mgmccarthy:
@tonysurma, I smoke tested all the changes against real azure blobs, and everything works, so you can merge this PR whenever you want. I didn't tag anyone for a code review, b/c I figured it's not a very interesting change.

PR #1991, opened 2017-05-07
changing strings to use nameof() #340 (partial)

johnmwright:
action names, model property names (in error messages)
and changing ResourceAdminController.Create from nameof() to a string b/c it was referencing the wrong class (a model and not the controller)


mgmccarthy:
@johnmwright did you smoke test this change?

> changing ResourceAdminController.Create from nameof() to a string b/c it was referencing the wrong class (a model and not the controller)

johnmwright:
@mgmccarthy In addition to running all existing unit and Scenario tests, I manually walked through the process of creating a new campaign resource to validate the specific code change you mentioned.

mgmccarthy:
@tonysurma :shipit:

PR #1994, opened 2017-05-07
Task Attachment Renames

### LINKED ISSUE #1939: Task Attachment Renames
In PR #1660 some inconsistencies were introduced that should be fixed:
- The interface ITaskAttachmentService should be renamed to IVolunteerTaskAttachmentService. The taskId parameter of the method UploadAsync should be volunteerTaskId.
- The class TaskAttachmentService should be renamed to VolunteerTaskAttachmentService. The taskId parameter of the method UploadAsync should be volunteerTaskId. Finally the Azure path should be `'volunteerTask/'` instead of `'task/'`
- The fields of FileAttachmentModel should be renamed: TaskId should be VolunteerTaskId and Task should be VolunteerTask. This will require a proper database migration.

mk0sojo:
Renamed TaskAttachment to VolunteerTaskAttachment for interfaces, models and service.

Closes #1939


mk0sojo:
Noticed when doing the migration that Add-Migration did fall over because AllReadyContext has a parameterless constructor. It works fine if that constructor is removed, but that breaks the mocking in a few unit tests so I left it there for now.

PR #1997, opened 2017-05-08
Campaign Managers can view list of Managed Campaigns #1894

### LINKED ISSUE #1894: Campaign Managers can view list of Managed Campaigns
Breakout from #1802 

Update campaign lister if possible or add a new lister page where a campaign manager can see links to all campaign(s) they have access to manage.

Clicking the link should take them to the admin campaign page.

HamidMosalla:
Fixes #1894


AllReadyApp/Web-App/AllReady/Views/Campaign/ManageCampaign.cshtml
```diff
@@ -0,0 +1,102 @@
+﻿@model IList<AllReady.ViewModels.Campaign.CampaignViewModel>
+@using AllReady.Security
+@using AllReady.ViewModels.Campaign
+
+@{
+    ViewData["Title"] = "All Registered Campaigns";
+    @addTagHelper "*, Microsoft.AspNetCore.Mvc.TagHelpers"
```
mgmccarthy: this should be removed b/c C:\Projects\allReady\AllReadyApp\Web-App\AllReady\Views\_ViewImports.cshtml already contains an `@addTagHelper` directive.

AllReadyApp/Web-App/AllReady/Views/Campaign/ManageCampaign.cshtml
```diff
@@ -0,0 +1,102 @@
+﻿@model IList<AllReady.ViewModels.Campaign.CampaignViewModel>
+@using AllReady.Security
+@using AllReady.ViewModels.Campaign
+
```
mgmccarthy: remove the two using lines above this comment. These are not being used on this page

AllReadyApp/Web-App/AllReady/Views/Campaign/ManageCampaign.cshtml
```diff
@@ -0,0 +1,102 @@
...
+        </div>
+    </div>
+</div>
+
+@section scripts {
+    <script>
+        var modelCampaigns = @Json.Serialize(Model);
+    </script>
+    <script src="~/js/campaigns.js"></script>
+}
```
mgmccarthy: Looking at this view as a whole, and also taking into consideration that it directly relies on C:\Projects\allReady\AllReadyApp\Web-App\AllReady\wwwroot\js\campaigns.js in order to function, why do we need to return an `AllReady.ViewModels.Campaign.CampaignViewModel` from `AuthorizedCampaignsQueryHandler` in order for this page to work?

Try this. List out all the properties you need in order for this page/ko code to run. If there are any un-used properties from the CampaignViewModel in the .js/cshtml files, then you should not re-use CampaignViewModel. Instead, create a new model, optimized for consumption by this view/ko code. Only query for what you need in `AuthorizedCampaignsQueryHandler` and only pass back a model to be used by this  view/ko code.

Reusing views is not really a good practice if you're querying for, building a model, and then passing back a model that is only using 15-20% of the properties that the view (and in this case, ko code) rely on.

This is especially true in this project b/c some of the ViewModels are so out of control. CampaignViewModel has 25 properties on it and also contains logic in some of those properties as well as logic in the constructor that takes `Model.Campaign`.

I can't see how a developer is supposed to know which constructor to use in what circumstance when constructing this ViewModel

AllReadyApp/Web-App/AllReady/Features/Campaigns/AuthorizedCampaignsQueryHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+        private readonly AllReadyContext _context;
+
+        public AuthorizedCampaignsQueryHandler(AllReadyContext context)
+        {
+            _context = context;
+        }
+
+        public async Task<List<CampaignViewModel>> Handle(AuthorizedCampaignsQuery message)
+        {
+            return await _context.CampaignManagers.Where(c => c.UserId == message.UserId)
```
mgmccarthy: see my comment about if possible, creating a new/different ViewModel to pass back from CampaignController.ManageCampaign() and only querying for the data needed to build the model here instead of constructing a huge ViewModel whose properties might not be used by C:\Projects\allReady\AllReadyApp\Web-App\AllReady\Views\Campaign\ManageCampaign.cshtml

AllReadyApp/Web-App/AllReady/Views/Campaign/Index.cshtml
```diff
@@ -43,8 +44,14 @@
                         <i class="fa" data-bind="css: { 'fa-square-o': !campaigns.showOld(), 'fa-check-square-o': campaigns.showOld }"></i>
                         Include Completed Campaigns
                     </button>
-                    <span data-bind="text: '(' + total + ')' " ></span>
+                    <span data-bind="text: '(' + total + ')' "></span>
                 </div>
+                @if (await userAuthorizationService.IsCampaignManager())
```
mgmccarthy: instead of injecting `@inject IUserAuthorizationService userAuthorizationService` at then calling:
```
@if (await userAuthorizationService.IsCampaignManager())
```

here, can you instead inject `IUserAuthorizationService` into the controller, add an `IsCampaignManager` property to the model and update the model's property instead of doing this in the view?

Call me old school, but injecting services into Views and then executing conditional statements based on those services (especially a service like this that incurs a db hit, and or cache hit) is work that should be done by the controller and/or the handler building up the ViewModel.

AllReadyApp/Web-App/AllReady.UnitTest/Controllers/CampaignControllerTests.cs
```diff
@@ -343,6 +410,26 @@ protected override string UrlEncode(string value)
...
+                      new Mock<IdentityErrorDescriber>().Object,
+                      new Mock<IServiceProvider>().Object,
+                      new Mock<ILogger<UserManager<ApplicationUser>>>().Object)
+            { }
+
+            public override Task<ApplicationUser> GetUserAsync(ClaimsPrincipal user)
+            {
+                return Task.FromResult(new ApplicationUser { Id = "123" });
+            }
         }
```
mgmccarthy: do you need to use this class? I see it copied and pasted over a couple other test classes, and that's probably not a good thing. The GetUserAsync method is not even used by this controller test class.

Check out `AllReady.UnitTest.MockHelper.CreateUserManagerMock` and see if that gives you the functionality you need to test.

If not, opening a new Issue to centralize the `FakeUserManager` classes in:
ItineraryAdminControllerTests
CampaignControllerTests
UserAuthorizationServiceTests

might be worth it.

HamidMosalla: In this case I need `FakeUserManager` and its method `GetUserAsync` to get the user and pass its Id, I should open an issue for it.

AllReadyApp/Web-App/AllReady/Views/Campaign/Index.cshtml
HamidMosalla: Campaign Index has a `List<CampaignViewModel>` as its model, what I can do is to create the property that you've suggested and set it's value for one of the items in the collection and check if any of the items has true for this property, am I correct?

HamidMosalla:
Thanks for the review, I'll open an issue for the FakeUserManager and also should check with @stevejgordon  for a more specific requirement for this page to correct the view model.

AllReadyApp/Web-App/AllReady/Views/Campaign/ManageCampaign.cshtml
HamidMosalla: I don't know if this page needs all the functionality of the Campaign Index, but when I create a new view model, with its property reduced, the ko code doesn't work, I don't know how to use ko, I just previously assumed because the view model is the same, the code will work (feeling guilty about it), I should either remove the ko code, or learn how to use it in this case, I should check this with @stevejgordon.

AllReadyApp/Web-App/AllReady.UnitTest/Controllers/CampaignControllerTests.cs
mgmccarthy: @HamidMosalla, if `CreateUserManagerMock` doesn't work or doesn't get you what you need to test it, just leave `FakeUserManager` in the test class for now. I'll do some more analysis as to why it doesn't work, and see if centralizing FakeUserManager is something we want to do or not. Thanks!

AllReadyApp/Web-App/AllReady/Views/Campaign/Index.cshtml
mgmccarthy: @HamidMosalla you are correct. Go ahead and make that change, thanks!

AllReadyApp/Web-App/AllReady/Views/Campaign/ManageCampaign.cshtml
mgmccarthy: @HamidMosalla I don't know ko very well either. I'll have to use that page and see what it looks like in order to determine 

1. do we even need to use ko on this page?
2 . if we do, can we only use ko/viewmodel code we need in order to make this page run

I'll take care of looking into that.

HamidMosalla:
@mgmccarthy I made the changes you requested, including changes the view model to only include the needed properties, it now works with ko, but both views still share the same `campaign.js` file, also the tests still uses the `UserManagerFake`, let me know if there was any change in that regard.

tonysurma:
thanks @HamidMosalla for the PR and additions

@mgmccarthy please let me know if this is good to go with the changes.  thanks all

mgmccarthy:
@tonysurma  cc @HamidMosalla 

it might take me awhile to get to this... work is "spinning" back up for me and things are getting busy.

mgmccarthy:
@HamidMosalla cc @tonysurma 

we do need knockout on the page to power the Search text box and the "Include Completed Campaigns" button:
image

Well, we don't NEED it... in theory, you could just have the page post back to the GET action method and pass optional query string params and do away with the knockout code altogehter, but knockout gives a better user experience for those two aforementioned items. So, for now, we can keep the ko code there.

As far as the view model is concerned, it looks like you changed `AuthorizedCampaignsQueryHandler` to return a List<ManageCampaignViewModel> instead of a List<CampaignViewModel> (which is the right move).

However, there are still properties on ManageCampaignViewModel that you do not need and can remove. Those properties are:
- FullDescription
- ExternalUrl
- ExternalUrlText
- ManagingOrganizationId
- TimeZoneId
- ImageUrl
- Published

in addition to removing those properties from `ManageCampaignViewModel` can you also:
- remove the "default" constructor
- get rid of the constructor that takes a Campaign
- this will break the `ToManageCampaignViewModel` extension method.  That's fine, delete that method
- instead of using an extension method to build the ViewModel out of the Campain, use a LINQ projection in AuthorizedCampaignsQueryHandler to create the `ManageCampaignViewModel` from the EF query in there. 

If you need help with how projections work, take a look at the implementation of DeleteViewModelQueryHandler.

The reason I want to make the changes is the ViewModel having two constructors is confusing. If you don't know about the extension method, then you don't really know which constructor to use when creating a new ManageCampaignViewModel, which is confusing to new devs to the project (and it was confusing to me until I finally connected the dots on my own)..

we want to reduce cognitive load and make a low barrier to entry for contributors. I think these changes do just that.

Let me know if you need any help/have any questions

Thanks,
Mike

mgmccarthy:
@HamidMosalla, I found a way around FakeUserManager. I knew we had test helpers for `UserManager<ApplicationUser>` and `SignInManager<ApplicationUser>` dependencies.

Check out the AllReady.UnitTest.MockHelper class for more.

Here's a hint for how to delete `FakeUserManager` from the CampaignControllerTests class.

for `public async Task ManageCampaign_ReturnsUnAuthorized_IfUserIsNotCampaignManager()`, pass this into the constructor where `new FakeUserManager` is:
`var sut = new CampaignController(null, authorizationServiceMock.Object, MockHelper.CreateUserManagerMock().Object);`

for `public async Task ManageCampaign_ReturnsTheCorrectViewModel()` the constructor should look like this:
```
var userManager = MockHelper.CreateUserManagerMock();
userManager.Setup(x => x.GetUserAsync(It.IsAny<ClaimsPrincipal>())).ReturnsAsync(new ApplicationUser());
var sut = new CampaignController(mockedMediator.Object, authorizationServiceMock.Object, userManager.Object);
```

for `public async Task ManageCampaign_ReturnsTheCorrectView()` I'll let you take a crack at that one. 

once theses are fixed, you can delete the FakeUserManager class. I'm not too sure if this class can be removed from other Controller tests, but that's part of the analysis on Issue #1999 I'm doing.

Any questions, please feel free to reach out 

Thanks!

HamidMosalla:
@mgmccarthy Done cc @tonysurma. Just one question out of curiosity, why should I get rid of the constructor that takes a Campaign and not use `ToManageCampaignViewModel`? Is it a bad pattern? What is the ideal way of doing this? I mean from what I can tell it makes the query a little neater, Thanks.

mgmccarthy:
@HamidMosalla 

> why should I get rid of the constructor that takes a Campaign and not use ToManageCampaignViewModel?

this is a great question to ask. It's the "why" behind decisions that carries the real reasons we choose one approach over another.

Before you read why, keep in mind this is "IMHO". There are no hard and fast rules, and there is no such thing as a "best practice".  There is only what makes sense in a given use case and the simplest way to represent that functionality in code while following tried and true programming practices like SOLID (etc...)

The purpose of a ViewModel is to provide only what the consuming view needs in order to render the page. By providing an extension method that calls the ViewModel's constructor passing in an entity, it pretty much violates that usage. Why? Because we need maybe 1/4 of the properties on Campaign in order to create and populate ManageCampaignViewModel.

That, plus the fact that now ManageCampaignViewModel knows about Campaign, which is not a good thing. We create ViewModel's as an abstraction over our entities. If we had a page that basically worked with every field on Campaign, we shouldn't really use a ViewModel there, we could just pass Campaign back and forth. In this case, we need only a subset on Campaign's fields, so a ViewModel is the way to go.

All of the work of creating the ViewModel for view consumption should be owned and encapsulated by the MediatR query handlers. This way, any changes to Campaign will be centralized there. Since ViewModels usually should not be reused, we'd only need to write the projection once in the query handler anyway, so this one class not only owns the query of how to get the data from the system we need, but it also owns the "smarts" of how to compose together a viable ViewModel for consumption by the controller that sends the query message. This way, the controllers simply pass the ViewModel from the return of the query handler to the view.

And one more thing, the ViewModel with two constructors is confusing. If I'm a new contributor, and I'm looking at this class w/out any prior system knowledge, then which constructor am I supposed to call? There is an empty constructor, then a constructor that takes a Campaign? But the fields are getter/setters, so the objects now immutable. If the object is not immutable, then why do I have a constructor that takes an entity in the first place?

Sorry for the long-winded explanation, but you asked ;)

mgmccarthy:
@HamidMosalla thanks for the PR and working through the changes. @tonysurma :shipit:

HamidMosalla:
> Sorry for the long-winded explanation, but you asked ;)
@mgmccarthy  On the contrary, I'm very grateful for your complete answers and keep some of them for future reference, Thanks!

PR #2002, opened 2017-05-14
Changed locale on the datetime picker to match current culture

mk0sojo:
Changed locale on the datetime picker to match the culture set in startup.cs

#1903


stevejgordon:
I'll defer to @dpaquette on this one. Seems reasonable but not clear on the implications.

tonysurma:
thanks @mk0sojo for the PR and @stevejgordon for the first review.  @dpaquette can you take a review on this as our resident datetime expert?  Thanks!

dpaquette:
I think this is OK. As we add support for more locales on the server side we will just need to make sure we include those same momentjs locales on the client side. For now there is nothing else we need to do because  `System.Globalization.CultureInfo.CurrentCulture` is always en-US.

:shipit:

PR #2006, opened 2017-05-19
Fixed two skipped CampaignAdminControllerTests tests marked as broken

TristanGaydon:
Fixed two skipped CampaignAdminControllerTests tests marked as broken;
DeleteConfirmedSendsDeleteCampaignCommandWithCorrectCampaignId
PublishConfirmedSendsPublishCampaignCommandWithCorrectCampaignId

### DIFF
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
@@ -738,15 +738,14 @@ namespace AllReady.UnitTest.Areas.Admin.Controllers
                 .ShouldBe("NothingToDelete");
         }
 
-        [Fact(Skip = "Broken Test")]
+         [Fact]
         public async Task DeleteConfirmedSendsDeleteCampaignCommandWithCorrectCampaignId()
         {
             var viewModel = new DeleteViewModel { Id = 1, OrganizationId = 1 };
 
             var mediator = new Mock<IMediator>();
-            mediator.Setup(a => a.SendAsync(It.IsAny<DeleteCampaignCommand>())).Verifiable();
+            mediator.Setup(x => x.SendAsync(It.IsAny<AllReady.Areas.Admin.Features.Organizations.AuthorizableOrganizationQuery>())).ReturnsAsync(new FakeAuthorizableOrganization(false, false, true, false, 1));
             var sut = new CampaignController(mediator.Object, null);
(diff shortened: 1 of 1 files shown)


TristanGaydon:
I saw these two tests were marked as broken since [16/12/2016 07:59 / SHA 95732e4e](https://github.com/HTBox/allReady/commit/95732e4eec8e3c40244283c855fb76dac6e27b45) and they seemed quick and easy to fix.

AllReadyApp/Web-App/AllReady.UnitTest/Areas/Admin/Controllers/CampaignAdminControllerTests.cs
```diff
@@ -738,15 +738,14 @@ public void EditPostHasValidateAntiForgeryTokenttribute()
...
 
-        [Fact(Skip = "Broken Test")]
+         [Fact]
         public async Task DeleteConfirmedSendsDeleteCampaignCommandWithCorrectCampaignId()
         {
             var viewModel = new DeleteViewModel { Id = 1, OrganizationId = 1 };
 
             var mediator = new Mock<IMediator>();
-            mediator.Setup(a => a.SendAsync(It.IsAny<DeleteCampaignCommand>())).Verifiable();
+            mediator.Setup(x => x.SendAsync(It.IsAny<AllReady.Areas.Admin.Features.Organizations.AuthorizableOrganizationQuery>())).ReturnsAsync(new FakeAuthorizableOrganization(false, false, true, false, 1));
```
mgmccarthy: @stevejgordon 

I'm looking at the code in this unit test class, and it seems that we're using Fake `Authorizable` classes in the return from IMediator queries. 

All of the Authorizable classes implement an interface, so I'm not too sure why we're not creating a mock of the interface, doing setup on the object, and returning the mock object from the mocked mediator call?

Instead of creating a new `FakeAuthorizableOrganization` here with the correct constructor args, we could instead create a mock of `IAuthorizableOrganization`, setup the mock object to return true for `UserCanDelete` then return that mocked object from the mocked `.SendAsync` mediator call. This:
```
mediator.Setup(x => x.SendAsync(It.IsAny<AuthorizableOrganizationQuery>()))
    .ReturnsAsync(new FakeAuthorizableOrganization(false, false, true, false, 1));
```

becomes this:
```
var authorizableOrganization = new Mock<IAuthorizableOrganization>();
authorizableOrganization.Setup(x => x.UserCanDelete()).ReturnsAsync(true);
mediator.Setup(x => x.SendAsync(It.IsAny<AuthorizableOrganizationQuery>())).ReturnsAsync(authorizableOrganization.Object);
```

we could get rid of all the Fake Authorizable classes if we took this approach, seeing how they're only being utilized for unit tests.

thoughts?

Issue #1886, opened 2017-03-15
Create ManagementInvite Page


stevejgordon:
Breakout from #1804 and follow on to #1885

We should create a controller to handle invites with an action and view to support creating / sending an invite. This should accept an email address and optional custom message

Sending the invite should utilise a handler to update the database to record the invite details.

In a future issue we will add a notification to trigger sending an email.

pmarshallandrew:
picking up

EmilyLuijbregts:
Hi @pmarshallandrew - just wanted to check in and see how you were getting on?

mk0sojo:
I can look into this one

mk0sojo:
Submitted pull request for this one. Feel free to review.

mk0sojo:
Hi @stevejgordon, one thing that came up during the review of the PR is where the user should be redirected after the invite is created.
I think this depends on where it is created from (where the "Create Invite" button is).
My thought is that it would be under ~/Admin/Event/Details/X and ~/Admin/Campaign/Details/X (X is the ID) and after the invite is created will be redirected back to this page (in the same way as "Create Resource").
What is your view?

stevejgordon:
@mk0sojo That sounds perfect to me.

mk0sojo:
Should I do the Invite section on the campaign / event admin details pages as part of this issue or is that a separate task? Don't mind doing here, just wanted to check so no one else does it somewhere else as well.

mk0sojo:
Sorry, just another question. If an invite already exists for a user and you try to send another one, should we update the record in the database or show an error?

TimBibb:
Wrong email man, sorry.

TimBibb:
But it’s funny because I’ve been looking into HTBox

mgmccarthy:
@stevejgordon @mk0sojo I'll start to look into the messaging (email) portion of this issue...

BillWagner:
closed per #2009

Issue #1885, opened 2017-03-15
Model and Schema for ManagementInvites 


stevejgordon:
Breakout from #1804

Implement a model, db schema and migration to support management invites.

Database needs to hold details of ManagementInvites such as:
InviteId
EmailAddress
EventId
DateSent
SentByUserId
DateAccepted
DateRejected
DateRevoked
CustomMessage

There may be additional columns required to support the functionality which we can adjust later.

stevejgordon:
Will work on this before the codeathon so we can pick up items during the event

Issue #1804, opened 2017-02-24
Invite mechanism for use by Org Admins


stevejgordon:
In order to support the new concepts of Campaign and Event managers in allReady we have agreed that the best process will be an invite system. This will allow an org admin to invite someone to become a campaign or event manager.

This flow prevents org admins from explicitly having to have any kind of direct user access.

The flow will be...
1) Org admin sends invite to someone to be a campaign admin (this will require an email address for the invited person).
2) The invitee will get an email - If they are already a user, we'll send a link to accept the invite. If they are not a user we'll need to include a flow that sends the user to register and then allows them to accept the pending invite.
3) Once accepted the user will be updated to include the rights to manage the campaign

It might be useful to have somewhere in the UI that shows a user any pending invites (although this is a phase 2 and not crucial for v1)

This will need to be broken down further into smaller chunks of work.

mgmccarthy:
@stevejgordon 

cc @tonysurma 

> The invitee will get an email - If they are already a user, we'll send a link to accept the invite. If they are not a user we'll need to include a flow that sends the user to register and then allows them to accept the pending invite.

in the case where the recipient is not a user of AllReady, can we instead send two emails? Once will be the email we already have, confirm your account (this is assuming we can create an account on their behalf based on them being sent a management invite) and the other email will simply be the email to confirm or Reject the Invite...

if they click on the management invite first before confirming their account with AllReady, we can easily either redirect them to the correct action method to have them finish their registration or tell them to look in their email for a account confirmation email from us and to complete that first.

just thinking of a way to not have to build a bunch of custom code looking for a registered user when a recipient is sent a management invite

PR #2009, opened 2017-05-24
Issue branch 1886

mk0sojo:
Added view, controller and handlers for manager invite functionality


mk0sojo:
Closes #1886

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
```diff
@@ -0,0 +1,154 @@
...
+                {
+                    return BadRequest();
+                }
+
+                if (!IsUserAuthorizedToSendInvite(campaign.ManagingOrganizationId))
+                {
+                    return Unauthorized();
+                }
+
+                var user = await _userManager.GetUserAsync(User);
```
mgmccarthy: remove all the code above this line with the exception of the `if(ModelState.IsValid)` check. You made all these same checks in the matching GET action method...

for example, the user would not even be able to get to see the view from which this post action method is called if their not authorized to send an invite

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
```diff
@@ -0,0 +1,154 @@
...
+                await _mediator.SendAsync(new CreateInviteCommand { Invite = invite, UserId = user?.Id });
+
+                return RedirectToAction(actionName: "Details", controllerName: "Campaign", routeValues: new { area = "Admin", id = invite.CampaignId });
```
mgmccarthy: remove the `actionName:`, `controllerName:`, etc... items. they're not needed

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
```diff
@@ -0,0 +1,154 @@
...
+                    return BadRequest();
+                }
+
+                if (!IsUserAuthorizedToSendInvite(campaign.ManagingOrganizationId))
+                {
+                    return Unauthorized();
+                }
+
+                var user = await _userManager.GetUserAsync(User);
+                await _mediator.SendAsync(new CreateInviteCommand { Invite = invite, UserId = user?.Id });
```
mgmccarthy: You should break down the invite for the approapriate type into separate command and separate handlers... for instance, instead of having `CreateInviteCommand`, break the work up into smaller logical units with a `CreateManagerInviteCommand` and `CreateEventInviteCommand`. The same goes for the handlers of those commands. the handlers should be very simple, we don't want too much logic in them.

mgmccarthy: when would `User` every be null? There is no way user could be null if the user has been authenticated, and this controller has the `[Authorize]`  attribute at the top.
Change `UserId = user?.Id` to `UserId = user.Id`

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
```diff
@@ -0,0 +1,154 @@
...
+                {
+                    return BadRequest();
+                }
+
+                if (!IsUserAuthorizedToSendInvite(@event.Campaign.ManagingOrganizationId))
+                {
+                    return Unauthorized();
+                }
+
+                var user = await _userManager.GetUserAsync(User);
```
mgmccarthy: same feedback for this POST action method, remove all the code above here except for the ModelState.IsValid check

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                return RedirectToAction(actionName: "Details", controllerName: "Event", routeValues: new { area = "Admin", id = invite.EventId });
+            }
+
+            return View("Send", invite);
+        }
+
+        private bool IsUserAuthorizedToSendInvite(int organizationId)
+        {
+            if (!User.IsOrganizationAdmin())
```
mgmccarthy: we don't need to run this check here `!User.IsOrganiztaionAdmin`... this controller has this attribute on it:
```
[Authorize(nameof(UserType.OrgAdmin))]
```
so you're making the same check twice.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
```diff
@@ -0,0 +1,154 @@
...
+                return false;
+            }
+
+            var userOrganizationId = User.GetOrganizationId();
+
+            if (!userOrganizationId.HasValue || organizationId != userOrganizationId)
+            {
+                return false;
+            }
+            return true;
```
mgmccarthy: it looks like in general, you're trying to see if the user is an org admin of the organiztaionId passed into this method. There is an extension method for that:
`ClaimsExtensions.IsOrganizationAdmin(this ClaimsPrincipal user, int organizationId)`

your code can be:
```
if (!User.IsOrganizationAdmin(organiaztionId))
{
    return false;
}
```

for this method, and that's it.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+            _context = context;
+        }
+
+        public Func<DateTime> DateTimeUtcNow = () => DateTime.UtcNow;
+
+        protected override async Task HandleCore(CreateInviteCommand message)
+        {
+            if (message?.Invite == null) throw new ArgumentException("InviteViewModel must be set");
+            if (message?.UserId == null) throw new ArgumentException("User Id must be set");
+            if (string.IsNullOrWhiteSpace(message?.Invite?.InviteeEmailAddress))
```
mgmccarthy: remove this check. InviteeEmailAddress comes from the Send.cshtml view. This view uses the InviteViewModel, which has a required attribute on InviteeEmailAddress
```
[EmailAddress]
[Required]
[Display(Name = "Email")]
public string InviteeEmailAddress { get; set; }
```

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            _context = context;
+        }
+
+        public Func<DateTime> DateTimeUtcNow = () => DateTime.UtcNow;
+
+        protected override async Task HandleCore(CreateInviteCommand message)
+        {
+            if (message?.Invite == null) throw new ArgumentException("InviteViewModel must be set");
+            if (message?.UserId == null) throw new ArgumentException("User Id must be set");
```
mgmccarthy: Remove both of these checks. We don't do field validation in handlers, that type of validation should be delegated to controllers. Who is sending us the command? So we trust that controller has done it's job and validated the fields for us

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            switch (message.Invite.InviteType)
+            {
+                case ViewModels.Invite.InviteType.CampaignManagerInvite:
+                    await CreateCampaignManagerInvite(message.Invite, message.UserId);
+                    break;
+                case ViewModels.Invite.InviteType.EventManagerInvite:
+                    await CreateEventManagerInvite(message.Invite, message.UserId);
+                    break;
+                default:
```
mgmccarthy: remove this code as it's not used

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+                SenderUserId = userId,
+                Event = @event,
+            };
+            _context.EventManagerInvites.Add(eventManagerInvite);
+            await _context.SaveChangesAsync();
+        }
+
+        private async Task<Event> GetEvent(int eventId)
+        {
+            return await _context.Events.AsNoTracking().FirstOrDefaultAsync(e => e.Id == eventId);
```
mgmccarthy: why `FirstOrDefault` here? You're filtering on a primary key, so there can logically never be more than one Event with the same Id (Id is the primary key). Change this to `.SingleOrDefault`

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+                SenderUserId = userId,
+                CampaignId = invite.CampaignId,
+            };
+            _context.CampaignManagerInvites.Add(campaignManagerInvite);
+            await _context.SaveChangesAsync();
+        }
+
+        private async Task<Campaign> GetCampaign(int campaignId)
+        {
+            return await _context.Campaigns.AsNoTracking().FirstOrDefaultAsync(c => c.Id == campaignId);
```
mgmccarthy: why `FirstOrDefault` here? You're filtering on a primary key, so there can logically never be more than one Campaign with the same Id (Id is the primary key). Change this to `.SingleOrDefault`

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Invite/Send.cshtml
```diff
@@ -0,0 +1,56 @@
+﻿@model AllReady.Areas.Admin.ViewModels.Invite.InviteViewModel
+
+@{
+    string formAction;
+    if (Model != null && Model.InviteType == AllReady.Areas.Admin.ViewModels.Invite.InviteType.EventManagerInvite)
```
mgmccarthy: when would Model ever be null? You're explicitly creating it in the Get action method. Remove this check

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Invite/Send.cshtml
```diff
@@ -0,0 +1,56 @@
...
+        ViewData["Title"] = "Send Event Manager Invite";
+        formAction = "SendEventManagerInvite";
+    }
+    else
+    {
+        ViewData["Title"] = "Send Campaign Manager Invite";
+        formAction = "SendCampaignManagerInvite";
+    }
+}
+
```
mgmccarthy: if possible, try to move these conditional checks to the controller, and then return the Title of the page and the FormAction as properties on the view model. That way, both the Title, and the FormAction are unit testable, and the view does not contain conditional logic. We want to keep our views as simple and "stupid" as possible.

mk0sojo:
Thanks for the review comments. All makes sense.
Just a thought, would it be better to have two views / controllers, one for event manger invite and one for campaign manager invite? They don't really share as much code as I expected.
Also, currently when an invite is submitted the user is directed back to the Event / campaign page. Would a confirmation page be better?

mgmccarthy:
@mk0sojo I'd say keep the controller as is... you can dispatch commands that are "entity-specific" in this case from a single controller... if it starts getting too big going forward, we can always refactor and break it apart then.

As far as a confirmation page goes, that would be something the project owners should weigh in on. If we don't do a confirmation page, there should at least be some sort of "green" looking message on the view you come back to saying that the operation was a success.

mk0sojo:
Updated with review comments. Also added details and list views. Did split into two controllers / Views as keeping it in one controller / view got a bit complicated.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/InviteController.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+            switch (message.Invite.InviteType)
+            {
+                case ViewModels.Invite.InviteType.CampaignManagerInvite:
+                    await CreateCampaignManagerInvite(message.Invite, message.UserId);
+                    break;
+                case ViewModels.Invite.InviteType.EventManagerInvite:
+                    await CreateEventManagerInvite(message.Invite, message.UserId);
+                    break;
+                default:
+                    break;
```
mgmccarthy: @mk0sojo remove both `default` and `break` here as it's not used

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Invite/CreateInviteCommandHandler.cs
mgmccarthy: @mk0sojo this still needs to be done.

AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Invite/Send.cshtml
mgmccarthy: @mk0sojo this still needs to be done.

mk0sojo:
Sorry, something seems to have gone wrong when I merged it. CreateInviteController, Invite/Send.cshtml and CreateInviteCommand / Handler were supposed to be deleted. Should be fixed now.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -0,0 +1,114 @@
...
+                    ModelState.AddModelError("InviteeEmailAddress", $"An invite already exists for {invite.InviteeEmailAddress}.");
+                    return View("Send", invite);
+                }
+
+                var userToInvite = await _userManager.FindByEmailAsync(invite.InviteeEmailAddress);
+                if (userToInvite != null && userToInvite.ManagedEvents.Any(m => m.EventId == eventId))
+                {
+                    ModelState.AddModelError("InviteeEmailAddress", $"{invite.InviteeEmailAddress} is allready a manager for this event");
+                    return View("Send", invite);
+                }
```
mgmccarthy: actually, starting from the `var userToInvite = await _userManager.FindByEmailAsync(invite.InviteeEmailAddress);` line and then into the conditional check, can we move the `userToInvite.ManagedEvents.Any(m => m.EventId == eventId)` check earlier in the process? 

Just like the `UserHasEventManagerInviteQuery` check, we can probably catch this on the get action method and not allow the user to send a message...

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -0,0 +1,114 @@
...
+
+                invite.EventId = eventId;
+                var user = await _userManager.GetUserAsync(User);
+                await _mediator.SendAsync(new CreateEventManagerInviteCommand { Invite = invite, UserId = userToInvite?.Id });
```
mgmccarthy: when would `userToInvite` ever be null? It won't. Check my previous comment and drop the nullability on `userToInvite`

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -0,0 +1,114 @@
...
+                // TODO: Send invite
+
+                return RedirectToAction(actionName: "Details", controllerName: "Event", routeValues: new { area = "Admin", id = invite.EventId });
+            }
```
mgmccarthy: can you please remove `actionName:`, `controllerName:` and `routeValues:` from this RedirectToAction call? If you've added them anywhere else in the code, could you please remove them as well? thanks.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -0,0 +1,114 @@
...
+            return View("Send", invite);
+        }
+
+        [Route("[area]/[controller]/[action]/{inviteId:int}")]
+        public async Task<IActionResult> Details(int inviteId)
+        {
+            EventManagerInviteDetailsViewModel viewModel = await _mediator.SendAsync(new EventManagerInviteDetailQuery { EventManagerInviteId = inviteId });
```
mgmccarthy: change `EventManagerInviteDetailsViewModel ` to `var` please

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            _mediator = mediator;
+            _userManager = userManager;
+        }
+
+        // GET: Admin/Invite/Send/5
+        [Route("[area]/[controller]/[action]/{eventId:int}")]
+        public async Task<IActionResult> Send(int eventId)
+        {
+            var @event = await _mediator.SendAsync(new EventByEventIdQuery { EventId = eventId });
```
mgmccarthy: just a thought. `EventByEventIdQueryHandler` loads ALL of these entities:
```
return await _context.Events
.Include(a => a.Location)
.Include(a => a.Campaign).ThenInclude(c => c.ManagingOrganization)
.Include(a => a.RequiredSkills).ThenInclude(rs => rs.Skill).ThenInclude(s => s.ParentSkill)
.Include(a => a.VolunteerTasks).ThenInclude(t => t.AssignedVolunteers).ThenInclude(tu => tu.User)
.Include(a => a.VolunteerTasks).ThenInclude(t => t.RequiredSkills).ThenInclude(ts => ts.Skill)
.SingleOrDefaultAsync(a => a.Id == message.EventId);
```

builds and returns an entire `Event`. You need a very small subset of that data, namely, you need just enough to build  your ViewModel 
```
return View("Send", new EventManagerInviteViewModel
{
    EventId = eventId,
    EventName = @event.Name,
    CampaignId = @event.CampaignId,
    CampaignName = @event.Campaign.Name,
});
```

I'd write a new query and handler (maybe name it `EventManagerInviteQuery`) and just query the data you need to create the EventManagerInviteViewModel and return that from the query handler.

FYI, some of out controllers in the solution try to work with a ViewModel per controller (aka, do not share ViewModels, which is a good a practice to follow), or go even further if the controller is large or complicated, where there is a ViewModel per GET/POST action method. 

Check out [SiteAdminController's Index action method](https://github.com/HTBox/allReady/blob/62f5f6bb1edc8d91fde7c2aeb02a663cffcd3b61/AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/SiteAdminController.cs#L44), IndexQuery and IndexQueryHandler.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
```diff
@@ -0,0 +1,113 @@
...
+            _mediator = mediator;
+            _userManager = userManager;
+        }
+
+        // GET: Admin/Invite/Send/5
+        [Route("[area]/[controller]/[action]/{campaignId:int}")]
+        public async Task<IActionResult> Send(int campaignId)
+        {
+            var campaign = await _mediator.SendAsync(new CampaignByCampaignIdQuery { CampaignId = campaignId });
+            if (campaign == null) return NotFound();
```
mgmccarthy: I'll echo the same thing I reviewed on `EventManagerInviteController`, `CampaignByCampaignIdQueryHandler` is loading a lot of data to return an entire `Campaign`.

All you need it enough data to build your ViewModel:
```
return View("Send", new CampaignManagerInviteViewModel
{
    CampaignId = campaignId,
    CampaignName = campaign.Name,
});
```

I'd write a new query and handler called CampaignManagerInviteQuery and just query for the data you need, and let the handler build and return CampaignManagerInviteViewModel

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            if (ModelState.IsValid)
+            {
+                var campaign = await _mediator.SendAsync(new CampaignByCampaignIdQuery { CampaignId = campaignId });
+
+                if (campaign == null)
+                {
+                    return BadRequest();
+                }
```
mgmccarthy: remove this null check

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
```diff
@@ -0,0 +1,113 @@
...
+                    ModelState.AddModelError("InviteeEmailAddress", $"An invite already exists for {invite.InviteeEmailAddress}.");
+                    return View("Send", invite);
+                }
+
+                var userToInvite = await _userManager.FindByEmailAsync(invite.InviteeEmailAddress);
+                if (userToInvite != null && userToInvite.ManagedCampaigns.Any(m => m.CampaignId == campaignId))
+                {
+                    ModelState.AddModelError("InviteeEmailAddress", $"{invite.InviteeEmailAddress} is allready a manager for this campaign");
+                    return View("Send", invite);
+                }
```
mgmccarthy: I'd like to move the `UserHasCampaignManagerInviteQuery` and `userToInvite.ManagedCampaigns.Any(m => m.CampaignId == campaignId)` checks earlier in the process. Try to get them into the GET Send action method so we're not allowing the user to do something that we know wont' work. We'll let the UI disallow it, instead of letting then click on stuff and then at the very end tell them, "oh, too bad, you can't perform this operation"

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
```diff
@@ -0,0 +1,113 @@
...
+
+                invite.CampaignId = campaignId;
+                var user = await _userManager.GetUserAsync(User);
+                await _mediator.SendAsync(new CreateCampaignManagerInviteCommand { Invite = invite, UserId = user?.Id });
```
mgmccarthy: user will never be null. Please make user non-nullable (remove the `?`)

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
```diff
@@ -0,0 +1,113 @@
...
+
+                return RedirectToAction(actionName: "Details", controllerName: "Campaign", routeValues: new { area = "Admin", id = invite.CampaignId });
+            }
```
mgmccarthy: remove all the `x:` argument names in the RedirectToAction call

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            return View("Send", invite);
+        }
+
+        [Route("[area]/[controller]/[action]/{inviteId:int}")]
+        public async Task<IActionResult> Details(int inviteId)
+        {
+            CampaignManagerInviteDetailsViewModel viewModel = await _mediator.SendAsync(new CampaignManagerInviteDetailQuery { CampaignManagerInviteId = inviteId });
```
mgmccarthy: use `var` here instead of ` CampaignManagerInviteDetailsViewModel`

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/CampaignManagerInvites/CampaignManagerInviteDetailQueryHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        public async Task<CampaignManagerInviteDetailsViewModel> Handle(CampaignManagerInviteDetailQuery message)
+        {
+            var invite = await _context.CampaignManagerInvites.AsNoTracking()
+                .Include(i => i.Campaign)
+                .Include(i => i.SenderUser)
+                .SingleOrDefaultAsync(i => i.Id == message.CampaignManagerInviteId);
+
+            if (invite == null) return null;
```
mgmccarthy: I'm not too sure why invite here would ever be null. Our code provides the value for `message.CampaignManagerInviteId` via the inviteId  passed to CampaignManagerInviteController.Details action method. The action method is called via our razor views, which themselves are populated from our ViewModels/Query handlers/database...

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/CampaignManagerInvites/CreateCampaignManagerInviteCommandHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            _context = context;
+        }
+
+        public Func<DateTime> DateTimeUtcNow = () => DateTime.UtcNow;
+
+        protected override async Task HandleCore(CreateCampaignManagerInviteCommand message)
+        {
+            var campaign = await _context.Campaigns.AsNoTracking().SingleOrDefaultAsync(c => c.Id == message.Invite.CampaignId);
+            if (campaign == null) throw new ArgumentException("CampaignId cannot be null for campaign manager invite");
```
mgmccarthy: remove these two lines. We don't do this type of validation in command handlers. We should validate in the controller before we send the command.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/EventManagerInvites/CreateEventManagerInviteCommandHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+            _context = context;
+        }
+
+        public Func<DateTime> DateTimeUtcNow = () => DateTime.UtcNow;
+
+        protected override async Task HandleCore(CreateEventManagerInviteCommand message)
+        {
+            var @event = await _context.Events.AsNoTracking().SingleOrDefaultAsync(e => e.Id == message.Invite.EventId);
+
+            if (@event == null) throw new ArgumentException("EventId cannot be null for Event manager invite");
```
mgmccarthy: remove this check, we don't do this type of validation in command handlers. we validate in the controller that sends the command. Again, the value message.Invite.EventId is provided by our code and read from our db... so there is no reason that the event would not be found

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/EventManagerInvites/EventManagerInviteDetailQueryHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public async Task<EventManagerInviteDetailsViewModel> Handle(EventManagerInviteDetailQuery message)
+        {
+            var invite = await _context.EventManagerInvites.AsNoTracking()
+                .Include(i => i.Event).ThenInclude(e => e.Campaign)
+                .Include(i => i.SenderUser)
+                .SingleOrDefaultAsync(i => i.Id == message.EventManagerInviteId);
+
+            if (invite == null) return null;
```
mgmccarthy: again, not too sure what the logical reason it that invite would even  be null

AllReadyApp/Web-App/AllReady/Areas/Admin/ViewModels/ManagerInvite/ManagerInviteDetailsViewModel.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
```
mgmccarthy: out of curiosities sake, why did you use inheritance for these ViewModels? I can see the perceived benefit of "re-usability" but as I said earlier in some of my review comments, we want to minimize ViewModel sharing as much as possible b/c it leads to very tightly coupled code.

You can keep it for now, but I just wanted to know why. Thanks.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -85,7 +85,7 @@ public EventManagerInviteController(IMediator mediator, UserManager<ApplicationU
 
                 invite.EventId = eventId;
                 var user = await _userManager.GetUserAsync(User);
-                await _mediator.SendAsync(new CreateEventManagerInviteCommand { Invite = invite, UserId = userToInvite?.Id });
+                await _mediator.SendAsync(new CreateEventManagerInviteCommand { Invite = invite, UserId = user.Id });
                 // TODO: Send invite
```
mgmccarthy: please remove the //TODO comment here. The dispatching of the email will not be done through the controller.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
mk0sojo: Sounds like a good idea. Just a question, what do we want to do with the IsOrganizationAdmin check on the line after that uses the oranization ID. Should I add the organization ID to the ViewModel even if it is not used in the view or create another query just to get the oranization ID even if this would require another trip to the database?

mgmccarthy: @mk0sojo I would add the OrganizationId to the ViewModel so you can make this check. The ViewModel is there to support the GET operation and the razor view(s) that consume it. Adding the OrganizationId to the ViewModel is ok even if the razor view is not using it as the controller needs to make this check.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Campaigns/CampaignDetailQueryHandler.cs
```diff
@@ -82,5 +90,13 @@ public CampaignDetailQueryHandler(AllReadyContext context)
...
             return result;
         }
+
+        private CampaignDetailViewModel.CampaignManagerInviteStatus GetCampaignManagerInviteStatus(CampaignManagerInvite campaignManagerInvite)
+        {
+            if (campaignManagerInvite.IsAccepted) return CampaignDetailViewModel.CampaignManagerInviteStatus.Accepted;
+            else if (campaignManagerInvite.IsPending) return CampaignDetailViewModel.CampaignManagerInviteStatus.Pending;
+            else if (campaignManagerInvite.IsRejected) return CampaignDetailViewModel.CampaignManagerInviteStatus.Rejected;
+            else  return CampaignDetailViewModel.CampaignManagerInviteStatus.Revoked;
+        }
```
stevejgordon: Hi @mgmccarthy - Digging my mind back and having had a quick look at the code I think I know why it was built like that. We're actually storing dates against each status occuring in the db. So the date that the invite was created, date it was accepted etc. I can't recall exactly why now, but I think it stemmed from a discussion I'd had around there needed to be a way to view this history for an invite.

The booleans for the statuses then use the dates to interpret whether those states are in effect. They are not mapped to the db. I think it was a quick thing at the time as we hadn't stubbed out all of the required functionality and I wasn't sure where/when we'd use that info.

I can't think of a driving reason to leave it like that. If we can simplify the code and use for consumers, now would be a good time. There's no reason we couldn't have an enum version and load the appropriate value into that enum based on which status dates have been meet.

mgmccarthy: @stevejgordon thanks for the info/feedback. @mk0sojo, I'd say for now, let's keep the transformations in the CampaignDetailqueryHandler as-is. This PR is already big enough, and I'd like for us to finish it up and get it merged.

I'm going to do some analysis on this and figure out the best way forward. Right now, I'm thinking we can expose the enum on ManagerInvite, but we'll need to keep the dates when the different statuses were set for what I'm assuming is a history/audit trail purpose.

mgmccarthy: @stevejgordon one more thing. In the context of a ManagerInvite, what is the difference between Rejected and Revoked? My best guess is that Rejected means the person to whom the invite was sent has no interest in being a manager ... aka "thanks but no thanks". Revoked is a way to "recall" the message before the recipient has a chance to respond OR is it a way to remove a user as the manager of an Event and/or Campaign?

stevejgordon: @mgmccarthy Yep, 100% on your guess. And to the second part, I think revoked would simply recall the message. I'd have to double check but once accepted they would be added to the appropriate managers table and therefore the invite is accepted and no longer relevant accept for history. So to remove a manager that would be a separate process updating the managers table. I think! I'm hoping to pull the latest code and refresh my memory a bit tomorrow in readiness for the standup. Been a bit out of it for a fair few weeks :-(

mgmccarthy: > And to the second part, I think revoked would simply recall the message

how do you recall an email message? You can't. Once it's sent, it's sent.

Unless the way we're sending these invites is not through email, but some internal messaging functionality we intend to build?

stevejgordon: No it's via email, but when revoked the link should no longer be valid to accept the invite so no-ops if they click it. Not sure if that's specified somewhere/included already.

mgmccarthy: @stevejgordon makes sense. 

I'll look through the Issue and dig a bit deeper. Not to sure if in addition to revoking the validity of the link to accept the invite there is any type of time-based expiration for the link, or if the invite is valid "forever".

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    return Unauthorized();
+                }
+
+                if (await _mediator.SendAsync(new UserHasEventManagerInviteQuery { EventId = invite.EventId, InviteeEmail = invite.InviteeEmailAddress }))
+                {
+                    ModelState.AddModelError("InviteeEmailAddress", $"An invite already exists for {invite.InviteeEmailAddress}.");
+                    return View("Send", invite);
+                }
+
```
stevejgordon: @mgmccarthy We cannot make registered users searchable for org admin level since we will be potentially running the site in a shared mode. Therefore it was agreed that it would have to be invite which doesn't expose addresses. So yes, you are correct, if not registered the user will need to first signup. If the action handling the link is behind auth, then this is pretty much taken care of as the user will have to sign in, or first sign up. Then they will be able to click the invite link.

Not included yet, but I think will follow is some kind of notification when they login, if they have invites pending. That makes the flow smoother for unregistered user invites.

mgmccarthy: @stevejgordon  sorry about all these questions... I should have just traced down and read #1804 and #1885 first... I'm doing that now.

cc @mk0sojo once I understand how this piece fits into the whole of inviting org admins, I'll continue some of my analysis and/or update some of my code review findings....

I've been out of the loop too long! Gotta get back into it

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -0,0 +1,114 @@
...
+                var userToInvite = await _userManager.FindByEmailAsync(invite.InviteeEmailAddress);
+                if (userToInvite != null && userToInvite.ManagedEvents.Any(m => m.EventId == eventId))
+                {
```
mk0sojo: The userToInvite might be null if the user we try to invite is not registered yet. The check is just there to make sure we don't send an invite to someone who is already an event manager.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
mk0sojo: As for the event manager invite, the validation needs to be done in post as the user enters the email in a text box

mgmccarthy:
Yep. I realize now that we're allowing free form text box with email and why these validations are where they are

mk0sojo:
Updated. Hopefully the review comments should have been fixed now.

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/CampaignManagerInvites/CampaignManagerInviteQueryHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+        private AllReadyContext _context;
+
+        public CampaignManagerInviteQueryHandler(AllReadyContext context)
+        {
+            _context = context;
+        }
+
+        public async Task<CampaignManagerInviteViewModel> Handle(CampaignManagerInviteQuery message)
+        {
+            var campaign = await _context.Campaigns.AsNoTracking().SingleOrDefaultAsync(c => c.Id == message.CampaignId);
```
mgmccarthy: change .SingleOrDefault to .Single

AllReadyApp/Web-App/AllReady/Areas/Admin/Features/EventManagerInvites/EventManagerInviteQueryHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public EventManagerInviteQueryHandler(AllReadyContext context)
+        {
+            _context = context;
+        }
+
+        public async Task<EventManagerInviteViewModel> Handle(EventManagerInviteQuery message)
+        {
+            var @event = await _context.Events.AsNoTracking()
+                .Include(e => e.Campaign).SingleOrDefaultAsync(e => e.Id == message.EventId);
```
mgmccarthy: change .SingleOrDefault to .Single

mk0sojo:
Thanks. Those review comments should be fixed now.

MisterJames:
@mgmccarthy Looks like we should be able to wrap this one up now?

mgmccarthy:
@MisterJames yep

Issue #2011, opened 2017-06-16
Update Developer Documentation


stevejgordon:
After the VS 2017/csproj update https://github.com/HTBox/allReady/blob/master/docs/prerequisite_install_guide/prerequisite_install_guide.md is now out of date and misleading for new contributors.

We need to review this and ensure it reflects the current on-boarding/setup process for devs.

marigerr:
If this is open to new contributors, I can help with this.   I recently got everything installed with VS 2017, so familiar with some of the needed changes to documentation.

stevejgordon:
@marigerr Yes, please feel free to jump in on this. Would be greatly appreciated as you have the most recent experience.

marigerr:
great!  I'll get started.

jamesatgreymetis:
I worked through the guide yesterday and needed to install js tools from scratch, if it helps here is a snapshot of the current version numbers once I was finished:

git version 2.13.1.windows.1
node 6.11.0
npm 3.10.10
bower 1.8.0
gulp CLI version 1.3.0

The step to upgrade npm to v3 was not needed as it was already v3.10.10

Installing bower returned this message:
npm WARN deprecated bower@1.8.0: ..psst! While Bower is maintained, we recommend Yarn and Webpack for new front-end projects! Yarn's advantage is security and reliability, and Webpack's is support for both CommonJS and AMD projects. Currently there's no migration path but we hope you'll help us figure out one.

marigerr:
Thanks @jamesatgreymetis   I haven't had any problem myself running the latest versions of everything, so I was going to assume that it's ok to include latest versions of all the tools in docs?

I'm planning to do a new install of everything on a second machine and get screenshots for both VS 2017 installs and latest versions of node, git installs etc.  

Unsure if they are planning to switch from bower to webpack (or other tool), but might be useful for timebeing to include something in documentation that can ignore the warning.

MisterJames:
Closed via #2014

iloabn:
Last comment says that this is closed, but it still seems open.
I tried to install the project yesterday and noticed a slight difference between the short and longer installation instructions as well as a missing step.

The [short](https://github.com/HTBox/allReady/wiki/Developer-Setup) instructions doesn't mention .NET Core 2 SDK or the `F# language support`.

The [longer](https://github.com/HTBox/allReady/blob/master/docs/prerequisite_install_guide/prerequisite_install_guide.md) mentions `F# language support` but not .NET Core 2 SDK.

[...]

PR #2014, opened 2017-06-22
Issue 2011 update Prerequisite Install Guide to include VS2017

marigerr:
[Issue 2011](https://github.com/HTBox/allReady/issues/2011)  
PR updates the Web-App portion of the install guide to latest versions of tools and VS2017.

**Question:**
Has the mobile app been migrated to VS2017?    
When trying to open the mobile app with VS2017, I got the message about a one-way upgrade.

I've started to update Mobile app portion of guide, but not included yet in PR.


stevejgordon:
@marigerr The mobile app hasn't been touched in some time. I didn't migrate it since I wasn't very clear on it's direction. As part of v1.6 milestone which will start to populate soon, the mobile app does need to be re-invigorated so we should get that updated as necessary. I'll open an issue for that.

cc/ @tonysurma re. mobile app.

marigerr:
@stevejgordon  
Thanks!  
I can submit a 2nd PR (or add onto this one) to update the mobile app part as well since it will be migrated.  
However, I'm not familiar with the mobile-app, so have a couple questions.

- In the guide states:   
_"The mobile application uses the Ionic framework version 1"_.  
But later in instructions when installing Ionic with `npm install -g ionic`   , version 3.4.0 will be installed.  Is the latest version of Ionic ok, or should I add instructions on installing version 1?  
- When running `ionic state restore` , how should these 2 questions be answered? Both Yes?  

     _Detected .\gulpfile.js in project directory, but to integrate gulp with the CLI, you'll need to install ionic/cli-plugin-gulp. Install now? (Y/n)_   

     _Looks like this is an Ionic 1 project, would you like to install ionic/cli-plugin-ionic1 and continue? (Y/n)_ 

- VS2017 has tools bundled differently.  When running the Visual Studio Installer, to add the additional mobile tools, what is needed?   Did I miss anthing such as iOS or Windows phone tools or are they included with Cordova?
- [x] Tool for Apache Cordova
- [x] Java SDK 8
- [x] Android SDK setup (API level 19 and 21)
- [x] Android SDK setup (API level 22)
- [x] Android SDK setup (API level 23)
- [x] Apache Ant

stevejgordon:
I'd suggest a second PR since this one is good to go with the bits we need most urgently for new devs.

I'm not familiar with the mobile project to be able to answer those questions. It's taken a bit of backseat in the last 12 months as the functionality for the main site was updated/defined.

I think it's important we get docs updated so people can work on it again for v1.6 but possibly we make docs a P2 for the v1.6 milestone. Also will depend on the issue #2015 being completed as that may reveal some twists and turns that require changes.

marigerr:
okdoke, makes sense to hold off on updating the mobile part of the guide until after the mobile app has been migrated.
I wonder if it might be helpful to add a short note to the guide after the web-only portion, to indicate that the mobile app and documentation is pending update to VS2017.   
This would help avoid confusion if someone tries to follow the whole guide and comes across the mobile portions still referencing VS2015.

stevejgordon:
@marigerr That's a good idea, yes please.

Issue #1969, opened 2017-04-18
Consider creating a MobileAppOnly solution


SteveKennedy:
Consider creating a Mobile App only solution file, that contains just the mobile app project.  This will allow mobile developers to optionally focus on just aspects the mobile project.  Please assign this the mobile app label.

SteveKennedy:
These are just my _random_ notes I made during this task/issue, in case they're helpful to anyone in future:

1. It appears that the AllReadyWebOnly.sln has been upgraded to Visual Studio Community 2017 (Ver.15.0.26228.12).  It does not appear to contain the mobile project. 

2. The AllReadyApp.sln is still Visual Studio 2015 Update 3 (Ver.14.0.25420.1).  It has the mobile app project, but most of the projects in the solution fail to load because of bad references to .xproj files. I'm assuming the AllReadyApp.sln is out-dated and needs sunsetted.

3. The mobile project looks a little neglected. The last commit to the Mobile-App folder was Feb 24 (Latest commit 2f89fd3  on Feb 24).  The main .jsproj file was last modified over a year ago.  When loading & running the mobile app project, I get numerous JS console errors, some of which lead me to believe they contact a localhost version of an API - which might indicate coupling of web project controllers and the mobile app.

4. I will try to make notes of features/functions built in the mobile app, so that I can ensure they work as they did before the Visual Studio 2015 update.

a. Check-In = has ONLY bar code scanner
b. Event = shows up coming events in a list. Click on it, and it shows details. Very basic info.
c. Hours = acts like a stop watch. But doesn't persist?
d. Login = which does seem to verify auth, but not much more than that. No special difference.
e. Cache-Manager.service = seems to be for offline mode availability, but if so, dicey cause lazy.

5. Since the primary web application and associated projects load in AllReadyWebOnly.sln, I'm going to attempt to create a new solution (AllReadyMobileAppOnly.sln) that hosts only the mobile application.  I plan to do the bare minimum needed to update the main solution and associated projects, enough for the solution:

a. To build cleanly out-of-the-box.
b. To pass any existing tests (if any) 
c. To successfully flow through any existing automated builds, CI/CD, etc. process.
d. Adjust .gitIgnore for any new(er) machine-specific or user-specific file types associated with update.

6. I have VS 2017 Enterprise, but will download Visual Studio Community 2017 Version 15.2 to ensure availability & compatibility for future developers. 

[...]

PR #2028, opened 2017-07-08
Update Mobile-App To Visual Studio 2017

SteveKennedy:
This should address both Issue #2015 and #1969. Thoughts?


SteveKennedy:
These are just my _random_ notes I made during this task/issue, in case they're helpful to anyone in future:

1. It appears that the AllReadyWebOnly.sln has been upgraded to Visual Studio Community 2017 (Ver.15.0.26228.12).  It does not appear to contain the mobile project. 

2. The AllReadyApp.sln is still Visual Studio 2015 Update 3 (Ver.14.0.25420.1).  It has the mobile app project, but most of the projects in the solution fail to load because of bad references to .xproj files. I'm assuming the AllReadyApp.sln is out-dated and needs sunsetted.

3. The mobile project looks a little neglected. The last commit to the Mobile-App folder was Feb 24 (Latest commit 2f89fd3  on Feb 24).  The main .jsproj file was last modified over a year ago.  When loading & running the mobile app project, I get numerous JS console errors, some of which lead me to believe they contact a localhost version of an API - which might indicate coupling of web project controllers and the mobile app.

4. I will try to make notes of features/functions built in the mobile app, so that I can ensure they work as they did before the Visual Studio 2015 update.

a. Check-In = has ONLY bar code scanner
b. Event = shows up coming events in a list. Click on it, and it shows details. Very basic info.
c. Hours = acts like a stop watch. But doesn't persist?
d. Login = which does seem to verify auth, but not much more than that. No special difference.
e. Cache-Manager.service = seems to be for offline mode availability, but if so, dicey cause lazy.

5. Since the primary web application and associated projects load in AllReadyWebOnly.sln, I'm going to attempt to create a new solution (AllReadyMobileAppOnly.sln) that hosts only the mobile application.  I plan to do the bare minimum needed to update the main solution and associated projects, enough for the solution:

a. To build cleanly out-of-the-box.
b. To pass any existing tests (if any) 
c. To successfully flow through any existing automated builds, CI/CD, etc. process.
d. Adjust .gitIgnore for any new(er) machine-specific or user-specific file types associated with update.

6. I have VS 2017 Enterprise, but will download Visual Studio Community 2017 Version 15.2 to ensure availability & compatibility for future developers. 

7. We need to update any documentation, to inform future developers that the Mobile Application relies heavily on the Controller methods in the web project. Which means, the developer will also need to be running the web project. This ensures clarity, that the sole authoritative source for the Controller/API is via the web only project.

8. Long term, I think it might be good to upgrade this project to use IONIC 3/ANGULAR 4/Typescript.

MisterJames:
Looks good @SteveKennedy, thanks!

PR #2032, opened 2017-07-09
Eager-loading important Event properties into the EventViewModel

### LINKED ISSUE #2030: Provide important Event properties via Event Api
The Event API (EventApiController.cs) returns a list of EventViewModel.  That view model is populated by the EventsWithUnlockedCampaignsQueryHandler, which populates Events pulled from the data context.  

However, when the events returned do not contain some associated campaign data.  Meaning, API consumers (like the Cordova mobile app) will not receive or show associated campaign data to the end user. 

Consider eager-loading these missing objects into the EventViewModel, so  that the API consumer can get Campaign, Organization, Location, and Task information associated with an Event.

prtscr capture

SteveKennedy:
Eager-loading important properties into the EventViewModel, so that the API consumer can get Campaign, Organization, Location, and Task information associated with an Event. Added unit test.

Closes #2030


MisterJames:
:shipit:

PR #2059, opened 2017-08-06
Added UserDocTemplate.mdt (Markdown template)

DanKline:
Added Image/ALLReady.png


MisterJames:
Looks pretty good as a starting point! Let's proceed with this and we'll amend it as we go if we need to.  Thanks @DanKline!

Issue #2019, opened 2017-07-01
Data reporting: Export formats


EmilyLuijbregts:
Breakout from #2016 and #2018 

The dashboard built in #2018 needs to have reports that can be exported in the formats:

- .csv
- .xls
- .pdf

binaryjanitor:
Mine! @binaryjanitor

binaryjanitor:
Mine @binaryjanitor

binaryjanitor:
Closed!

PR #2064, opened 2017-08-06
#2016 & #2019 Dashboards including Export to various file formats; Closes #2016

### LINKED ISSUE #2016: Data Reporting - EPIC
**Summary**: The event manager is where you can see all of the events that are scheduled. We need to have an overview of seeing all of the possible data that can be mined from the events manager.

**What we want:** As an end user i want an easy to use dashboard for seeing all of the data that has been entered in the events manager, have the ability to create reports from this information and have it exportable to be able to report this to key stakeholders.

- [ ] We need to have a list of all the fields that are available for use in reports. This list can be provided in excel or another user friendly format (i.e. not a txt file). (see #2017) - Once this list is provided there will be further breakdowns of required reports by Pascal and Jim

- [ ] Initial fields that should be in the portal are:
-  how many volunteers in total were in a specific day (there should be a date range available in calendar format)
- how many hours were spent in total
- time tracking per volunteer
- human asset tracking
- how many smoke detectors were handed out in a specific day
- age of volunteers 
- gender
- age

- [ ] We need to have a dashboard built within the portal that would allow reports to be created and exported in .csv,.xls, .pdf.

binaryjanitor:


MisterJames:
@binaryjanitor I think I'm good with these changes...but do you think we should add packages for those CSS/JS dependencies, or are the CDN versions fairly solid?

binaryjanitor:
@MisterJames Yeah, these CDNs seem stable. Could be a future refactoring.

PR #2070, opened 2017-08-12
Hiding team management UI when user is not authorized to manage the team

### LINKED ISSUE #1992: Team Management by Team Lead
Breaking this out from #1787 Itinerary Team Lead EPIC.  

I made the following comment:
>At the moment, the Team Lead can manage team membership including changing the Team Lead which I don't think is a good idea. The Team Lead should only be able to add and remove non-team lead members. 

IMO - this is a _business rule_ related question.  Thoughts?  Is this even an issue?

stevejgordon:
Fixes #1992 

- Added property on detail view model which will be used to hide/show the team management UI elements
- Set the property from the AuthorizableItinerary.CanManageTeamMembers value
- Update UI with conditional blocks to hide the team management UI elements when TeamManagementEnabled is false
- Added test to validate value is being set as expected.


MisterJames:
Thanks @stevejgordon, good side fixes, too, as always!

PR #2073, opened 2017-08-14
When register a user or organization account, TimeZoneId should be from user's local TimeZoneInfo instead of from config.json

JingWendyLiu:
- Replaced the user TimeZoneId retrieved from config.json:General:DefaultTiemZone with the user's TimeZoneInfo.Local.Id in Account and Admin Controller;
- Updated the corresponding unit tests.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
+++ b/AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
@@ -1,4 +1,5 @@
-﻿using System.Security.Claims;
+﻿using System;
+using System.Security.Claims;
 using System.Threading.Tasks;
 using AllReady.Areas.Admin.Controllers;
 using AllReady.Configuration;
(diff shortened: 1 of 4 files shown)


dpaquette:
Thanks for this @JingWendyLiu. Unfortunately, this will not work because `TimeZoneInfo.Local` represents the local timezone of the server process, not the local timezone of the user who made the request. In the case of the application running in Azure (or on most web hosting providers), that would be UTC. Using `TimeZoneInfo.Local` here works for developers running the application from Visual Studio because it will default new users to the developer's local time zone but it will not work for the hosted application in Azure, at least not in the way we would like it to.

AllReadyApp/Web-App/AllReady/Controllers/AccountController.cs
```diff
@@ -23,11 +24,11 @@ public class AccountController : Controller
     {
         private readonly UserManager<ApplicationUser> _userManager;
         private readonly SignInManager<ApplicationUser> _signInManager;
-        private readonly IOptions<GeneralSettings> _generalSettings;
         private readonly IMediator _mediator;
         private readonly IExternalUserInformationProviderFactory _externalUserInformationProviderFactory;
         private readonly IRedirectAccountControllerRequests _redirectAccountControllerRequests;
         private readonly string _externalCookieScheme;
+        private readonly TimeZoneInfo _timeZone = TimeZoneInfo.Local;
```
JingWendyLiu: Thanks for your time David. Sorry, my bad.  :( 
I'll wait for  the update on solution of solving the default timezone for a new account. Like implementing nodatime as you suggested in Slack.

dpaquette: No need to apologize. The DateTime and TimeZone code in AllReady is confusing and that is not your fault.

PR #2088, opened 2017-08-24
Added email sending to event and campaign manager invites

### LINKED ISSUE #1889: Send ManagementInvite Email to Invitee
Breakout from #1804

We should add a notification (Mediatr) and handler to support adding an email message onto the queue when an invite is sent. This should send the invitee the email, with a link to accept the invite. We should perhaps include an optional link to decline the invite?

We should also include details of where to register if they do not have an account already.

mk0sojo:
Added functionality to send email invite when creating a new campaign or event manager invite

closes #1889


mk0sojo:
closes #1889

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/CampaignManagerInviteController.cs
```diff
@@ -68,15 +70,33 @@ public CampaignManagerInviteController(IMediator mediator, UserManager<Applicati
...
                     ModelState.AddModelError("InviteeEmailAddress", $"{invite.InviteeEmailAddress} is allready a manager for this campaign");
                     return View("Send", invite);
                 }
 
                 invite.CampaignId = campaignId;
                 var user = await _userManager.GetUserAsync(User);
-                await _mediator.SendAsync(new CreateCampaignManagerInviteCommand { Invite = invite, UserId = user.Id });
+                var inviteId = await _mediator.SendAsync(new CreateCampaignManagerInviteCommand { Invite = invite, UserId = user.Id });
+
+                await _mediator.PublishAsync(new SendCampaignManagerInvite
```
MisterJames: When publishing events, the event should be raised from where the action happened.  The event should be phrased as an event `SomethingHappened` and not a command `DoSomething`.  So just a couple of naming/location things to sort out.

As an example, if the event were named `CampaignManagerInvited` and raised from within the `CreateCampaignManagerInviteCommand` I think we'd be on track.

AllReadyApp/Web-App/AllReady/Areas/Admin/Controllers/EventManagerInviteController.cs
```diff
@@ -68,16 +70,33 @@ public EventManagerInviteController(IMediator mediator, UserManager<ApplicationU
...
                     ModelState.AddModelError("InviteeEmailAddress", $"{invite.InviteeEmailAddress} is allready a manager for this event");
                     return View("Send", invite);
                 }
 
                 invite.EventId = eventId;
                 var user = await _userManager.GetUserAsync(User);
-                await _mediator.SendAsync(new CreateEventManagerInviteCommand { Invite = invite, UserId = user.Id });
+                var inviteId = await _mediator.SendAsync(new CreateEventManagerInviteCommand { Invite = invite, UserId = user.Id });
 
+                await _mediator.PublishAsync(new SendEventManagerInvite
```
MisterJames: Moving events inside of the trigger that causes them prevents having to publish them from multiple locations.

MisterJames:
@mk0sojo Thanks for this PR!! Great work!

I added a couple of notes related to the location/naming of the events. I think it will also simplify a couple of your tests (i.e. you'll only have to test that the handler of the command publishes the event, and you won't have to test for it elsewhere). 

Typically in an commanding/eventing scenario we prefer to see the command that executes to raise the event, rather than relying on the caller of the command to do so. This prevents errors of omission later on, where someone would invoke a command but forget (or not know of the responsibility to) publish the event. 

Cheers!

mk0sojo:
Thanks for the comments @MisterJames .
The code has been updated.

MisterJames:
Great, thanks @mk0sojo !

PR #2098, opened 2017-08-31
Fix date field on edit campaign view

### LINKED ISSUE #2097: Start and end date validations and datepicker broken in edit campaign view
it's showing as a datetime field when it should be a date and showing both the bootstrap datepicker and what looks like a built in asp.net one

image

sdekok:
Input type was not set on start date and end date resulting in asp.net setting it to datetime-local which added validations that expected a time and it's own datepicker that conflicted with the bootstrap datepicker.

Fixed by setting the type to text.

Fixes #2097


sdekok:
@dpaquette, this is the pull request for the datepicker issue we discussed on slack

MisterJames:
Thanks @sdekok and @dpaquette !

Issue #2104, opened 2017-09-16
Make radio buttons and checkboxes great again


jgauffin:
The `.form-control` class is not intended to be used on radio buttons and checkboxes.
Thus they look not so great when the class is used.

jgauffin:
Pull request: #2105

MisterJames:
Closed via #2105

PR #2105, opened 2017-09-16
#2104 - Updated how checkboxes and radio buttons appear

jgauffin:
Added a specialization of the `form-control` css class so that it works for checkboxes and radio buttons too.


MisterJames:
@jgauffin got the build running here as well (seemed to be some stickiness with the bower process, just restarted build). Cheers!

Issue #2117, opened 2017-09-17
Use azure-functions to port legacy Web Jobs code


MisterJames:
The new functions project can be used to house the existing web jobs. The intent here is to deprecate the web jobs project and use the new functions project going forward. 

PRs for this should be targeted to `azure-functions`.

We will need to move:
 - `ProcessSmsQueueMessage`
 - `ProcessEmailQueueMessage`

This can be done independently of the build/deployments steps in #2086

MisterJames:
Closed with #2123

PR #2123, opened 2017-09-22
Port legacy WebJobs to Azure Functions

tony-ho:
Issue #2117 

Port existing `ProcessSmsQueueMessage` and `ProcessEmailQueueMessage` WebJobs to Azure Functions.

Consistency with the existing WebJobs has mostly been retained except for the following changes:
- Storage account application setting changed to `AzureWebJobsStorage` from `Data:Storage:AzureStorage` as this is the default setting for Function Apps
- Use Twilio output binding instead of REST client
ref: https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.Twilio/1.0.0
- Use SendGrid output binding instead of Web transport
ref: https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.SendGrid/2.0.0
- `Sendgrid` NuGet package updated to 8.0.5 as required by `Microsoft.Azure.WebJobs.Extensions.SendGrid`. Note: `Sendgrid` version 9.0.0 is not supported by `Microsoft.Azure.WebJobs.Extensions.SendGrid` version 2.0
- As a side effect of using the SendGrid output binding, the output mail message is constructed using the `SendGrid.Helpers.Mail.Mail` class instead of `SendGrid.SendGridMessage`
- Another side effect of using the SendGrid output binding is that authentication must be done using an API key instead of username/password. A new application setting `Authentication:SendGrid:ApiKey` has been added to accommodate this.

### DIFF
--- a/AllReadyApp/NotificationProcessing/NotificationProcessing.csproj
+++ b/AllReadyApp/NotificationProcessing/NotificationProcessing.csproj
@@ -3,7 +3,15 @@
     <TargetFramework>net461</TargetFramework>
   </PropertyGroup>
   <ItemGroup>    
-    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="1.0.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.SendGrid" Version="2.0.0" />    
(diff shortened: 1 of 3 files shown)


MisterJames:
@tony-ho brilliant! Thanks for this!  I'll review and merge in later today.

PR #2126, opened 2017-09-30
App Insights TelemetryConfiguration error.

dchristensen:
I think there is a residual error from the .NET core 2.0 upgrade.  Working from a completely clean installation I'm getting the following exception:

````
ComponentNotRegisteredException: The requested service 'Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration' has not been registered. To avoid this exception, either register a component to provide the service, check for service registration using IsRegistered(), or use the ResolveOptional() method to resolve an optional dependency.
````

I can't see anywhere this is used, so I've just removed the import.


MisterJames:
I actually just removed this in my own branch and was about to PR this lol. 

Thanks @dchristensen!

PR #2128, opened 2017-10-03
#2109 - Date picker on event page broken in responsive layout

### LINKED ISSUE #2109: Date picker on event page broken in responsive layout
Example from {rootUrl}/Event/9

Screenshot

dchristensen:
Fixes #2109 

This PR changes the CSS so that the `max-width` on form inputs is only applied to screens larger than 768px.  In smaller layouts form fields are stacked to preserve horizontal space, so we should be utilizing all of the screen anyway.

Below is a screenshot of the corrected layout:
Screenshot


MisterJames:
@dchristensen thanks so much for the explanation, issue reference and screen shot. This makes reviewing so much easier!

Issue #2148, opened 2017-10-05
Some validations issues at Form


sofilibi:
1.- Go to http://allready-test.azurewebsites.net/Account/Register
2.- Complete the form
![evidence_defect_form_validation_chrome](https://user-images.githubusercontent.com/4973956/31241074-61207878-a9c8-11e7-83dc-3d5ab49bf264.jpg)

3.- There are some missing validations, like: Name/last name should allow only alpha values
4.- Phone number text should allow only numbers

joro550:
Raised https://github.com/HTBox/allReady/pull/2160 for this issue

stevejgordon:
Closing as a PR was merged for this issue.

thelem:
This has been implemented differently to what is described above, see #2160 for details (the pull request mentioned looks to be a followup rather than the main changes)

PR #2160, opened 2017-10-13
#2148 - Adding regular expressions

joro550:
Altered the regular expression for first/last name
* Full disclosure I was wondering whether I needed to support spaces and hyphens (-) for double barreled names but seeing as the issues says "Name/last name should allow only alpha values" I thought I'd implement that and await feedback

Added regular expression to phone number
* Again full disclosure this does not account for the ++ area codes i.e. ++44 for uk, again issue states "Phone number text should allow only numbers"

Happy to change if this doesn't reflect what you want/need 

Thanks :)

### DIFF
--- a/AllReadyApp/Web-App/AllReady/ViewModels/Account/RegisterViewModel.cs
+++ b/AllReadyApp/Web-App/AllReady/ViewModels/Account/RegisterViewModel.cs
@@ -5,12 +5,12 @@ namespace AllReady.ViewModels.Account
     public class RegisterViewModel
     {
         [Required]
-        [RegularExpression(@"^(?!\W).+$", ErrorMessage = "The {0} should not contain any special characters.")]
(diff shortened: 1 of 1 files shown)


MisterJames:
Hi @joro550, I think you're onto something there.

I know folks named D'Arcy and Jenna-Rae and Renée and they'd be excluded here. I think we either have to be more inclusive, or explicitly exclusive.  

Here's another approach: https://regex101.com/r/qZ9uL9/1

`(.*?)\s([\wáâàãçéêíóôõúüÁÂÀÃÇÉÊÍÓÔÕÚÜ]+\-?\'?\w+\.*?$)`

We can use this on both first and last name.
For the phone number...what you have is great and we can update it as needed.

MisterJames:
Thanks again @joro550!

PR #2162, opened 2017-10-16
Fixing #2152 open the campaign info by default

### LINKED ISSUE #2152: StarWest Usability - Safety Kit Distribution Weekend - Campaign Info expand on load
As a User in the - Safety Kit Distribution Weekend - the Campaign Info field at the top should be expanded on load, to allow for visibility of where the user is in the site.

nemesv:
Fixes #2152

The `accordion` binding already supported the `openItem` option however there was a bug in the implementation which prevented to open the first element:

Because this condition evaluates to false when `openItem: 0`
```
if (openItem) {	
    items.eq(openItem).collapse("show");
};
```

The fix is to check for `openItem > -1`. I've found a similar binding which also has the same fixed condition
https://stackoverflow.com/questions/30893880/collapsable-panel-with-bootstrap-and-knockout/30948743

The `parseInt` change was required to make the openItem option "optional" otherwise when the `openItem` option was omitted it always opened the first accordion element.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Views/Event/EventWithTasks.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Views/Event/EventWithTasks.cshtml
@@ -65,7 +65,7 @@
                     <p>@Model.Description</p>
                 </div>
             </div>
-            <div class="row" id="CampaignInfo" data-bind="accordion">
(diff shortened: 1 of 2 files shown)


MisterJames:
Thanks @nemesv, looks great.

PR #2165, opened 2017-10-17
Remove dependency on bower

### LINKED ISSUE #2127: Discussion: Remove Bower dependency
Currently when running `npm install` for the first time, we are getting the following warning:

```
npm WARN deprecated bower@1.8.0: ..psst! While Bower is maintained, we recommend Yarn and Webpack for *new* front-end projects! Yarn's advantage is security and reliability, and Webpack's is support for both CommonJS and AMD projects. Currently there's no migration path, but please help to create it: https://github.com/bower/bower/issues/2467
```

I think it would make sense to remove the additional Bower dependency and just use npm to manage client side dependencies.  This should remove a couple of steps from the initial project setup and in general make it easier for new contributors to get started.

VishalMadhvani:
Removes use of bower in favour of NPM.
Maintained bower'esque behaviour of copying across entire module folder to wwwroot/lib.

fixes #2127


dchristensen:
@VishalMadhvani @dpaquette Is this going to require new developers to run the gulp script in order to get started?  If so, that probably needs to be documented, since currently gulp was not previously required to get up and running in the dev environment.  I'm not super familiar with npm, but would it be possible to run the gulp script when `npm install` is run to avoid having to instruct new developers they need to run gulp?  We should also remove the bower sections from existing docs.

VishalMadhvani:
@dchristensen done :)
Happy up update the wiki's once this is merged in.

PR #2171, opened 2017-11-10
Fixing incorrect javascript path after refactor of bower

stimms:
Looks like there is a slight change from bower to npm. This should address the path differences.


MisterJames:
Thanks @stimms!

PR #2172, opened 2017-11-11
Converting HomeController pages/action to Razor pages

stevejgordon:
NOTE: We were doing some interesting stuff to work out the current page in order to modify the menus. I had to remove this for now since it doesn't work without some re-work.

I felt it was better to get the example of Razor pages in and revisit the menu highlighting later since it's not a critical function.


VishalMadhvani:
any reason for switching to razor pages over mvc?

stevejgordon:
It's mostly about simplicity. Most of the actions on the HomeController were just there to return a view, with no logic or model behind them. With Razor Pages we can just have the page and no worry about maintaining redundant Actions.

MisterJames:
@VishalMadhvani We're thinking that we're going to refresh the tech a little. I'll create an issue or wiki entry on it with the update, but our plan is loosely: Razor Pages for simple, MVC + Razor for more complex rendering/server side bits, and MVC + React for views with complex client UX.

VishalMadhvani:
@MisterJames awesome... interested in what you have in mind for the MVC + React stuff because I was thinking along the same lines. I want to look at simplifying the front end dependencies and build.

PR #2175, opened 2017-11-12
Fix Security.cs Formatting

khalidabuhakmeh:
Saw that the formatting was off, so I reformatted it to
be correct.


MisterJames:
Thanks @khalidabuhakmeh!

Issue #2190, opened 2017-12-12
Add C# based Integration Tests


stevejgordon:
I have recently been trying with some integration testing on a demo project using the TestServer. This supports firing up the application in memory and hitting endpoints using a client to validate the responses.

I know we have some Scenario tests written using F#, however I have never run those personally. I know for some they do not load since the F# components need to be installed also. Having reviewed them briefly there seems to be quite a lot of ceremony to configuring the flow.

I therefore suggest we look at introducing some C# based integration tests. These may augment or replace the F# ones. They may be more maintainable for those without an F# background too.

@MisterJames What are you thoughts?

I have a branch with an initial demo of the concept which I could submit a PR for. We could then look to build these over time as part of on-going issues?

oneolddev:
I've started implementing a browser based testing project implemented in C#/Selenium/XUnit/Core 2.1.  I'm hoping it will address the issues in the AllReady.ScenarioTest project - please see #2345.  The project takes no dependencies from any other projects in the AllReady solution.

PR #2196, opened 2018-01-02
Integration test1

stevejgordon:
Adds an initial structure and some basic integration tests against allReady which use the TestServer.

@MisterJames - Could you take a look and see if this is worth including? My gut is that it makes sense to have this level of testing to fully validate the site functions as expected since it's getting more complex as time goes on. It's also a good learning example.

If it looks okay I think we can get it into the project and we can expand on the tests over time.

Relates to #2190


MisterJames:
Okay, good start here @stevejgordon. I'm seeing in the slacks that we're also on task to kill the F# code, which would be fine given this work. I'd also like to have a conversation in our standup (Tuesday, if you can join us) about this and the longer term value/burden position.

PR #2199, opened 2018-01-06
Issue branch 340

gparlakov:
Hello, I want to tackle the Controller names magic strings issue.

I found 3 use cases:
1. Need the name from the context of the Controller itself
2. Need the name for another Controller from the context of a Controller
3. Need the name for a Controller from a general context 

Case 3 is covered by the ControllerNames.NameOf and cases 1. and 2. are using it and are just convenience methods to use from the context of a controller

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Constants/ControllerNames.cs
+++ b/AllReadyApp/Web-App/AllReady/Constants/ControllerNames.cs
@@ -1,7 +1,27 @@
-﻿namespace AllReady.Constants
+using System;
+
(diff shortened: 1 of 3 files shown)


MisterJames:
Hrm....I want to discuss this with a few regulars on the team.

IMO, if we're going to try to tackle the magic names on controllers, why don't we just instead drop the `Controller` suffix?  We inherit from `Controller` anyway, and our controllers all live in a `/Controllers` folder. Maybe we don't need the extra chars?

Issue #2042, opened 2017-08-01
Ability to add information to a request


EmilyLuijbregts:
This is a breakout from Events Epic #2039 

**Requirement**:
-	As a user I want to be able to add information to a request. This information is the detail for what the volunteer needs to be doing so should not be limited to a short number of characters.

**Definition of Done:**
- It should be possible for [Specified users] to add information to a request

**Additional Information:**
Information from Steve: THIS IS POSSIBLY COMPLETE. We have a description field for Tasks. Is this sufficient?

**Questions for @OhMcGoo** 
- Please confirm if the question from @stevejgordon is sufficient.
- Please confirm who should be able to add information to a task (i.e. just admins? all users?)

MisterJames:
I think we need a "comment stream" concept that is associated with the request. 

We need a `RequestComments` table and the UI related to add comments to the request. We should capture:

 - userID
 - create date
 - message
 - request ID

There will need to be a command/handler written to save comments to the request.

There could be communication opportunities applied later, but for now, we'll limit this to collecting a set of non-threaded comments to a request.

stevejgordon:
I'm picking this up.

stevejgordon:
Work restarted as I somehow lost my branch for this!

stevejgordon:
I've submitted a partial PR for this. Once merged this can be considered available for the codeathon. There's still a fair amount of work left to do and possibly we can break into smaller pieces. I can advise further in person.

PR #2216, opened 2018-01-15
Initial work for Request Comments

stevejgordon:
I'm submitting this partial PR which starts the work for #2042 

This piece is safe to add at this stage and includes the new schema (initial) and migration. It also includes the command to add a request comment and the initial work on a view model.

I'm submitting this now so that someone else may be able to continue the issue during our codeathon. If not picked up I can grab it again!


AllReadyApp/Web-App/AllReady/Areas/Admin/ViewModels/Request/AddRequestCommentViewModel.cs
```diff
@@ -0,0 +1,12 @@
+namespace AllReady.Areas.Admin.ViewModels.Request
+{
+    public class AddRequestCommentViewModel
+    {
+        // possibly have a list of the comments which will be displayed below the dialog boxes
+
```
MisterJames: Great idea @stevejgordon, would you mind adding a corresponding issue so that we can track this into the getter command for this VM?

MisterJames:
@stevejgordon looks great! Please add that other issue so that we can track in the appropriate UI and getter command for those comments.

Issue #1871, opened 2017-03-14
account page - seperate tabs for 'skills' from rest of 'account info' [JT]


tonysurma:

babagy:
Hi, I am currently looking into this issue. Thanks.

shaunluttin:
Given that the last comment was March 18, 2017, I am assuming that no other developers are working on this issue. Is that right?

Also, if I understand correctly, this issue would like to have the skills appear on a separate tab.

![capture](https://user-images.githubusercontent.com/2066637/34706491-e99e1e28-f4bc-11e7-8ad9-87878bc9e8a3.PNG)

This seems like a good first issue for me, because I have experience with both Bootstrap tabs and with MVVM frameworks.

PR #2231, opened 2018-01-16
Move skills to own tab in profile management page.

shaunluttin:
This is a work in progress that resolves https://github.com/HTBox/allReady/issues/1871.

* [x] Add tabs for Account and Skills. 
* [x] Move all existing forms in to the Account tab. 
* [x] Move the skills form in to the Skills tab. 
* [x] Add unit tests.

Moving the skills form into the Skills tab requires a decision about whether to have

* a single save button for all the tabs or 
* a save button per tab.

My own sense is to use a save button per tab. Doing that might require adding another `Action` to the `ManageController`, which seems fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -241,3 +241,6 @@ bower_components
 /smoke-alarm-requests-sample-data-EMPTY.csv
 /smoke-alarm-requests-sample-data-DUPLICATE.csv
 /AllReadyApp/Web-App/AllReady/wwwroot/upload
+
+# VS Code Assets
+.vscode/
--- a/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/ManageControllerTests.cs
+++ b/AllReadyApp/Web-App/AllReady.UnitTest/Controllers/ManageControllerTests.cs
@@ -223,6 +223,7 @@ namespace AllReady.UnitTest.Controllers
                 nameof(ManageController.SaveProfile), 
                 new[] { typeof(ProfileViewModel) });
         }
+
         [Fact]
         public async Task SaveProfilePostSendsRemoveUserProfileIncompleteClaimCommandWithCorrectUserIdWhenUsersProfileIsComplete()
         {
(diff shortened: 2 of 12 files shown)


MisterJames:
@shaunluttin I agree, button per tab is fine. Either you can adorn the button with a command name and interpret on the existing action, or you can create a new action. Either approach is fine.  Great start here!

shaunluttin:
I changed computers recently, which slowed me down, due to having to set up the development environment.

shaunluttin:
As of 647b8252b4f3417bf330e151ead77cba875cc265 this PR is feature complete. A weak point of the implementation is that it runs model validation only when saving the profile.

shaunluttin:
My next step is to determine why the Jenkin's build is failing.

shaunluttin:
Looks like the unit testing are failing. :-) 

> Controllers\ManageControllerTests.cs(144,36): error CS1503: Argument 1: cannot convert from 'AllReady.ViewModels.Manage.IndexViewModel' to 'AllReady.Controllers.ManageController.ManageMessageId?' [C:\projects\allready\AllReadyApp\Web-App\AllReady.UnitTest\AllReady.UnitTest.csproj]

This is great! I am not used to having unit tests for the UI. 👏 :clap:

shaunluttin:
This is mostly done. 

There is a fundamental problem we need to work out though. 

When the model validation fails for `SaveProfile`, the address bar says `~/Manage/SaveProfile` whereas we expected it to say `~/Manage/Index`.

This is not necessarily bad; it might, though, point out a problem with the design that we are using.

shaunluttin:
I looked more into the situation (and possible UX problem) that I mentioned four days ago about the tabs. 

As far as I could tell, there are no other tabbed forms in the AllReady UI. So, the team needs to make a decision about how we handle form submission in a tabbed layout. E.g. 

1. Full page post back.
2. Partial page post back.

Based on that...

1. If full page post pack, do we worry about address bar consistency in the case of validation of errors?
2. If full page post back, do we reload the current tab or the main tab?
3. If partial page post back, what method do we use? JQuery, Knockout, other?

Those are some initial considerations.

shaunluttin:
Closing because this PR has not received a review.

PR #2253, opened 2018-01-20
label changes

Michael-Marshall:


AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Campaign/Edit.cshtml
```diff
@@ -73,7 +73,7 @@
                     <input type="hidden" asp-for="ImageUrl" id="image-url-input" />
                 }
                 <div class="form-group">
-                    <label asp-for="Name" class="col-md-2 control-label"></label>
+                    <label asp-for="Name" class="col-md-2 control-label">Campaign Title</label>
```
stevejgordon: The label actually binds from the property in the ViewModel and we can actually achieve this by adding a DisplayName attribute to the property in question. See #2252

Could this be updated to that approach please?

stevejgordon:
@Michael-Marshall Looks like we need to do the address display names and then we're good to go.

PR #2256, opened 2018-01-20
Add social buttons to Event detail page

### LINKED ISSUE #2245: Event Details (Public) - Add Social Share Buttons
We have publicly facing events in the allReady site which can be browsed by users, including public, unauthenticated users.

e.g. http://allready-d.azurewebsites.net/Event/1

It would be nice to include social share buttons on the event details page to allow people to easily share the event. We have this behaviour on the campaign details page currently so can hopefully replicate it for events.

dracan:
I've also added "Check out this X" text to the Campaign version too, aswell as fixed dodgy indentation.

Resolves #2245


dracan:
Note that I couldn't fully test the Facebook version, as Facebook failed due to what I assume was because the link was _localhost_. The existing Campaign version doesn't the same though, which I assume does work with a valid domain name. Likewise with the email, as I don't have a `mailto:` association on my machine. So this will also need testing.

Issue #2241, opened 2018-01-16
Event Details (Admin) - Hide Create New Task button when Event in Past


stevejgordon:
When logged in as a site admin or org admin, campaign manager or event manager, you can create Tasks for Events in allReady.

This is initiated from the admin campaign details page (e.g. http://allready-d.azurewebsites.net/Admin/Event/Details/18)

Above the tasks list there is a "Create Task" button. It makes sense to hide this button when the end date of the event is in the past. This will avoid someone trying to create new tasks in the past also.

It would be best to hide the button and instead display a message such as:
"It is now past the end date for this event. No new tasks can be created."

stevejgordon:
Reserved for .NET South East codeathon

redware:
I will look at this if that is OK.

PR #2267, opened 2018-01-20
#2241 Event Details (Admin) - Hide Create New Task button when Event …

redware:
…in Past

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Event/Details.cshtml
+++ b/AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Event/Details.cshtml
@@ -83,7 +83,15 @@
 {
     <div class="row">
         <div class="col-md-12">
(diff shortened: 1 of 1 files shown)


AllReadyApp/Web-App/AllReady/Areas/Admin/Views/Event/Details.cshtml
```diff
@@ -83,7 +83,15 @@
 {
     <div class="row">
         <div class="col-md-12">
-            <a asp-area="@AreaNames.Admin" asp-controller="VolunteerTask" asp-action="Create" asp-route-eventId="@Model.Id" class="btn btn-default">Create Task</a>
+            @if (Model.EndDateTime.CompareTo(DateTime.Now) < 0)
```
stevejgordon: I think the dates are UTC based so might need to use UtcNow instead here

MisterJames:
@redware are you able to make this small adjustment so that we can merge?

stevejgordon:
Looks like this got implemented in #2305

Issue #2208, opened 2018-01-15
Create Organization - Form changes (Part 3)


stevejgordon:
**_NOTE: To be picked up after #2207 is completed and merged._**

When logged in as a site admin, the user can create new organizations in allReady.

Currently we have a requirement to provide either a URL to a privacy policy or the actual privacy policy text. We require that one of these be provided and not both of them. Is there a better UI/UX that we can implement (perhaps a Vue.js component) that can make it clearer how this should be completed?

May need some design and front end thought first.

stevejgordon:
Reserved for the .NET South East codeathon

i-hardy:
I'll take this one! I think a good solution for this would be a block above with buttons that allow the user to select which one they want to fill out, and then show that form element. Any particular guidance on wording?

stevejgordon:
Closed in PR #2271

PR #2271, opened 2018-01-20
#2208 Privacy policy inputs now show/hide on button toggle

i-hardy:
I've added a pair of corresponding buttons over the privacy policy inputs and hidden both of them by default. You can switch back and forth between the two input types, but once one of them has been opened you can't hide both of them again. Also moved up the asp-validation-for spans to sit under the buttons so they don't get hidden.

### DIFF
--- a/AllReadyApp/AllReady.IntegrationTest/Pages.fs
+++ b/AllReadyApp/AllReady.IntegrationTest/Pages.fs
@@ -93,6 +93,9 @@ module AdminCampaignDetails =
         click createNew
 
 module AdminOrganizationCreate =
+    let privicyPolicyFieldVisible () =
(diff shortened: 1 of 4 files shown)


stevejgordon:
Looks good, however we have some UI tests (that I have very little idea about how they actually function) which are now breaking. It's presumably due to adding/renaming elements. Not sure if you can work out how to amend them? They'd be in AllReadyApp\AllReady.IntegrationTest for example.

i-hardy:
I've found the test in question and have an idea of how to fix it, but I'm in absolute hell trying to get the tests to run on my machine so I can check if I'm on the right track.

i-hardy:
I can't believe that worked

PR #2280, opened 2018-01-20
Fields marked with * are required in Edit

djhmateer:


djhmateer:
This was some text I forgot in the initial PR

PR #2294, opened 2018-01-21
Small code clean-up in startup

valdisiljuconoks:


valdisiljuconoks:
what is general policy with PRs like these? Does it make sense to continue looking for some smaller improvements in codebase while getting myself familiar with that?

MisterJames:
@valdisiljuconoks yes, this is a good strategy to get involved.

One suggestion would be to first create an issue that would detail what your proposal is. We'll try to be good at responding quickly!

Issue #2247, opened 2018-01-18
Investigate (and Create?) Container Images with Requisite Tools to Contribute


MisterJames:
We have had a couple of requests that the start-up process can be a bit intimidating.

For those who live in a containerized world, having a Docker image or Azure VM that was prepped to go would be a handy way to jump into a pre-configured world of allReady.

Admittedly, I'm not well enough versed in this area to know how big this task is.  We are looking for an SME that would be willing to offer some advice and/or create an easy "on-ramp" to help other who would like to contribute to allReady.

jonparker:
There's instructions for using a docker image for sql server for Linux / Mac. This could be used for devs running Windows too who don't have sql or are not wanting to install it. For the nodejs and other tools it is a bit more complicated. I started a box starter script for installing the pre reqs on a new machine. I'll try and find the link and post it here.

jonparker:
I'm wondering if providing a linux docker image with Nodejs, Rider and SQL server pre-installed would help... I'm wondering if people are finding it difficult due to not being familiar with Node/Git/Visual Studio/MS SQL Server.

MisterJames:
@jonparker I know for sure that would have helped at a few of our code-a-thons. 

As far as distribution goes, we'd likely want to set such images up through a CDN or similar, right? Should I start getting some accounts together? If so, who are the popular hosts/registries?

jonparker:
hub.docker.com account would be good for docker images. For boxstarter you don't need an image as it's a script that downloads everything for you.

Potentially we could do a azure image too but it's probably not worth it unless there's a need for people to develop in a VM.

I think we should solve the problem of setting up a local/container environment first.

jonparker:
Here's the link to the boxstarter script for setting up an environment for the mobile app development.

https://gist.github.com/jonparker/3f45edbd7eb17ba02c783d9e3b91bb23

mk0sojo:
Wouldn't something like this help https://docs.docker.com/compose/aspnet-mssql-compose/

Workflow for a user would be something like this:
1. Clone the allready repo
2. Run docker "docker-compose up" and that should set everything up to run the the Allready Web app.

I haven't worked on the mobile app so I'm not that familiar with that.

mk0sojo:
Had a go at creating docker support for debugging. Any feedback welcome.

PR #2298, opened 2018-01-27
Added docker support for debugging

mk0sojo:
For #2247

First shot at debugging the web app in a docker containers (one for the app and one for SQL server). This is the steps to try this out:

1. Download and install Git, Docker and Editor or IDE (e.g. VS Code).
2. Make sure Docker is set up to use Linux containers (if running on Windows) and got 4 GB memory or more allocated to run SQL Server (set in Preferences).
3. Fork and clone repository
4. Open terminal or Powershell
5. Cd to the AlreadyApp directory and type “docker-compose up”. This will take some time the first time it runs as it will download ASP.Net core and SQL Server base images.
6. Open a browser and navigate to localhost:8000

The main remaining issue is after changing anything in the project the whole project is restored, build and nom packages are installed (even if it is just a change to a HTML file)

Will try to solve it, but please let me know what you think so far. I have only tried it on a Mac so far but it “should” work on Windows as well.

### DIFF
--- a/AllReadyApp/Web-App/AllReady/Startup.cs
+++ b/AllReadyApp/Web-App/AllReady/Startup.cs
@@ -52,8 +52,11 @@ namespace AllReady
             });
 
             // Add Entity Framework services to the services container.
-            services.AddDbContext<AllReadyContext>(options => options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
-
+            var allReadyDbConnectionString = Environment.GetEnvironmentVariable("ALLREADY_DOCKER_DEBUG") == "TRUE" ?
+                Configuration["Data:DockerDefaultConnection:ConnectionString"] :
+                Configuration["Data:DefaultConnection:ConnectionString"];
+            services.AddDbContext<AllReadyContext>(options => options.UseSqlServer(allReadyDbConnectionString));
+            
             Options.LoadConfigurationOptions(services, Configuration);
 
             services.AddIdentity<ApplicationUser, IdentityRole>(options =>
@@ -168,7 +171,10 @@ namespace AllReady
             });
 
             //Hangfire
-            services.AddHangfire(configuration => configuration.UseSqlServerStorage(Configuration["Data:HangfireConnection:ConnectionString"]));
+            var hangFireConnectionString = Environment.GetEnvironmentVariable("ALLREADY_DOCKER_DEBUG") == "TRUE" ?
+                Configuration["Data:DockerHangfireConnection:ConnectionString"] :
+                Configuration["Data:HangfireConnection:ConnectionString"];
+            services.AddHangfire(configuration => configuration.UseSqlServerStorage(hangFireConnectionString));
 
             services.AddScoped<IAllReadyUserManager, AllReadyUserManager>();
             services.AddScoped<IUserAuthorizationService, UserAuthorizationService>();
@@ -184,7 +190,6 @@ namespace AllReady
         {
             // Put first to avoid issues with OPTIONS when calling from Angular/Browser.
             app.UseCors("allReady");
-
             app.UseSession();
 
             // Add the following to the request pipeline only in development environment.
--- a/AllReadyApp/docker-compose.yml
+++ b/AllReadyApp/docker-compose.yml
@@ -1,0 +1,22 @@
+version: "3"
+services:
+    web:
+        build: 
+            context: .
+            dockerfile: Dockerfile-web
+        ports:
+            - "8000:80"
+        depends_on:
+            - db
+    db:
+        image: "microsoft/mssql-server-linux:2017-CU3"
+        volumes: 
+            - sql-server-data:/var/opt/mssql
+        ports: 
+            - "1433:1433"
+        environment:
+            SA_PASSWORD: "yourStrong(!)Password"
+            ACCEPT_EULA: "Y"
+
+volumes:
+    sql-server-data:
--- a/AllReadyApp/entrypoint.sh
+++ b/AllReadyApp/entrypoint.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+run_cmd="dotnet run --server.urls http://*:80"
+
+until dotnet ef database update; do
+>&2 echo "SQL Server is starting up"
+sleep 1
+done
+
+>&2 echo "SQL Server is up - executing command"
+exec $run_cmd
--- a/AllReadyApp/Dockerfile-web
+++ b/AllReadyApp/Dockerfile-web
@@ -1,0 +1,12 @@
+FROM microsoft/aspnetcore-build:2.0
+ENV ASPNETCORE_ENVIRONMENT=Development
+ENV ALLREADY_DOCKER_DEBUG=TRUE
+COPY . /app
+WORKDIR /app
+RUN ["dotnet","restore","AllReadyWebOnly.sln"]
+WORKDIR /app/Web-App/AllReady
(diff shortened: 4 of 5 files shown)


jonparker:
@mk0sojo That's awesome work. Hopefully this will make it much easier for people to get started.
Have you tested debugging with VS code or Visual Studio?

mk0sojo:
@jonparker  Still needs to add a better debugging experience and need to speed up the restore / build process a bit.
Will look into it a bit more.

mk0sojo:
@jonparker @MisterJames Added support for debugging in VS Code and tried to speed up the build time a bit.
To try:
1. Clone repo
2. In AllreadyApp folder run docker-compose up
3. When done open another terminal and type docker ps. allreadyweb and allreadysqlserver should be running
4. Open the AllreadyApp folder in VS Code. Go to the debug tab and press play (or just F5). Brake points should now work

image

The only (known) limitation is that it is not possible to edit an HTML or CSS file without rebuilding. 

Let me know what you think.

jonparker:
That's great. Have you heard of docker-sync? It allows for you to build within the container while having the files checked out on the host. 

http://docker-sync.io

mk0sojo:
It looks like it could be usefull, my only concern is that it looks a bit complicated to install (on windows in particular https://github.com/EugenMayer/docker-sync/wiki/docker-sync-on-Windows). I'm not sure if that is any better than the installation instructions we currently have running everything locally? On our codathons, is there any particular step that users find difficult or is it just the number of steps?

tony-ho:
@mk0sojo I tested this on Windows and there was an issue with the line endings in `AllReadyApp/entrypoint.sh`:

```
allreadyweb | ../../entrypoint.sh: line 2: $'\r': command not found 
: invalid option/../entrypoint.sh: line 3: set: - 
allreadyweb | set: usage: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...] 
allreadyweb | ../../entrypoint.sh: line 12: syntax error: unexpected end of file 
allreadyweb exited with code 2 
```

The cause was the bash script being checked with the Windows line-ending, `\r\n`. When run on Linux inside the Docker container, the `\r` breaks the script.

The issue is described is more detail here:
https://techblog.dorogin.com/case-of-windows-line-ending-in-bash-script-7236f056abe

The solution is to configure `.gitattributes` to convert bash script line endings to LF on checkout so it will retain the Unix line-endings regardless of operating system.

I've made this change to a clone of your branch here:
https://github.com/tony-ho/allReady/tree/issue-branch-2247

I also rewrote the Docker connection string logic. Instead of using the `ALLREADY_DOCKER_DEBUG` environment variable to conditionally set Docker-specific connection strings, the connection strings can be passed as environment variables via `docker-compose.yml`. These will override the values in `appSettings.json`.

Let me know if and how you want to merge these changes.

mk0sojo:
@tony-ho Thanks for the feedback and testing it on Windows. The changes to the connection string logic looks great as well. Let me know if you have any more feedback.

MisterJames:
This is great stuff.

I'd like to merge this in, do we need another commit @mk0sojo for those line endings?

mk0sojo:
@MisterJames I have added @tony-ho changes to the pull request so it should be ok to merge now.

mk0sojo:
@MisterJames Is there anything else missing here before it can be merged?

AllReadyApp/Web-App/AllReady/Dockerfile.debug
```diff
@@ -0,0 +1,12 @@
+FROM microsoft/aspnetcore-build:2.0
```
gparlakov, line 1: @mk0sojo This whole dockerfile seems to do nothing more than just install the vs debugger and then on startup just keep the process running. What purpose does it serve?

AllReadyApp/Web-App/AllReady/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+WORKDIR /app/Web-App/AllReady
+
+# Build project
+RUN ["dotnet","build"]
+
+# Run gulp tasks
+RUN gulp clean && gulp min
+EXPOSE 80/tcp
+RUN chmod +x ../../entrypoint.sh
+CMD /bin/bash ../../entrypoint.sh
```
gparlakov, line 30:
```suggestion
CMD /bin/bash "dotnet ef database update && dotnet AllReady.dll"
```
Instead of having a whole separate file (`entrypoint.sh`) I believe we could just simply chain the migration and the app start. That we we can drop the .sh file and also the end-of-line issues discussed in the PR comments.

AllReadyApp/Web-App/AllReady/Startup.cs
```diff
@@ -169,7 +169,7 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
 
             //Hangfire
             services.AddHangfire(configuration => configuration.UseSqlServerStorage(Configuration["Data:HangfireConnection:ConnectionString"]));
-
+            
             services.AddScoped<IAllReadyUserManager, AllReadyUserManager>();
```
gparlakov, line 173: This file seems to only have whitespace changes - perhaps it could be excluded/rolled-back from the PR (i.e. `github checkout origin/master -- AllReadyApp/Web-App/AllReady/Startup.cs`)

AllReadyApp/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+            dockerfile: Web-App/AllReady/Dockerfile
+        container_name: allreadyweb
+        environment:
+            ASPNETCORE_ENVIRONMENT: Development
+            Data__DefaultConnection__ConnectionString: "Server=db,1433;Initial Catalog=master;Persist Security Info=False;User ID=sa;Password=yourStrong(!)Password;MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=True;Connection Timeout=30;"
+            Data__HangfireConnection__ConnectionString: "Server=db,1433;Initial Catalog=master;Persist Security Info=False;User ID=sa;Password=yourStrong(!)Password;MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=True;Connection Timeout=30;"
+        ports:
+            - "8000:80"
+        depends_on:
+            - db
```
gparlakov, line 15: Since the db container takes a bit of time to spin up, I'd suggest adding
```yaml
restart: on-failure
```
So that docker could retry starting the app if the db is not yet up and running.

AllReadyApp/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+        volumes: 
+            - sql-server-data:/var/opt/mssql
+        ports: 
+            - "1433:1433"
+        environment:
+            SA_PASSWORD: "yourStrong(!)Password"
+            ACCEPT_EULA: "Y"
+
+volumes:
+    sql-server-data:
```
gparlakov, line 28: Also I could not see where is the logic that actually creates the database table `AllReady` like done in the appveyor run script https://github.com/HTBox/allReady/blob/37833fcf2ea4d1f54d0e846c78ff79b307509809/appveyor.yml#L46-L47 for example

AllReadyApp/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+run_cmd="tail -f /dev/null"
+until dotnet ef database update; do
+>&2 echo "SQL Server is starting up"
+sleep 1
+done
+
```
gparlakov, line 9: This file seems to only run the migration and does not actually start the app. i.e. `dotnet AllReady.dll` is nowhere to be seen...

Issue #2287, opened 2018-01-20
Re-use component for editing Locations


morcs:
In Campaign\Edit.cshtml the location fields are rendered using

`@Html.EditorFor(m => m.Location)`

In Event\Edit.cshtml the markup is inline. It looks like we could just re-use the EditorFor here. The only differences I can see are:

`if (Model.EventType == EventType.Rally)` it sets the "Address 1" field to have `data-val-required`, but I think we might be requiring an address on all events anyway, and using EditorFor will include this. 

There's an extra CSS class `location-address1` too but searching the code it seems to be unused.

stevejgordon:
Looks like this can be closed

PR #2299, opened 2018-02-15
Re-use component for editing Locations #2287

SirwanAfifi:
As suggested on #2287 we can re-use EditorTemplate for `Event\Edit.cshtml` . for required field (Address1) we can pass additional parameter to the EditorTemplate:
`@Html.EditorFor(m => m.Location, new { RequiredAddressOne = Model.EventType == EventType.Rally })`


MisterJames:
Great stuff, thanks @SirwanAfifi !

PR #2302, opened 2018-02-23
Sending email to inviter when invite accepted

### LINKED ISSUE #1893: Send Invite Acceptance Email to Sender of the Invite
Breakout from #1804 

When an invite is accepted we should notify the sender of the invite via email of the status change. We should implement this by firing an invite accepted notification (Mediatr) and then have a handler which will send the email message to the queue as necessary.

mk0sojo:
Closes #1893


MisterJames:
Great stuff, thanks @mk0sojo!

PR #2305, opened 2018-03-06
Only show create task button before event end. Fixes #2241

### LINKED ISSUE #2241: Event Details (Admin) - Hide Create New Task button when Event in Past

stimms:
Unsure if there are tests which exercise the front end.


MisterJames:
Thanks @stimms!

PR #2307, opened 2018-03-06
Fix async call in startup

### LINKED ISSUE #2250: Potential deadlocks due to .Result calls?
Scanning through the code ready for the codeathon tomorrow, I've spotted a few `.Result` calls. One example is in `SendRequestStatusToGetASmokeAlarm.Send`, where it does this: `_httpClient.SendAsync(request).Result`. There are a few more examples though. This can cause deadlocks and should be avoided. I've lost countless hours in the past due to someone using `.Result`, and causing random deadlocks.

MisterJames:
closes #2250


valdisiljuconoks:
guess it should be renamed to `CreateAdminUserAsync`

PR #2313, opened 2018-03-07
Fixing some weird EF stuff

dpaquette:
- Every time you saved a Campaign, Event or Org, a new Location record was being created
- We had a custom `AddOrUpdate` extension method that was probably created as a workaround to an issue in a very early beta of EF Core. I fixed all usages of that and removed the method.


AllReadyApp/Web-App/AllReady/Areas/Admin/Features/Events/EditEventCommandHandler.cs
```diff
@@ -76,7 +77,8 @@ public EditEventCommandHandler(AllReadyContext context)
             return await _context.Events
                 .Include(a => a.RequiredSkills)
                 .Include(a => a.VolunteerTasks)
+                .Include(a => a.Location)
```
dpaquette: We need to include Location here, otherwise it creates a new location every time

AllReadyApp/Web-App/AllReady/Models/Location.cs
```diff
@@ -111,7 +111,7 @@ public static Location UpdateModel(this Location location, LocationEditViewModel
                 }
                 else
                 {
-                    if (location == null || locationEditModel.Id.GetValueOrDefault() != 0)
+                    if (location == null || locationEditModel.Id.GetValueOrDefault() == 0)
```
dpaquette: This was also causing a new location to be added every time

PR #2351, opened 2019-03-13
WIP - Displays version of Node and NPM used for build

c0g1t8:
**Do not merge this PR**

Need to confirm the AppVeyor build server setup.  The initial assumption was the Node 10 was being used but other tests indicate that it's Node 8.15.


c0g1t8:
The following are the reported version numbers on the build server.
> 367 Write-Host "Version of Node and NPM used for Build"
> 368 Version of Node and NPM used for Build
> 369 node -v
> 370 v8.15.0
> 371 npm -v
> 372 6.4.1

The next test will be to ensure that the **lib** folder is being published.  Errors seen in the **ScenarioTest** could be due to missing javascript libraries.

c0g1t8:
The contents of the wwwroot/lib did contain the expected JavaScript libraries.

This exercise did reveal that I made an initially erroneous assumption that the build system was on Node 10.  It is actually on Node 8.15 which explains why _gulpfile.js_ worked on the build server but not locally where Node 10 was installed.

PR #2370, opened 2021-04-22
Dead link fixed

craftsbyshuvro:
Dead link fixed, If you need other revisions of the link you can get that from https://web.archive.org/ . You just need to search by the link and choose available revisions according to date.


craftsbyshuvro:
Can you specifically point where was the problem in my commit? I just changed the link in the comment.

